PR #3, opened 2026-03-24
Add Responses MVP

jiahuei:
## Summary

This PR adds a full OpenAI Responses API implementation for vLLM. It passes all OpenResponses compatibility tests and ships with a complete MkDocs documentation site and CI pipeline.

## What's working

- Responses API backed by Chat Completions (or stateless Responses if desired from core vLLM)
- Verified against all [OpenResponses](https://www.openresponses.org/compliance) compatibility tests
- Reasoning/thinking pass-through tested with models that support interleaved thinking (GLM, MiniMax)
- 2 built-in tools already working via MCP interface: code interpreter and web search
- Support for remote MCP servers
- Full documentation site with MkDocs + CI pipeline

## Why Chat Completions over raw `generate`?

There has been discussion around bypassing Chat Completions entirely and building the Responses API against vLLM's raw `generate` (token-in, token-out) endpoint instead. We believe that path introduces significant engineering and maintenance costs:

- **Ecosystem fragmentation:** Every model support addition and bug fix in core vLLM would need to be duplicated in this project. That's hundreds of existing models, plus every new one going forward.
- **Duplicated effort:** Chat templating, tokenization, performance metrics, and correctness tests would all need to be re-implemented outside core vLLM. This is work that already exists and is battle-tested upstream.
- **Upstream dependency risk:** The token-in-token-out endpoint is still WIP. Building on it means getting blocked by its progress, whereas Chat Completions is mature and stable today.

For the one known gap, which is GPT-OSS/Harmony models where Chat Completions may lose channel nuance — the right fix is to ensure Harmony is correctly supported via chat templates and schema extension in core vLLM, not to bypass the entire Chat Completions stack. @bbrowning (Red Hat) agrees: token-to-request conversion logic should live in core vLLM, not be reimplemented externally.


noobHappylife:
Hi @leseb, 

Referring to the [meeting notes](https://docs.google.com/document/d/1zJUfxdloxu9_oBYfbfTtgbz5sSZxwAnqvlnt72T67oU/edit?tab=t.i3nieswphudh) this is a reference design to show how we imagine the implementation would be. 

Understood that there's a new discussion on having some RFCs on the design now,  We are also preparing some RFCs so we can keep the ball rolling on the MVP implementation details. Thanks :)

Issue #14, opened 2026-03-25
RFC: Response Store semantics for `previous_response_id`


noobHappylife:
## Summary

This RFC proposes introducing a **Response Store** as the state layer behind `previous_response_id`.

The goal is to let clients continue a conversation by referencing a prior response, without having to resend the entire conversation history on every turn.

This RFC focuses on the **behavioral model and design questions** for the MVP:

- what a stored response should represent
- what data must be preserved for continuation
- what should be inherited across turns
- what tradeoffs the MVP should accept
- whether the storage model should begin as a single-table design or a more normalized multi-table design

---

## Problem

A Responses-style API needs a way to continue from an earlier response.

Without a server-side state layer, the client must resend the entire prior conversation each time. That creates several problems:

- larger payloads
- more client-side state management
- more chances for client/server divergence
- weaker ergonomics for multi-turn usage
- less clarity around what exactly is being continued

For `previous_response_id` to be useful, the server needs a clear model for what is stored and how that stored state is used to reconstruct the next turn.

---

## Proposal

Introduce a **Response Store** that treats a stored response as a **continuation checkpoint**.

At a conceptual level, a stored response should preserve enough information for the server to:

1. understand the full context used to generate that response
2. understand the output produced by that response
3. reconstruct the next request when a new turn references that response via `previous_response_id`

A likely continuation model is:

> next turn context = prior context + prior output + new input

This suggests that a stored response may need to be more than just a raw model output. It may need to act as a reusable state object for future turns.

---

## Goals

- make `previous_response_id` usable without requiring full history replay
- define what the server considers the canonical continuation state
- keep the continuation model predictable and debuggable
- preserve enough structure to support richer outputs, not just plain text
- leave room for the storage format and schema to evolve over time

---

## Non-Goals

This RFC is not trying to define:

- long-term memory
- semantic memory or summarization
- cross-conversation knowledge
- analytics/reporting architecture
- exact database technology or infrastructure wiring

This is specifically about the state model required for **conversation continuation**.

---

### 1. What should a stored response represent?

Should a stored response be:

- the minimum data needed to continue the next turn, or
- a fuller checkpoint that also supports inspection, debugging, and auditability?

This affects both the shape of the storage model and the guarantees of `previous_response_id`.

---

### 2. What data must be preserved for continuation?

Possible candidates include:

- prior flattened input history
- prior model output
- tool calls
- tool results
- reasoning items
- instructions / system input
- tool configuration
[...]

PR #15, opened 2026-03-27
[Doc] ADR-02 — persistent store architecture and community decisions

noobHappylife:
## Summary

Adds ADR-02 as a discussion-oriented ADR for response store semantics behind previous_response_id.

The document is structured around the main decision points still under discussion:

- continuation shape: single-hop checkpoint vs linked chain
- storage model: single-table vs multi-table
- backend model: relational-first vs non-relational-first
- what needs to be stored for correct rehydration

It also incorporates rationale from the RFC issue thread #14 and compatibility-oriented notes based on observed OpenAI-style behavior. 

This is a Draft. Please poke holes.


docs/adr/ADR-02_response_store.md
```diff
@@ -0,0 +1,316 @@
...
+A single `response_store` table holds a small set of queryable metadata columns plus a flexible payload for the stored checkpoint body.
+
+This tends to align with single-hop continuation because the persisted row is close to the rehydration unit.
+
+### Option B: Multi-table normalized model
+
+Separate entities capture different concerns:
+
+- `responses`
+- `prompts`
```
maralbahari: If we suppose to choose a multi-table design for a start point we can mirror the OpenAI Responses API like:

- **`responses`** entity ([ref](https://developers.openai.com/api/reference/resources/responses/methods/create)):
  the inference artifact — one LLM turn, input in, output out. stores the checkpoint needed for
  `previous_response_id` rehydration. ephemeral by nature, has a TTL.

- **`conversations`** entity ([ref](https://developers.openai.com/api/reference/resources/conversations/methods/create)):
  the durable thread — groups multiple responses together so the full history is queryable without
  walking the `previous_response_id` chain. persists beyond individual response TTLs.

- **`items`** entity ([ref](https://developers.openai.com/api/reference/resources/conversations/subresources/items/methods/create)):
  the atomic content unit — one row per message, tool call, or tool result within a conversation.
  lets you read, paginate, or display history without deserializing full response blobs.

A `prompts` table is not useful because a prompt has no independent lifecycle — it cannot exist
without a response, and its content is already captured inside the response row.

leseb (commented):
Thanks for the proposal! The ADR is missing a multi-user story. Since previous_response_id is the primary entrypoint to this store, ownership and access control aren't future concerns, they're baseline requirements.

IMO we should add a Decision Point 5 covering who owns a response and how access is enforced. it affects the tradeoff analysis for D1-D4.

Thanks!

chaunceyjiang (commented):
When designing the persistent storage,there’s another thing we need to consider: in the future, the Responses API will definitely support **PD (prefill/decode) disaggregation** — see https://github.com/vllm-project/vllm/pull/37534.

When a request includes `previous_response_id`, it will first go to the **prefill node** for processing, and then the request will be forwarded to the **decode node**. In that case, it’s clear that **both the prefill and decode nodes** should be able to reconstruct the full context based on `previous_response_id`.

jiahuei:
> Thanks for the proposal! The ADR is missing a multi-user story. Since previous_response_id is the primary entrypoint to this store, ownership and access control aren't future concerns, they're baseline requirements.
> 
> IMO we should add a Decision Point 5 covering who owns a response and how access is enforced. it affects the tradeoff analysis for D1-D4.
> 
> Thanks!

@leseb While RBAC is a valid concern, we feel it would complicate the design at this stage:

1. RBAC adds significant surface area for an MVP, especially since we decided to start simple.
2. ⁠Production deployments already have an auth layer that handles authentication, rate limiting, and access control.

Since Responses API already includes an optional `safety_identifier` string field, perhaps we could use this field to scope session ownership using a simple exact-match rule. A mismatch returns 404. No special cases and no branching logic.

This means:

1. Single-user / dev setups just work. Omit the field on every request and all responses are in the same pool.
2. Multi-tenant setups get session isolation by passing a consistent identifier per caller. Operators who need full RBAC handle and validate it at the gateway layer

tjtanaa:
> Thanks for the proposal! The ADR is missing a multi-user story. Since previous_response_id is the primary entrypoint to this store, ownership and access control aren't future concerns, they're baseline requirements.

@leseb  As discussed https://vllm-dev.slack.com/archives/C0AJQFD1REY/p1774068786903749?thread_ts=1773952355.057799&cid=C0AJQFD1REY, to get wide audience traction, a best way to is deliver the same vllm serve user experience and first stage I would recommend to focus on single device small scale deployment feature without heavy setups, and no configuration setup out of the box. (for extra or advanced feature, user can specify custom configurations)

MVP should focus on getting Agent features out to help user solve problems e.g. tooling, web_search, mcp tools, code interpreter.

We can expand later on. Moreover, right now there are already the llm-d and llamastack that offers very rich feature sets.

Many companies already have their own stack ready.  A more beneficial direction for this opensource project is to reach out opensource community and focuses on ease of use.

RBAC will definite be the next step but shouldn't be included in MVP.

docs/adr/ADR-02_response_store.md
noobHappylife: - for `conversations` and `items`, i think it's a question do we want to do the conversations API?

noobHappylife:
@chaunceyjiang 
Definitely, we can put it into future roadmap for supporting PD inference. We can basically target both the prefill and decode instance so that the history is properly send to both.

docs/adr/ADR-02_response_store.md
franciscojavierarceo: I think we will need to do the conversations API

docs/adr/ADR-02_response_store.md
```diff
@@ -0,0 +1,316 @@
...
+**Self-contained checkpoint**
+
+Pros:
+
+- simpler rehydration path
+- clearer continuation semantics
+- easier realtime lookup behavior
+- easier to reason about partial reconstruction failures
+
+Cons:
```
franciscojavierarceo: It will also require more writes to a single table which means reconstructing failures could be harder

docs/adr/ADR-02_response_store.md
```diff
@@ -0,0 +1,316 @@
...
+
+### Tradeoffs
+
+**Self-contained checkpoint**
+
+Pros:
+
+- simpler rehydration path
+- clearer continuation semantics
+- easier realtime lookup behavior
```
franciscojavierarceo: and faster

docs/adr/ADR-02_response_store.md
```diff
@@ -0,0 +1,316 @@
...
+
+- less duplicated data
+- cleaner normalized history model
+- more direct mapping to turn-by-turn conversation structure
+
+Cons:
+
+- more request-time orchestration
+- harder failure semantics
+- more complex history retrieval
```
franciscojavierarceo: yes and slower

docs/adr/ADR-02_response_store.md
```diff
@@ -0,0 +1,316 @@
...
+### Rationale from issue discussion
+
+The issue comments pushed on two related concerns:
+
+- single-hop retrieval is easier to explain and may better support realtime history access
+- normalized multi-entity designs may be more extensible in the long run
+
+Those concerns are both valid. The main question is whether MVP should optimize first for correctness and simple rehydration, or for normalized modeling and future lifecycle separation.
+
+Observed OpenAI-compatible behavior suggests that continuation may still succeed when a follow-up request provides previous_response_id and tool output without resending tools. That implies the effective tool context from the prior turn must be recoverable during rehydration. That is a useful signal that the effective tool context from the prior turn must be recoverable somewhere, which in turn strengthens the case for a richer checkpoint and simpler one-hop rehydration.
```
franciscojavierarceo: Agree with one-hop hydration

docs/adr/ADR-02_response_store.md
franciscojavierarceo: >A prompts table is not useful because a prompt has no independent lifecycle — it cannot exist
without a response, and its content is already captured inside the response row.

That's actually not true. OpenAI's full prompts API is not available but I implemented the closed source pieces in Llama Stack. It would be natural for us to replicate that behavior so that it can be supported by the Responses API.

franciscojavierarceo: >Chain-based history retrieval is also possible with these entities without walking the
previous_response_id linked list. on every response completion, each turn's input and output
messages are appended as individual rows into conversation_items, all tagged with the same
conversation_id.

Agree here.

docs/adr/ADR-02_response_store.md
```diff
@@ -0,0 +1,316 @@
...
+### Rationale from issue discussion
+
+The comments on [Issue #14](https://github.com/vllm-project/agentic-stack/issues/14) explicitly surfaced this tension:
+
+- single-table is attractive because retrieval can stay close to O(1)-style single lookup behavior
+- multi-table is attractive because it may extend more cleanly as the product grows
+
+This is likely the most important architectural decision in the document because it shapes both the rehydration path and the future data model.
+
+Another practical consideration is schema evolution. A single-table model with a small queryable envelope plus versioned payload can absorb early payload changes with less migration churn while continuation semantics are still settling.
```
franciscojavierarceo: I recommend separate tables. That's what we've done in Llama Stack and the isolation is useful.

jiahuei: May we know what is the isolation useful for?
I prefer single table not just because it's faster, but it won't require DB migration if or when new models introduce new chat fields

franciscojavierarceo: The isolation provides several key benefits that become important as the system scales:

1. Data integrity and consistency: Each entity has distinct constraints and relationships. Responses contain the actual inference artifacts, conversations track the sequence and metadata, and prompts (if
implemented) handle template reuse. Mixing these creates a wide table with many nullable fields and makes it harder to enforce proper constraints.
2. Query optimization: Different access patterns for each entity type. Conversations are queried by user/session for listing, responses by ID for continuation, and prompts for template matching. Separate tables
 allow targeted indexing rather than compromising with a one-size-fits-all approach.
3. Security and access control: Different permission models may be needed. For example, response content might need different access controls than conversation metadata for compliance reasons.
4. Operational flexibility: Independent policies for backup/archival, storage tiers (response payloads might need blob storage while conversation metadata needs fast access), and different scaling
characteristics.
5. Future-proofing for access control: Even though we're not implementing RBAC in the MVP, the database structure needs to accommodate it eventually. With separate tables, adding user ownership, permissions,
and multi-tenancy becomes straightforward - you add the necessary foreign keys and policies to each table as needed.

With a single table approach, retrofitting access control becomes chaotic because:
- You'd need complex row-level security policies that understand different entity types within the same table
- User ownership becomes ambiguous (who owns a "row" that contains both conversation metadata AND response content?)
- Auditing becomes nearly impossible when different logical entities are mixed in the same table
- Data retention policies become a nightmare (conversations might need 30-day retention while responses need 7-day retention)

Regarding schema evolution: While single tables avoid migrations, they create nullable column proliferation as different model types add fields.

Not supporting RBAC until later strengthens this the case for separate tables. This is because the later you wait, the more expensive and risky the migration becomes. With users and data in production, you
can't easily restructure a monolithic table without significant downtime and data migration complexity.

IMO it is better to start with the right structure that can gracefully evolve than to paint yourself into a corner that requires a major rewrite when multi-tenancy becomes essential. The join overhead you mention is
valid for simple cases, but becomes negligible compared to these benefits as you scale to complex queries across thousands of conversations.

jiahuei: Oh I think this issue might be misunderstood. The single table here refers strictly to the strategy for storing conversation history and messages, not literally one table for the entire product (which would indeed be a very wide table).

To be specific, we are in favor of storing the chat history as a non-nullable JSON column (ie chat as the basic unit), instead of storing messages as the basic unit in its own table (which would require history reassembling, resolving message ordering, etc). Not mentioning message schema changes, for example OpenAI just added [phase](https://developers.openai.com/api/docs/guides/prompt-guidance#phase-parameter).

Further, if we intend to implement Conversations, then it would live in its own DB table.

jiahuei: This is what we propose:

```
erDiagram
    Conversation {
        UUIDv7 id PK
        JSON metadata
        datetime created_at
        datetime updated_at
    }

    Response {
        UUIDv7 id PK
        UUIDv7 conversation_id FK "nullable"
        UUIDv7 previous_response_id FK "nullable, self-ref"
        JSON history
        datetime created_at
        datetime updated_at
    }

    Conversation ||--o{ Response : "has many"
    Response ||--o| Response : "previous"
```

## How It Works

- `history` stores the **complete** chat history on every response — no chaining needed to reconstruct context.
- `previous_response_id` is stored for **traceability** (knowing which response was continued from), not for history reconstruction.
- `conversation_id` is **nullable** — a response can exist standalone without belonging to a conversation.

## Usage Flow

1. **Create Conversation X** — gets back `X.id`.
2. **Create Response Y** with `conversation_id = X.id` — history is built fresh, stored in full on Y.
3. **Create Response Z** referencing either:
   - **Conversation X** → `conversation_id = X.id`, app loads the latest response's history, appends new turn, stores complete history on Z, sets `previous_response_id` to the latest response.
   - **Response Y** → `previous_response_id = Y.id`, app loads Y's history, appends new turn, stores complete history on Z, inherits `conversation_id` from Y.

docs/adr/ADR-02_response_store.md
zilchfp: > A prompts table is not useful because a prompt has no independent lifecycle 

I think the prompts table will be useful for troubleshooting or reproducing some bugs

docs/adr/ADR-02_response_store.md
noobHappylife: @franciscojavierarceo @leseb 
Any other comments on the proposal from @jiahuei ? We can use this as the initial MVP first?

noobHappylife: We may need to store the conversation items (each message as 1 entry) if we are going to fully implement the conversation API (including manipulating the items: https://developers.openai.com/api/reference/resources/conversations/subresources/items)

jiahuei: We can still store items as a list while supporting the Items API. In this case, `Conversations` will have an additional JSON column `items`, and `GET` and `DELETE` will utilise JSON operators to fetch items or modify the list in place.

```
erDiagram
    Conversation {
        UUIDv7 id PK
        JSON items
        JSON metadata
        datetime created_at
        datetime updated_at
    }

    Response {
        UUIDv7 id PK
        UUIDv7 conversation_id FK "nullable"
        UUIDv7 previous_response_id FK "nullable, self-ref"
        JSON history
        JSON metadata
        datetime created_at
        datetime updated_at
    }

    Conversation ||--o{ Response : "has many"
    Response ||--o| Response : "previous"
```

docs/adr/ADR-02_response_store.md
maralbahari: For the MVP rehydration path, the core requirement is retrieving prior conversation history to reconstruct the full input for the next turn, which can be achieved through the conversation and responses tables similar to the OpenAI API design. The relationship between a prompt table and the rehydration flow is not yet clear. If the community can propose an ER diagram showing how the prompt table fits into this flow that would help move the discussion forward..

docs/adr/ADR-02_response_store.md
franciscojavierarceo: I would recommend looking at Llama Stack here for guidance. We have built, tested, and verified the Conversation API implementation rather rigorously. See here for the models and DB updates.

https://github.com/llamastack/llama-stack/blob/adb81a7c96db26a6125983dfdb04667fdd10be35/src/llama_stack_api/conversations/models.py

https://github.com/llamastack/llama-stack/blob/adb81a7c96db26a6125983dfdb04667fdd10be35/src/llama_stack/core/conversations/conversations.py

We support a SQL backend for it.

franciscojavierarceo: @jiahuei I do not recommend a JSON history though, that can get complicated in serialization or in maximum length.

jiahuei: @franciscojavierarceo not to say I am advocating for JSON history, but can you help me understand how is it more complicated than reading multiple rows from DB and reconstructing the history? Even if we don't use JSON, we still would need to parse the items in order to store into DB right?

In fact for our MVP, we first explored storing flat items/messages instead of JSON history, but found that it complicates our code a lot.

franciscojavierarceo: If history grows larger than maximum character length for our DB we will have to split history across multiple rows and have branching logic just for that. To me, it seems more obvious to just design for multiple rows and not have to deal with that logic.

jiahuei: @franciscojavierarceo hmm ok I totally understand your concern.

However I do want to highlight that:
- Even the default limits are absurdly huge: 1GB which is anywhere from 50M to 250M tokens, no LLM comes even close to it
- We would be trading off this theoretical concern with multiple reads and O(N) per retrieval

jiahuei: @franciscojavierarceo it also seems like [LlamaStack stores messages as JSON?](https://github.com/llamastack/llama-stack/blob/34edf53b48eae6f2cad79c7f27e7f7460a78871f/src/llama_stack/providers/utils/responses/responses_store.py#L74)

franciscojavierarceo: @jiahuei good catch!

Let me correct myself — when I said "separate tables," I was thinking about how we built the **Conversations API**, where we store items as individual rows in a `conversation_items` table. That's the part I wrote.

I went back and checked our **Responses** storage and it actually does what you're proposing: each response blob accumulates the full history from all prior turns. 

I think that's wrong in Llama Stack too. On turn `N`, you're storing turns `1–N` in the blob, which means `O(N^2)` total storage across the conversation, and if you ever need to redact a message from turn 5 you have to patch every subsequent response.

Your concern about `O(N)` retrieval with individual rows is valid. What works well for us on the Conversations side is a checkpoint cache (store items as individual rows (source of truth) but maintain a denormalized message checkpoint that updates after each turn). You get `O(1)` reads for continuation without snowballing blobs.

Reference:
- [`conversation_items` + `conversation_messages` tables](https://github.com/llamastack/llama-stack/blob/main/src/llama_stack/core/conversations/conversations.py)
- [`responses_store.py`](https://github.com/llamastack/llama-stack/blob/main/src/llama_stack/providers/utils/responses/responses_store.py).

franciscojavierarceo (approved):
This lgtm.

I mentioned in slack it may make sense for us to implement conversation id generation regardless if the client asks for `conversation_id` as that may avoid us having to store history entirely but I'm not going to block on that as we had come to an agreement yesterday. 

Thanks for drafting such a great ADR!

PR #16, opened 2026-04-01
[FEAT] vLLM Responses API proxy gateway

maralbahari:
## Summery

Implements the MVP gateway for agentic-stack: a FastAPI app that proxies `POST /v1/responses`
directly to an upstream vLLM server.

### What's in this PR

- `POST /v1/responses` endpoint that forwards requests to vLLM via httpx, with SSE
  passthrough and hop-by-hop header stripping
- `vllm serve <model> --agentic-stack` CLI — starts vLLM as a subprocess and the gateway
  in the foreground (addresses Q10 from ADR-01)
- `agentic-stack --llm-api-base <url>` CLI for standalone mode when vLLM is already running
- Upstream readiness polling before the gateway starts accepting traffic

### What's not in this PR (ADR-01 scope)

- No response store or `previous_response_id` rehydration
- No SSE normalization — raw bytes from vLLM are forwarded as-is
- No tools, MCP, or DB backend

### Implementation Detail:

**Layer 1 - Entry Points (AFR-01 §4)**
- `entrypoints/app.py` — `create_app()` assembles the FastAPI app with lifespan,
  CORS middleware, and the responses router.
- `entrypoints/serve.py` — polls `GET /v1/models` until the upstream is ready, then
  starts uvicorn.
- `entrypoints/cli.py` — `agentic-stack` console script for standalone mode
  (`--llm-api-base` required).
- `entrypoints/vllm_cli.py` — `vllm` console script that intercepts
  `vllm serve <model> --agentic-stack`, spawns vLLM as a subprocess, and runs the
  gateway in the foreground.

**Layer 2 — HTTP Routing** (ADR-01 §4)
- `routers/responses.py` — single `POST /v1/responses` endpoint. Implements D1
  (upstream interface is vLLM's Responses API, not Chat Completions or `llm.generate()`)
  and the endpoint portion of D4 (MVP single endpoint).

**Layer 3 — Core Orchestration (upstream forwarding slice only)** (ADR-01 §4)
- `core/proxy.py` — `proxy_responses()` covers the upstream LLM call portion of the
  orchestration layer: forwards requests to `{upstream}/v1/responses` via httpx.

**Layer 6 — Config** (ADR-01 §4)
- `config/runtime.py` — `RuntimeConfig` pydantic model holding upstream URL, API
  key, gateway bind address/port, and upstream readiness poll settings.

## Test Plan

```bash
uv run pytest
```

Tests cover non-streaming and streaming passthrough, header stripping, auth injection,
error passthrough, mid-stream failures, and the vllm CLI intercept logic. All in-process —
no GPU or real vLLM needed.

## Test Results

```
configfile: pyproject.toml
plugins: anyio-4.13.0
collecting ... collected 24 items

tests/test_proxy.py::test_proxy_responses_non_stream_passthrough[asyncio] PASSED [  4%]
tests/test_proxy.py::test_proxy_responses_stream_passthrough[asyncio] PASSED [  8%]
tests/test_proxy.py::test_proxy_hop_by_hop_headers_stripped[asyncio] PASSED [ 12%]
tests/test_proxy.py::test_proxy_authorization_env_key_injected[asyncio] PASSED [ 16%]
tests/test_proxy.py::test_proxy_authorization_client_header_takes_precedence[asyncio] PASSED [ 20%]
tests/test_proxy.py::test_proxy_upstream_http_error_passthrough[asyncio] PASSED [ 25%]
tests/test_proxy.py::test_proxy_stream_mid_stream_failure_closes_without_synthetic_events[asyncio] PASSED [ 29%]
tests/test_proxy.py::test_proxy_connect_error_maps_to_502[asyncio] PASSED [ 33%]
tests/test_proxy.py::test_proxy_timeout_maps_to_504[asyncio] PASSED      [ 37%]
tests/test_proxy.py::test_proxy_responses_non_stream_passthrough[trio] PASSED [ 41%]
tests/test_proxy.py::test_proxy_responses_stream_passthrough[trio] PASSED [ 45%]
tests/test_proxy.py::test_proxy_hop_by_hop_headers_stripped[trio] PASSED [ 50%]
tests/test_proxy.py::test_proxy_authorization_env_key_injected[trio] PASSED [ 54%]
tests/test_proxy.py::test_proxy_authorization_client_header_takes_precedence[trio] PASSED [ 58%]
tests/test_proxy.py::test_proxy_upstream_http_error_passthrough[trio] PASSED [ 62%]
tests/test_proxy.py::test_proxy_stream_mid_stream_failure_closes_without_synthetic_events[trio] PASSED [ 66%]
tests/test_proxy.py::test_proxy_connect_error_maps_to_502[trio] PASSED   [ 70%]
tests/test_proxy.py::test_proxy_timeout_maps_to_504[trio] PASSED         [ 75%]
======================== 24 passed, 7 warnings in 0.07s ========================

### CODE REVIEW

**Comment by leseb** [comment]
@maralbahari PTAL https://github.com/vllm-project/agentic-api/pull/17

File: src/agentic_api/config/runtime.py [hunk 1]
```diff
@@ -0,0 +1,20 @@
+from dataclasses import dataclass
+from typing import Literal
+
+RuntimeMode = Literal["standalone", "integrated"]
+
+
+@dataclass(frozen=True, slots=True)
+class RuntimeConfig:
+    # Upstream vLLM server
+    llm_api_base: str
```
leseb: please use Pydantic Field with a `description`

tests/test_vllm_cli.py
```diff
@@ -0,0 +1,108 @@
+import pytest
```
leseb: These are all testing that argparse does what argparse does, not sure sure if that's relevant, let's remove?

src/agentic_api/entrypoints/vllm_cli.py
```diff
@@ -0,0 +1,133 @@
+import argparse
```
leseb, line 1: i think this wrapper is a bit premature and the ADR was blurry on that, i'd postpone that but others might disagree.

src/agentic_api/entrypoints/serve.py
```diff
@@ -0,0 +1,55 @@
...
+import uvicorn
+
+from agentic_stack.config.runtime import RuntimeConfig
+from agentic_stack.entrypoints.app import create_app
+
+
+def _wait_upstream_ready(runtime_config: RuntimeConfig) -> None:
+    """Poll vLLM /v1/models until it responds 200 or timeout is reached."""
+    base = runtime_config.llm_api_base.rstrip("/")
+    url = f"{base}/v1/models"
```
leseb: vllm has `/health`, should we use this instead?

src/agentic_api/entrypoints/serve.py
```diff
@@ -0,0 +1,55 @@
...
+    interval_s = runtime_config.upstream_ready_interval_s
+    start = time.perf_counter()
+    last_notice = 0.0
+
+    with httpx.Client(timeout=httpx.Timeout(2.0), headers=headers) as client:
+        while True:
+            elapsed = time.perf_counter() - start
+            if elapsed > timeout_s:
+                raise TimeoutError(
+                    f"upstream did not become ready within {timeout_s:.0f}s: {url}"
```
leseb:
```suggestion
                    f"vLLM did not become ready within {timeout_s:.0f}s: {url}"
```

src/agentic_api/entrypoints/serve.py
```diff
@@ -0,0 +1,55 @@
...
+            if elapsed - last_notice >= interval_s:
+                last_notice = elapsed
+                print(f"[serve] waiting for upstream ({elapsed:.0f}s elapsed): {url}")
+
+            time.sleep(interval_s)
+
+
+def run(runtime_config: RuntimeConfig) -> None:
+    _wait_upstream_ready(runtime_config)
+    print(f"[serve] upstream ready: {runtime_config.llm_api_base}")
```
leseb: use a logger

src/agentic_api/entrypoints/app.py
```diff
@@ -0,0 +1,31 @@
...
+        title="Agentic Stack",
+        lifespan=lifespan,
+    )
+    app.add_middleware(
+        CORSMiddleware,
+        allow_origins=["*"],
+        allow_credentials=True,
+        allow_methods=["*"],
+        allow_headers=["*"],
+    )
```
leseb, line 29: Why do we need CORS?

src/agentic_api/entrypoints/cli.py
```diff
@@ -0,0 +1,59 @@
+import argparse
+import sys
+
+from agentic_stack.config.runtime import RuntimeConfig
+from agentic_stack.entrypoints.serve import run
+
+
+def _normalize_base_url(url: str) -> str:
```
leseb, line 8: use urlparse? i believe this will take care of some of that

src/agentic_api/entrypoints/cli.py
```diff
@@ -0,0 +1,59 @@
...
+        prog="agentic-stack",
+        description="Start the agentic-stack gateway in front of a running vLLM server.",
+    )
+    parser.add_argument(
+        "--llm-api-base",
+        required=True,
+        help="Base URL of the upstream vLLM server (e.g. http://127.0.0.1:8000). /v1 suffix is optional.",
+    )
+    parser.add_argument(
+        "--openai-api-key", default=None, help="API key forwarded to upstream."
```
leseb, line 27: why do we need this?

src/agentic_api/entrypoints/cli.py
```diff
@@ -0,0 +1,59 @@
...
+    )
+    parser.add_argument("--gateway-host", default="0.0.0.0")
+    parser.add_argument("--gateway-port", type=int, default=9000)
+    parser.add_argument("--gateway-workers", type=int, default=1)
+    parser.add_argument(
+        "--upstream-ready-timeout",
```
leseb, line 33: let's re-use `llm-api-base`

src/agentic_api/core/proxy.py
```diff
@@ -0,0 +1,171 @@
...
+        if self._stream_client is not None:
+            await self._stream_client.aclose()
+            self._stream_client = None
+        if self._non_stream_client is not None:
+            await self._non_stream_client.aclose()
+            self._non_stream_client = None
+
+
+def proxy_error(*, status_code: int, code: str, message: str) -> ORJSONResponse:
+    return ORJSONResponse(
```
leseb: please use JSONResponse from fastapi

pyproject.toml
```diff
@@ -20,7 +20,23 @@ classifiers = [
     "Programming Language :: Python :: 3.12",
     "Programming Language :: Python :: 3.13",
 ]
-dependencies = []
+dependencies = [
+    "fastapi>=0.115.0",
+    "uvicorn[standard]>=0.30.0",
+    "httpx>=0.27.0",
+    "pydantic>=2.7.0",
+    "orjson>=3.10.0",
```
leseb: please remove

leseb (changes requested):
Thanks for the PR, next time could you please use your fork instead of pushing directly in this remote?

tjtanaa:
> Thanks for the PR, next time could you please use your fork instead of pushing directly in this remote?

Cool, shall we all set a protocol to develop in our own fork branch to keep the repo branch list cleaner? On vLLM there are tons of development branches >. <

src/agentic_api/entrypoints/cli.py
tjtanaa: vLLM supports the feature to set API Key to block unauthorize requests.

src/agentic_api/entrypoints/vllm_cli.py
tjtanaa: It has been decided to retain the `vllm serve` command so it will in the form of adding a argument to setup and expose the API endpoints from agentic-api. I am open to the name of the argument.

`vllm serve <model> --agentic` so that user experience is not affected.
making it as easy as like on vllm omni, the command is `vllm serve <model> [arg] --omni`

tests/test_vllm_cli.py
maralbahari: @leseb alright so remove the whole test_vllm_cli.py?

src/agentic_api/entrypoints/serve.py
maralbahari: @leseb sure definitely ..` /health`  makes more sense

tests/test_vllm_cli.py
leseb: yes

pyproject.toml
maralbahari: @leseb for dependencies want to keep a `requirement.txt` if they're removed from here?

src/agentic_api/entrypoints/cli.py
leseb: Oh ok always thought the key was a noop, thanks!

leseb:
> > Thanks for the PR, next time could you please use your fork instead of pushing directly in this remote?
> 
> Cool, shall we all set a protocol to develop in our own fork branch to keep the repo branch list cleaner? On vLLM there are tons of development branches >. <

yeah i'd be in favor of that!
i think doing devs on fork help reduce clutter on the release branches on the upstream remote.

src/agentic_api/entrypoints/cli.py
maralbahari: @leseb could you elaborate on re-use `llm-api-base` ? the `llm-api-base` suppose to be url that vLLM is served by user separately with `vllm serve ...` command then the url `http://localhost:<port>` is passed to `agentic-api `

pyproject.toml
jiahuei: > please remove

why

leseb: we don't need that extra `orjson` we can use `JSONResponse` from fastapi instead

src/agentic_api/entrypoints/cli.py
leseb: sorry, sometimes we refer vllm as `llm-api-base` sometimes as `upstream-`, we should consolidate.

pyproject.toml
maralbahari: @leseb alright so just the `orjson` ... cause the whole dependency block is commented. thought you mean remove all!

jiahuei: > we don't need that extra `orjson` we can use `JSONResponse` from fastapi instead

Thanks for the explanation. Using `ORJSONResponse` was a recommended pattern by FastAPI since `orjson` is more performant. But now the recommendation is to always return Pydantic types (which uses Rust) so I think we should change to that pattern instead.

src/agentic_api/entrypoints/vllm_cli.py
leseb: oh no `argparse` is fine, it's the CLI itself - i guess i'm not a big fan of embedding the vllm process inside our `agentic-api`, it feels like this opens the door for bugs, debugging issue, signal mgmt etc. also this `vllm` CLI actually overrides the system's one right? that doesn't seem like the good idea.

The standalone mode (agentic-stack --llm-api-base http://...) is clean and sufficient for MVP. The integrated mode is solving a convenience problem that creates real operational problems.

src/agentic_api/entrypoints/cli.py
maralbahari: @leseb the `urlparse` returns a `ParseResult` object would need to use `urlunparse`  to return back to `str`. since user can either pass `http://host:8000` or `http://host:8000/v1` yet to striping logic remain the same in terms of line of code does not help. working with `str` is easy and simple in this case.

leseb: ok cool then!

src/agentic_api/entrypoints/vllm_cli.py
maralbahari: @leseb yes I thought it has been decided that we would want to allow user to either run `vllm serve <model> --agentic-api`

src/agentic_api/entrypoints/app.py
maralbahari: It was added as a convenience for browser-based clients so that they don't get blocked by the same-origin policy. With allow_origins=["*"] it permits any origin. it is also used in vllm  `vllm/vllm/entrypoints/openai/api_server.py` https://github.com/EmbeddedLLM/vllm/blob/58ee61422169ce17e08248f8efa1e9df434fe395/vllm/entrypoints/openai/api_server.py#L257. 
if you think it is not necessary for at this moment it can be omitted.

maralbahari:
@leseb addressed the changes you have requested. kindly review again.

src/agentic_api/entrypoints/app.py
leseb: ok it's fine i guess

src/agentic_api/entrypoints/cli.py
leseb: not addressed? let me know if this wasn't clear

src/agentic_api/entrypoints/vllm_cli.py
leseb: if there is consensus then i'm fine being the only one pushing back for this

leseb (approved):
this is good enough for now thank you, i'm still -1 on the vllm command override though but it seems that others agreed

src/agentic_api/core/proxy.py
```diff
@@ -0,0 +1,171 @@
...
+    runtime_config: RuntimeConfig,
+    proxy_client_manager: ProxyClientManager,
+) -> Response:
+    body = await request.body()
+    upstream_headers = _filter_request_headers(
+        request.headers.items(),
+        runtime_config=runtime_config,
+    )
+    upstream_params = list(request.query_params.multi_items())
+    client = proxy_client_manager.get_client(allow_sse_passthrough=True)
```
franciscojavierarceo: Why is `allow_sse_passthrough` hardcoded to True? 

I believe this means non-streaming requests also use the stream client (`read=None` timeout). Should this inspect the request body's "stream" field so non-streaming requests get a read timeout?

src/agentic_api/entrypoints/vllm_cli.py
franciscojavierarceo: I'm good with this

franciscojavierarceo (approved):
one small nit/question but this lgtm!

src/agentic_api/entrypoints/cli.py
maralbahari: @leseb what is the change you are exactly having in mind? and what do you mean by re-use `llm-api-bas` ?

src/agentic_api/core/proxy.py
maralbahari: @franciscojavierarceo thank you for pointing out. I have pushed a commit that reads the body of request to determine the stream and non-stream request.

PR #20, opened 2026-04-20
ADR-03 for Responses tool interface boundaries

noobHappylife:
## Summary

Adds ADR-03 as a discussion-oriented ADR for Responses API tool interface boundaries, especially gateway-owned and server-side executed tools.

The ADR is structured around four tool interface categories:

- Built-in tools: stable gateway-owned tools with model-facing contracts, such as `code_interpreter`, `web_search`, and `file_search`.

- Dynamic model-facing toolsets: discovered tool inventories exposed directly to the model, such as hosted MCP or request-declared remote MCP.

- Backend adapters: hidden backend-facing interfaces used by built-in tools when the model-facing contract should stay stable while the execution backend can vary.

- Shared runtime clients: shared lifecycle/status/inventory clients for managed MCP backends, kept separate from per-request handling.

This is a draft. Comments, suggestions, and holes to poke are welcome.

### DIFF
--- a/docs/adr/ADR-03_gateway_owned_tool_interfaces.md
+++ b/docs/adr/ADR-03_gateway_owned_tool_interfaces.md
@@ -1,0 +1,285 @@
+# ADR-03: Gateway-owned tool interface contracts
+
+> **Status:** Draft
+
+## Intention
+
+Define the gateway-owned tool interface contracts for `agentic-apis`, so new Responses API tools can be added
+without coupling each tool directly to model transport, provider-specific tool calling, or the public streaming
+composer.
+
+This ADR focuses on the contract shape for building tools:
+
+- which tool interface types the gateway should support
+- what each interface is good for
+- what trade-offs each interface creates
+- what minimum work is required for `code_interpreter`, `web_search`, hosted MCP, remote MCP, and `file_search`
+
+The goal is to extract the durable decision points needed to start implementing tools with stable boundaries.
+
+## Context
+
+Responses API tools are not all the same kind of thing.
+
+Some tools are stable product features owned by the gateway, such as `code_interpreter` or `web_search`. The model
+should see a stable tool contract even if the gateway changes the backend implementation later.
+
+Other tools are dynamic external inventories, such as MCP tools discovered from a hosted or request-declared server.
+In those cases, the model may need to see the discovered tool names, descriptions, and JSON schemas directly.
+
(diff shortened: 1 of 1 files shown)


docs/adr/ADR-03_gateway_owned_tool_interfaces.md
```diff
@@ -0,0 +1,303 @@
...
+Define the gateway-owned tool interface contracts for `agentic-apis`, so new Responses API tools can be added
+without coupling each tool directly to model transport, provider-specific tool calling, or the public streaming
+composer.
+
+This ADR focuses on the contract shape for building tools:
+
+- which tool interface types the gateway should support
+- what each interface is good for
+- what trade-offs each interface creates
+- what minimum work is required for `code_interpreter`, `web_search`, hosted MCP, remote MCP, and `file_search`
```
leseb, line 16: given `file_search` implies vectors and files, could we postpone it? technically, we could handle all tools separately. like start with web_search, then code_interpreter and finally file_search?

noobHappylife: Yes thank you for pointing this out. The minimum work is more like the eventual tools that this repo will have. Will update this.

franciscojavierarceo: FWIW I put a demo together of that in https://github.com/vllm-project/agentic-api/pull/22

And I'll leave the comment I made there: 

>Note: This is a draft — we may instead integrate OGX's existing file_search implementation rather than maintaining a separate one here. OGX already implements prompts, conversations, and file_search with a battle-tested retrieval stack.

OGX also supports MCP and I believe has support for `web_search` (though I haven't tested it much myself)

franciscojavierarceo: FWIW, if we decide not to go with OGX, the surface area of a `file_search` tool with feature/performance parity is quite large (which you can look through the OGX code to confirm the magnitude).

franciscojavierarceo, line 16: `code_interpret` feels like a big rabbit hole, that I do want to invest time on coming with a thoughtful proposal because it has broad security implications that we would need to be thoughtful about.

franciscojavierarceo: The `web_search` tool seems manageable though.

noobHappylife:
please see the #20

PR #24, opened 2026-05-11
feat: implement Rust proxy gateway for vLLM Responses API

leseb:
## Summary

- Implements the full Rust proxy gateway for the vLLM Responses API, replacing the Python gateway stub
- Supports both streaming (SSE) and non-streaming proxy modes with proper hop-by-hop header filtering
- Includes API key injection from environment, comprehensive error mapping (502 for connection errors, 504 for timeouts), and CORS support
- Adds CLI with two modes: standalone (`--llm-api-base`) and integrated (`serve <model>` spawning vLLM as a subprocess)
- Modular architecture: `config`, `proxy`, `app`, `server` modules with a clean separation of concerns

## Test Plan

- 11 tests covering: non-stream passthrough, stream passthrough, hop-by-hop header stripping, auth injection, client auth precedence, upstream HTTP error passthrough, mid-stream failure handling, connection error → 502, timeout → 504
- All tests pass: `cargo test` (11/11 green)
- Clippy clean: `cargo clippy --all-targets -- -D warnings`
- Formatting clean: `cargo fmt -- --check`


src/main.rs
```diff
@@ -1,11 +1,104 @@
-fn main() {
-    println!("agentic-api");
+use clap::{Args, Parser, Subcommand};
+
+use agentic_api::config::{RuntimeConfig, normalize_base_url};
+use agentic_api::server;
+
+#[derive(Args, Clone)]
+struct GatewayOpts {
```
maralbahari: On src/config.rs + src/main.rs:

`GatewayOpts`  in main.rs duplicates every field of RuntimeConfig in config.rs, and build_config is a manual field-by-field copy between them. When a new config field is added, it has to be added in three places — the struct, the CLI args, and the mapping function.

Since RuntimeConfig and the CLI args map 1-to-1, we can collapse this by deriving Parser directly on RuntimeConfig and using #[arg(skip)] on llm_api_base (the only field that doesn't come from CLI). Then a single RuntimeConfig::from_cli(base_url) constructor replaces both `GatewayOpts` and `build_config`:

```rust
// config.rs
#[derive(Debug, Clone, Parser)]
pub struct RuntimeConfig {
    #[arg(skip)]
    pub llm_api_base: String,

    #[arg(long, env = "OPENAI_API_KEY")]
    pub openai_api_key: Option<String>,

    #[arg(long, default_value = "0.0.0.0")]
    pub gateway_host: String,

    #[arg(long, default_value_t = 9000)]
    pub gateway_port: u16,

    #[arg(long, default_value_t = 600.0)]
    pub upstream_ready_timeout_s: f64,

    #[arg(long, default_value_t = 2.0)]
    pub upstream_ready_interval_s: f64,
}

impl RuntimeConfig {
    pub fn from_cli(llm_api_base: &str) -> Self {
        let mut config = Self::parse();
        config.llm_api_base = normalize_base_url(llm_api_base);
        config
    }
}

// main.rs — GatewayOpts and build_config gone entirely
let config = RuntimeConfig::from_cli(&base);
```

src/config.rs
```diff
@@ -0,0 +1,36 @@
+#[derive(Debug, Clone)]
```
maralbahari: converting `src/config.rs`  into a proper submodule folder. the same pattern as Python packages with `__init__.py`:

```
src/config.rs          →   src/config/mod.rs      (the __init__.py equivalent)
                            src/config/runtime.rs  (RuntimeConfig lives here)

Comment (leseb) [reply src/config.rs]: This is what Claude told me:

> This is a style preference. In idiomatic Rust, flat module files (src/config.rs) are standard and preferred for small modules — the Rust community moved away from mod.rs in the 2018 edition. The mod.rs approach is fine but adds indirection for a single-struct module. This might make more sense later when config grows to contain multiple sub-modules.

do you still want to do this?

File: src/main.rs (hunk 1 above)
Comment (leseb) [reply src/main.rs]: 048a4a51002682e901dc80c414c7b29eb702ef6f

File: src/config.rs (hunk 2 above)
Comment (maralbahari) [reply src/config.rs]: sure we can keep it as is.

**Comment by maralbahari** [comment]
@leseb I was wondering if we should validate that the library choices are optimal and verify stateless proxy throughput/latency against the Python baseline. When I was working on translating Python to Rust, I noticed some performance regression compared to the Python baseline, so I thought it would be best to try and avoid such regressions from the start.

**Comment by leseb** [comment]
> @leseb I was wondering if we should validate that the library choices are optimal and verify stateless proxy throughput/latency against the Python baseline. When I was working on translating Python to Rust, I noticed some performance regression compared to the Python baseline, so I thought it would be best to try and avoid such regressions from the start.

Can you share more? I'll add a criterion benchmarks

**Comment by leseb** [comment]
@maralbahari any thoughts here?

**Review by franciscojavierarceo — APPROVED** [review]

File: tests/common/mod.rs [hunk 3]
```diff
@@ -0,0 +1,207 @@
...
+use axum::routing::{get, post};
+use bytes::Bytes;
+use futures::stream;
+use http::StatusCode;
+use tokio::net::TcpListener;
+
+use agentic_api::config::RuntimeConfig;
+use agentic_api::proxy::ProxyState;
+
+pub fn test_config(vllm_url: &str) -> RuntimeConfig {
```
maralbahari: after reading the official rust documentation examples. The isolated unit tests that are verifying a single module in this case the `RuntimeConfig struct` in config we should implement the test into the same file as source code in `src/config.rs`  so the test_config_no_key() and test_config cases should move there. you can look at the example [here from documentation](https://doc.rust-lang.org/stable/rust-by-example/testing/unit_testing.html).

while the `tests/` folder is used for [integration tests ](https://doc.rust-lang.org/stable/rust-by-example/testing/integration_testing.html)  where you would need to test multiple modules relying on each other for an integration case. and `test/common/mod.rs` use to keep the common setup in this file 
like make_runtime_config() make_response_handler(). but I think for this proxy feature we don't need a integration tests as proxy is the single module.

leseb: fair point, i'll remove the integration tests since it's too much of mocking anyways and not using a real vllm server, i'll add more unit tests

leseb: done in https://github.com/vllm-project/agentic-api/pull/24/commits/a4080864aaec446af9a73b2c2e77723bae585864

leseb:
Test plan:

Run vLLM:

```
podman run -d --name vllm \
    -p 8000:8000 \
    -v ~/.cache:/root/.cache \
    -e VLLM_CPU_KVCACHE_SPACE=2 \
    public.ecr.aws/q9t5s3a7/vllm-arm64-cpu-release-repo:latest \
    --host 0.0.0.0 \
    --model Qwen/Qwen3-0.6B \
    --max-model-len 4096 \
    --served-model-name Qwen/Qwen3-0.6B \
    --enable-auto-tool-choice \
    --tool-call-parser hermes
```

Check it's healthy:

```
curl -s http://localhost:8000/health 2>&1
```

Check status code:

```
curl -s -o /dev/null -w '%{http_code}' http://localhost:8000/health 2>&1
200
```

Build the project:

 ```
cargo build --release
```

Run it:

 ```
./target/release/agentic-api --llm-api-base http://localhost:8000
2026-05-26T09:35:20.424460Z  INFO agentic_api::server: vLLM ready: http://localhost:8000
2026-05-26T09:35:20.425334Z  INFO agentic_api::server: gateway listening on 0.0.0.0:9000
```

Test a request:

```
curl -s http://localhost:9000/v1/responses \
  -H "Content-Type: application/json" \
  -d '{"model": "Qwen/Qwen3-0.6B", "input": [{"role": "user", "content": "Count to 3"}], "stream": false}'

{"id":"resp_be8ba46fc00a1a27","created_at":1779788351,"incomplete_details":null,"instructions":null,"metadata":null,"model":"Qwen/Qwen3-0.6B","object":"response","output":[{"id":"msg_9cc874b6892a835a","content":[{"annotations":[],"text":"<think>\nOkay, the user is asking to count to 3. Let me think about how to approach this. First, I need to make sure I understand the request correctly. Counting to 3 is straightforward, but maybe the user is looking for a specific method or a visual representation. Since the user hasn't specified any additional context, I should provide a simple and clear answer.\n\nI should start by explaining what counting to 3 means. It's the sequence 1, 2, 3. Then, I can list the numbers in order. It's important to confirm that the answer is direct and doesn't require any extra information. Also, maybe mention that counting helps in developing number sense. But since the user's query is very direct, keeping it concise is better. Let me check if there's any possible misunderstanding. No, counting to 3 is just 1, 2, 3. So the answer should be straightforward.\n</think>\n\nCount to 3: 1, 2, 3.","type":"output_text","logprobs":null}],"role":"assistant","status":"completed","type":"message"}],"parallel_tool_calls":true,"temperature":0.6,"tool_choice":"auto","tools":[],"top_p":0.95,"background":false,"max_output_tokens":4084,"max_tool_calls":null,"previous_response_id":null,"prompt":null,"reasoning":null,"service_tier":"auto","status":"completed","text":null,"top_logprobs":null,"truncation":"disabled","usage":{"input_tokens":12,"input_tokens_details":{"cached_tokens":0,"input_tokens_per_turn":[],"cached_tokens_per_turn":[]},"output_tokens":210,"output_tokens_details":{"reasoning_tokens":0,"tool_output_tokens":0,"output_tokens_per_turn":[],"tool_output_tokens_per_turn":[]},"total_tokens":222},"user":null,"input_messages":null,"output_messages":null}%
```

PR #25, opened 2026-05-12
[FEAT] Adds the database configuration and schema management.

maralbahari:
## Summary

Adds the database configuration, schema management, and CRUD layer for the three-table persistence schema used by the Responses and Conversation APIs.

### What's in this PR

- **`RuntimeConfig`**:  single struct deriving both `clap::Parser` and `validator::Validate`; CLI args and field validation in one place with no `GatewayOpts` duplication
- **`DbPool` via `sqlx::Any`**:  single pool type that works with both SQLite and PostgreSQL at runtime; no per-dialect code duplication
- **Three-table schema**: `conversations`, `items`, `responses` tables matching the Python schema exactly; managed via `sqlx` migrations in `migrations/0001_initial.sql`
- **`SchemaManager`**:  runs migrations on startup; respects `AA_DB_SCHEMA_READY` env var so worker processes skip DDL; uses sqlx's built-in migration lock replacing Python's PostgreSQL advisory lock
- **CRUD operations** : `database/conversation.rs`, `database/item.rs`, `database/response.rs`.
- 

## Test Plan

> **Note:** `src/bin/` does not exist in this branch. To verify schema initialisation, create it temporarily:
>
> ```bash
> mkdir -p src/bin
> ```
>
> Then add `src/bin/init_db.rs`:
>
> ```rust
> use agentic_api::database::db::create_pool;
> use agentic_api::database::schema::SchemaManager;
>
> #[tokio::main]
> async fn main() -> Result<(), Box<dyn std::error::Error>> {
>     let db_url = std::env::args().nth(1)
>         .unwrap_or_else(|| "sqlite://./agentic_api.db".to_string());
>     let pool = create_pool(&db_url).await?;
>     SchemaManager::new(&pool).ensure_ready().await?;
>     println!("Schema ready.");
>     Ok(())
> }
> ```

```bash
# Schema verification
cargo run --bin init_db
cargo run --bin init_db -- "sqlite:///path/to/db"

# Check tables
sqlite3 agentic_api.db "SELECT name FROM sqlite_master WHERE type='table' ORDER BY name;"
# Expected: _sqlx_migrations, conversations, items, responses

# Full build must pass
cargo build
```

## Test Results

<details>
<summary>Both SQLite paths verified — default and explicit</summary>

```
==> Test 1: default DB (no argument)...
     Running `target/debug/init_db`
Connecting to: sqlite://./agentic_api.db
Running schema migrations...
Schema ready. Tables created successfully.

==> Tables in agentic_api.db...
_sqlx_migrations
conversations
items
responses

==> Test 2: explicit DB path...
     Running `target/debug/init_db 'sqlite:///home/tunjitan/ellmm/test_agentic_api.db'`
Connecting to: sqlite:///home/tunjitan/ellmm/test_agentic_api.db
Running schema migrations...
Schema ready. Tables created successfully.

==> Tables in test_agentic_api.db...
_sqlx_migrations
conversations
items
responses
```

</details>


migrations/0001_initial.sql
```diff
@@ -0,0 +1,31 @@
+CREATE TABLE IF NOT EXISTS conversations (
```
franciscojavierarceo, line 1: conversations can be a bit of a lot. Maybe we should start off narrow with just the rust server for responses models and use ogx as an MVP? That way we get all of the API surface area without having to maintain that implementation right away? We can discuss tomorrow of course.

CC @leseb @jiahuei

maralbahari: @franciscojavierarceo yes I agree we can focus on responses first. less complexity

PR #27, opened 2026-05-15
feat: migrate to Praxis filter-based proxy architecture

franciscojavierarceo:
## Summary

Replaces the hand-rolled Axum proxy with [Praxis](https://github.com/praxis-proxy/praxis), a composable filter-based reverse proxy framework built on [Pingora](https://github.com/cloudflare/pingora). Each gateway concern — proxying, auth, state hydration, tool dispatch, agentic looping — is an independent filter wired together via YAML configuration.

### Why Praxis

- **Each filter is self-contained** — implements `HttpFilter` with hooks for request/response. Filters don't know about each other.
- **YAML-configured pipeline** — adding, removing, or reordering filters requires no code changes.
- **Native SSE streaming** — Praxis/Pingora proxies the upstream response stream directly to the client. No buffering, no reqwest intermediary.
- **Hot reload** — filter pipelines can be reloaded without restarting the server.

### What changed

**Filters introduced:**
- `responses_proxy` — sets `ctx.upstream` to vLLM's `/v1/responses` endpoint and injects auth credentials. Praxis/Pingora handles the actual proxying and streaming natively.
- `state_hydration` — stub filter for conversation-state hydration. Inspects request body for `previous_response_id` and will call the state store to hydrate conversation history.
- `agentic_loop` — stub filter for agentic re-inference. Inspects response body for `function_call` output items and will re-enter the inference loop.
- `tool_dispatch` — stub filter for tool execution. Inspects response body for tool calls and will dispatch them.

**Removed:**
- `src/app.rs`, `src/proxy.rs`, `src/server.rs` — replaced by filters + Praxis server runtime
- `benches/proxy_bench.rs` — benchmark harness for the old Axum proxy (will be re-added)

**Dependencies:**
- Praxis crates (`praxis`, `praxis-proxy-core`, `praxis-proxy-filter`, `praxis-test-utils`) via git at rev `2f7ea31`
- Base URLs ending with `/v1` are normalized to avoid `/v1/v1/responses` double-prefix

**Docs:**
- Updated `README.md` with architecture diagram, filter table, and run instructions
- Updated `docs/index.md` with architecture overview and Praxis context
- Added `docs/architecture/index.md` with Mermaid diagram, filter pipeline reference, streaming details, and component descriptions
- Added Architecture page to mkdocs nav

**Health endpoint:**
- Provided by Praxis's built-in admin endpoint (`admin: { address: "127.0.0.1:9901" }` in config)

### Filter pipeline

```yaml
filter_chains:
  - name: agentic
    filters:
      - filter: state_hydration
        store_base_url: "http://localhost:8080"
      - filter: agentic_loop
        max_iterations: 10
      - filter: tool_dispatch
      - filter: responses_proxy
        vllm_base_url: "http://localhost:8000"
```

## Test plan

- [x] `cargo build` succeeds
- [x] `cargo clippy --all-targets -- -D warnings` clean
- [x] `cargo fmt -- --check` clean
- [x] `pre-commit run --all-files` clean
- [x] All 9 tests pass (2 unit + 7 integration):
  - `test_non_stream_passthrough` — JSON request/response round-trip
  - `test_stream_passthrough` — SSE streaming passthrough
  - `test_auth_injection` — API key injected from config
  - `test_client_auth_precedence` — client-supplied auth preserved
  - `test_vllm_http_error_passthrough` — upstream 429 forwarded
  - `test_mid_stream_failure_closes_cleanly` — partial stream handled
  - `test_connect_error_maps_to_502` — unreachable vLLM returns 502

### DIFF
--- a/config/agentic-api.yaml
+++ b/config/agentic-api.yaml
@@ -1,0 +1,22 @@
+admin:
+  address: "127.0.0.1:9901"
+
+listeners:
+  - name: agentic-api
+    address: "0.0.0.0:9000"
+    filter_chains: [agentic]
+
+filter_chains:
+  - name: agentic
+    filters:
+      - filter: state_hydration
+        store_base_url: "http://localhost:8080"
+      - filter: agentic_loop
+        max_iterations: 10
+      - filter: tool_dispatch
+      - filter: responses_proxy
+        vllm_base_url: "http://localhost:8000"
+
+clusters:
+  - name: vllm
+    endpoints: ["127.0.0.1:8000"]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,5 +3,3906 @@
 version = 4
 
 [[package]]
+name = "adler2"
+version = "2.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
+
+[[package]]
 name = "agentic-api"
 version = "0.1.0"
+dependencies = [
+ "async-trait",
+ "axum",
+ "bytes",
+ "clap",
+ "futures",
+ "http",
+ "praxis",
+ "praxis-proxy-core",
+ "praxis-proxy-filter",
+ "praxis-test-utils",
+ "reqwest",
+ "serde",
+ "serde_json",
+ "serde_yaml",
+ "thiserror 2.0.18",
+ "tokio",
+ "tokio-stream",
+ "tracing",
+]
+
+[[package]]
+name = "ahash"
+version = "0.8.12"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5a15f179cd60c4584b8a8c596927aadc462e27f2ca70c04e0071964a73ba7a75"
+dependencies = [
+ "cfg-if",
+ "const-random",
+ "getrandom 0.3.4",
+ "once_cell",
+ "version_check",
+ "zerocopy",
+]
+
+[[package]]
+name = "aho-corasick"
+version = "1.1.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ddd31a130427c27518df266943a5308ed92d4b226cc639f5a8f1002816174301"
+dependencies = [
+ "memchr",
+]
+
+[[package]]
+name = "aliasable"
+version = "0.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "250f629c0161ad8107cf89319e990051fae62832fd343083bea452d93e2205fd"
+
+[[package]]
+name = "alloc-no-stdlib"
+version = "2.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cc7bb162ec39d46ab1ca8c77bf72e890535becd1751bb45f64c597edb4c8c6b3"
+
+[[package]]
+name = "alloc-stdlib"
+version = "0.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "94fb8275041c72129eb51b7d0322c29b8387a0386127718b096429201a5d6ece"
+dependencies = [
+ "alloc-no-stdlib",
+]
+
+[[package]]
+name = "allocator-api2"
+version = "0.2.21"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
+
+[[package]]
+name = "anstream"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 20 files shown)


config/agentic-api.yaml
```diff
@@ -0,0 +1,22 @@
+admin:
```
maralbahari, line 1: need documentation on what each fields means in this yaml file and what are they used for.
some fields are confusing to guess what it meant to do. like `store_base_url` like related to storage but it's not database url?!

Overall the `Praxis` library is relatively new not sure if it is suitable to rely on it. the maintenance is costly.

it is difficult to review this PR and judge as would require the maintainer on `agentic-ap` to be familiar with `Praxis`.

meanwhile natively writing our own http request gateway would allow us flexibility especially in SSE stream and tool calls.
in terms of testing the `agentic-api` repo functionality as a whole system now it's relying entirely on  Praxis is maintained and tested. What If we encounter bugs from Praxis we would need to wait for bug fixes there. 

I thought based on the last community meeting we would use OGX for CRUD features as it is a well-maintained project?!

noobHappylife:
Thanks for putting this together.

I like the direction of making gateway concerns more composable, and praxis does look like a strong proxy framework. the filter model is useful for auth, rate limit, tenant routing, quota, policy, request validation, header injection, deployment guardrails etc.

But i don’t think praxis should be the core boundary for agentic-api, my main concern is that this moves too much of the actual agentic runtime into proxy filters. things like:

  - previous_response_id rehydration
  - response/message store
  - stable response item ids
  - SSE event ordering/timing
  - tool call detection/execution/result injection
  - multi-turn loop continuation
  - cancellation/timeout/partial failure semantics
  - trajectory/trace capture for RL rollout
  - vLLM-native metadata, cache/session lineage, token/logprob metadata (which is important for RL usecase)

These, I feel, are not really generic proxy concerns. they are the main state machine of agentic-api. Splitting this into filters like state_hydration, agentic_loop, tool_dispatch, responses_proxy may look composable, but in practice they are tightly coupled by shared state, response semantics, and stream ordering. i worry we end up encoding the main transaction as middleware, which is harder to reason about and test.

Praxis can still be useful, in an architecture like:

  client / codex cli / agent harness
    -> praxis gateway, optional
         auth / rate limit / tenant routing / policy / coarse guardrails
    -> agentic-api
         responses state machine, message store, SSE semantics,
         tool loop, tool execution, trajectory capture
    -> vLLM responses or internal vLLM-native backend

So praxis as an outer gateway in front of agentic-api, not where we decompose the agentic loop itself.

And for OGX. I think OGX can be a backend/service provider for built-in tools like file_search, vector stores, files, or other stateful services. but agentic-api should still decide when/how those tools participate in the responses loop.

So my recommendation is: don’t make praxis the core architecture boundary for agentic-api. if we support praxis, I’d rather make it an optional outer gateway integration, or a very thin adapter that delegates into an explicit agentic-api orchestration core.

leseb:
> Thanks for putting this together.
> 
> I like the direction of making gateway concerns more composable, and praxis does look like a strong proxy framework. the filter model is useful for auth, rate limit, tenant routing, quota, policy, request validation, header injection, deployment guardrails etc.
> 
> But i don’t think praxis should be the core boundary for agentic-api, my main concern is that this moves too much of the actual agentic runtime into proxy filters. things like:
> 
> * previous_response_id rehydration
> * response/message store
> * stable response item ids
> * SSE event ordering/timing
> * tool call detection/execution/result injection
> * multi-turn loop continuation
> * cancellation/timeout/partial failure semantics
> * trajectory/trace capture for RL rollout
> * vLLM-native metadata, cache/session lineage, token/logprob metadata (which is important for RL usecase)
> 
> These, I feel, are not really generic proxy concerns. they are the main state machine of agentic-api. Splitting this into filters like state_hydration, agentic_loop, tool_dispatch, responses_proxy may look composable, but in practice they are tightly coupled by shared state, response semantics, and stream ordering. i worry we end up encoding the main transaction as middleware, which is harder to reason about and test.
> 
> Praxis can still be useful, in an architecture like:
> 
> client / codex cli / agent harness -> praxis gateway, optional auth / rate limit / tenant routing / policy / coarse guardrails -> agentic-api responses state machine, message store, SSE semantics, tool loop, tool execution, trajectory capture -> vLLM responses or internal vLLM-native backend
> 
> So praxis as an outer gateway in front of agentic-api, not where we decompose the agentic loop itself.
> 
> And for OGX. I think OGX can be a backend/service provider for built-in tools like file_search, vector stores, files, or other stateful services. but agentic-api should still decide when/how those tools participate in the responses loop.
> 
> So my recommendation is: don’t make praxis the core architecture boundary for agentic-api. if we support praxis, I’d rather make it an optional outer gateway integration, or a very thin adapter that delegates into an explicit agentic-api orchestration core.

All great points, i'm proposing a design that should align all parties so let's discuss over this ADR https://github.com/vllm-project/agentic-api/pull/28 :)

leseb (changes requested):
This is how I see the flow #24 ships the proxy logic, #29 splits it into the layered crates, and #27 rebases on top to add the Praxis filter chain that wraps agentic-core functions.

PR #28, opened 2026-05-19
docs: add ADR-03 gateway-agnostic layered architecture

leseb:
## Summary

Adds ADR-03 which captures the architectural decision for how agentic-api integrates with gateway proxies (Praxis, Kong, etc.). Defines a three-layer crate structure: core library (`agentic-core`), HTTP server (`agentic-server`), and thin gateway adapters (`agentic-praxis`). The agentic loop is an explicit Rust state machine rather than decomposed into proxy filter chains. Includes a balanced comparison of both approaches with pros/cons.

## Test Plan

- Documentation-only change, no code impact.
- ADR reviewed iteratively for technical accuracy and tone.

### DIFF
--- a/docs/adr/ADR-03_gateway_integration.md
+++ b/docs/adr/ADR-03_gateway_integration.md
@@ -129,7 +129,7 @@ pub async fn assemble_response(...) -> Result<Response, Error>
 pub async fn persist_response(...) -> Result<(), Error>
 ```
 
-`execute()` is a convenience that composes these steps with the default loop logic. Consumers who need fine-grained control (custom middleware between steps, per-step observability, conditional branching) call the individual functions directly.
+`execute()` is a convenience that composes these steps with the default loop logic. Consumers who need fine-grained control (custom middleware between steps, per-step observability, conditional branching) call the individual functions directly. Each function can also be wrapped in its own gateway filter — a consumer who wants per-step filters can build them from these primitives without the core prescribing the decomposition.
 
 Dependencies: `tokio`, `reqwest`, `serde`, `serde_json`, `sqlx`, `thiserror`. No server-side framework dependencies (`axum`, `praxis`, `tower`).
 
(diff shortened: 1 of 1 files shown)


docs/adr/ADR-03_gateway_integration.md
```diff
@@ -0,0 +1,385 @@
...
+## Crate Structure
+
+```
+agentic-api/
+  Cargo.toml              # [workspace]
+
+  crates/
+    agentic-core/          # Layer 1: pure library
+      Cargo.toml           # [lib], deps: tokio, reqwest, serde, sqlx
+      src/
```
maralbahari, line 92: the exact files names and design details might not be implemented exactly as following. maybe we leave those out as the implementation PRs land in.
but overall functionalities discussed in `decisions  Three-layer crate architecture`  section above layouts out the overview features what agentic-core need to handle.

maralbahari (approved):

docs/adr/ADR-03_gateway_integration.md
```diff
@@ -0,0 +1,385 @@
...
+
+---
+
+## Crate Structure
+
+```
+agentic-api/
+  Cargo.toml              # [workspace]
+
+  crates/
```
maralbahari, line 89: how about instead of crates/ we keep the main src/ follows by  agentic-core/src, agentic-praxies/src, agentic-server/src?

docs/adr/ADR-03_gateway_integration.md
```diff
@@ -0,0 +1,385 @@
...
+    agentic-core/          # Layer 1: pure library
+      Cargo.toml           # [lib], deps: tokio, reqwest, serde, sqlx
+      src/
+        lib.rs
+        executor.rs        # Loop state machine
+        store.rs           # Response store (trait + impls)
+        conversation.rs    # Conversation manager
+        inference.rs       # vLLM proxy / inference caller
+        tools/
```
maralbahari, line 98: I'm thinking how about the tools to own its separate crate package agentic-tools. in case that for some tool we need to rely on a 3rd library that is not native rust?

docs/adr/ADR-03_gateway_integration.md
leseb: thanks! i'm fine to be a little "relaxed" between ADR and final code in PR as soon as it follows the 3 crates approach :)

docs/adr/ADR-03_gateway_integration.md
leseb: i think it's fair feedback but doesn’t need ADR changes, just flexibility during implementation? we don’t know yet which tools will need what. Better to start with tools in agentic-core and split when there’s an actual reason to. I'll update the ADR to mention this as a future consideration.

docs/adr/ADR-03_gateway_integration.md
leseb: The crates/ pattern is the standard Rust convention for workspaces, most major projects use it (bevy, ruff, nushell). Having src/ alongside agentic-core/, agentic-server/, etc. at the root gets confusing fast, what does src/ belong to? With crates/ it’s immediately clear what’s a workspace member and what’s not.

PR #29, opened 2026-05-19
refactor: restructure into 3-crate workspace per ADR-03

leseb:
## Summary

Restructures the flat single-crate layout into a Cargo workspace with three crates per ADR-03:

- **agentic-core**: framework-agnostic library — proxy logic, config, error types, LLM readiness. No axum/clap/tower dependencies. Exposes `proxy_request()` as a composable building block.
- **agentic-server**: standalone axum binary — thin handler adapter, CLI (`clap`), server lifecycle, vLLM subprocess management.
- **agentic-praxis**: stub for future Praxis gateway adapter.

Built on top of PR #24's proxy gateway code (cherry-picked with full commit history). Renames `vllm_*` identifiers to `llm_*` for gateway-agnostic naming. Includes improvements to error logging, multi-value header forwarding, and CLI arg deduplication.

## Test Plan

- `cargo test --workspace` — 16 tests pass (7 unit in agentic-core + 9 integration in agentic-server)
- `cargo bench -p agentic-server --no-run` — benchmarks compile


leseb:
draft because this is on top of #24 and implements proposed ADR in #28

ashwing (commented):
I've been following this refactor and ADR-03 closely — with #24 and #28 now merged, this looks ready to land. Left a few observations while the code is fresh; nothing blocking, mostly forward-looking questions and one small gap I'd be happy to contribute as a follow-up.

crates/agentic-server/src/app.rs
```diff
@@ -0,0 +1,11 @@
...
+use axum::Router;
+use axum::routing::post;
+
+use crate::handler::proxy_responses;
+
+pub fn build_router(state: ProxyState) -> Router {
+    Router::new()
+        .route("/v1/responses", post(proxy_responses))
+        .with_state(state)
+}
```
ashwing, line 11: The gateway router exposes only `POST /v1/responses` but no health endpoint of its own. In K8s or docker-compose deployments, the container orchestrator needs liveness/readiness probes for the gateway process itself — separate from the backend LLM's `/health` that `wait_llm_ready` polls.

Would it make sense to add `GET /health` (200 when listening) and optionally `GET /ready` (also checks LLM reachability via a lightweight probe)? This could live here in `build_router` since it's server-specific, not core logic. Happy to contribute this as a follow-up.

crates/agentic-core/src/proxy.rs
```diff
@@ -0,0 +1,221 @@
+use std::pin::Pin;
```
ashwing, line 1: The original single-crate had unit tests for the internal filter functions (`filter_request_headers`, `filter_response_headers`, `is_sse_content_type`, etc.). In this refactor those are covered indirectly by the integration tests in `agentic-server/tests/proxy_test.rs`, but there are no direct unit tests on the core proxy functions. 

Was that intentional (integration tests are sufficient for the current scope), or would you want unit tests added back in `agentic-core` for faster feedback on regressions — especially as more logic lands in core (executor, store, tool dispatch)?

crates/agentic-server/benches/proxy_bench.rs
```diff
@@ -0,0 +1,174 @@
+use std::convert::Infallible;
```
ashwing, line 1: When `agentic-core` gains the executor loop (ADR-03's `rehydrate → inference → tool_dispatch → assemble → persist`), it'd be valuable to have a benchmark variant that measures per-iteration overhead with a mock tool call loop (1 tool call → re-inference → done). That way we'd catch if the state machine or store adds unexpected latency per loop iteration. Not needed in this PR — just thinking ahead.

crates/agentic-server/src/app.rs
leseb: I think this makes sense as a followup, also see https://github.com/vllm-project/vllm/issues/36960 - thanks!

crates/agentic-server/benches/proxy_bench.rs
leseb: it's certainly something we want to track, although those type of micro benchmark are not easy. happy to discuss more over an issue - thanks!

crates/agentic-core/src/proxy.rs
leseb: Certainly not intended, https://github.com/vllm-project/agentic-api/pull/24/changes/a4080864aaec446af9a73b2c2e77723bae585864 has the unit tests and those have been preserved. thanks!

leseb:
@ashwing thanks for the review, i'd be happy to discuss some of your followup items in issues, thanks!

crates/agentic-server/Cargo.toml
```diff
@@ -0,0 +1,33 @@
+[package]
+name = "agentic-server"
+description = "Standalone axum server for agentic-api"
+version.workspace = true
+edition.workspace = true
+license.workspace = true
+repository.workspace = true
+
+[dependencies]
+agentic-core = { path = "../agentic-core" }
```
maralbahari, line 10: I think we could use agentic-core.workspace = True 
similar to https://github.com/vllm-project/vllm/blob/e1814f822d51225e0c6240386d65b2182758c2a4/rust/src/server/Cargo.toml#L35-L39

maralbahari:
@leseb thank you for the PR.
I think we can rename crates folder to src so the agentic-api/src is not left empty
I think we should follow the rust frontend based on vllm https://github.com/vllm-project/vllm/tree/main/rust/src
while each module is maintaining its own crate package like vllm-tokenizer, vllm-chat, etc but there's still under same source.

crates/agentic-server/src/server.rs
```diff
@@ -0,0 +1,75 @@
+use agentic_core::config::Config;
+use agentic_core::error::Error;
+use agentic_core::proxy::ProxyState;
```
maralbahari, line 3: I think the proxy should be in `agentic-server`? basically the `agentic-server` should handle setting up gateway to vllm-server/LLM-server right and agentic loop is handled in `agentic-server`? 
so basically users would only need to run `agentic-serv`er and it will be our means of communication and http requests between LLM server and parsing events to either call `agentic-core` or `agentic-praxies`?

crates/agentic-core/src/error.rs
```diff
@@ -5,11 +5,11 @@ pub enum Error {
...
 
-    #[error("vLLM not ready within {timeout_s:.0}s at {url}")]
-    VllmTimeout { url: String, timeout_s: f64 },
+    #[error("LLM not ready within {timeout_s:.0}s at {url}")]
+    LlmTimeout { url: String, timeout_s: f64 },
 
-    #[error("vLLM subprocess exited before becoming ready: {status}")]
-    VllmProcessExited { status: String },
+    #[error("LLM subprocess exited before becoming ready: {status}")]
+    LlmProcessExited { status: String },
```
maralbahari, line 12: let's separate out the error that is related to `agentic-server` to there. since `agentic-server` would be setting up the gateway so some of the errors would belong to server only. IO error can stay here. what do you think?

crates/agentic-core/src/config.rs
```diff
@@ -1,24 +1,9 @@
...
-    #[arg(long, default_value_t = 9000)]
-    pub gateway_port: u16,
-
-    #[arg(long, default_value_t = 600.0)]
-    pub vllm_ready_timeout_s: f64,
-
-    #[arg(long, default_value_t = 2.0)]
-    pub vllm_ready_interval_s: f64,
+    pub llm_ready_timeout_s: f64,
+    pub llm_ready_interval_s: f64,
```
maralbahari, line 6: Similar to `errors.rs` let's only keep in config that `agentic-core` would need. 
I think as off now `agentic-core` wouldnt need any `config` till we setup database and the rest of the features in core.

crates/agentic-core/src/readiness.rs
```diff
@@ -0,0 +1,137 @@
...
+        timeout_s,
+    }
+}
+
+/// Poll LLM `/health` until it responds 200 or the timeout is reached.
+///
+/// # Errors
+///
+/// Returns an error if the LLM does not become ready within the configured timeout.
+pub async fn wait_llm_ready(config: &Config) -> Result<(), Error> {
```
maralbahari, line 30: so if `agentic-server` handles the gateway and proxy setup to  either use pure proxy and `agentic-core` or setup the `agentic-praxies` we wouldn't need this file?
we can keep `agentic-core` independence from gateway setup. it solely focuses on core features.
the entrypoint and agentic loop start can be part of `agentic-server` where it decides to dispatch use praxies or core. 
let me know your thoughts. I am not sure what I'm suggesting is the best choice.

crates/agentic-server/Cargo.toml
leseb: will do - thanks

leseb:
> @leseb thank you for the PR. I think we can rename crates folder to src so the agentic-api/src is not left empty I think we should follow the rust frontend based on vllm https://github.com/vllm-project/vllm/tree/main/rust/src while each module is maintaining its own crate package like vllm-tokenizer, vllm-chat, etc but there's still under same source.

i think we discussed it in another pr but this is not how the broad rust ecosystem is doing things, crates/ is the Rust community standard for workspace members (ripgrep, bevy, nushell, deno all use it). vllm uses rust/src/ because Rust is a secondary language in that repo. the crates live inside a Python project. In our repo, Rust is the primary language, so crates/ is idiomatic. Also, src/ would be confusing since each crate already has its own src/ directory.

crates/agentic-core/src/config.rs
franciscojavierarceo: I think a timeout is a rational config and having a config for the runtime feels okay but that's just my opinion

crates/agentic-core/src/error.rs
franciscojavierarceo: agree with this @leseb can we revise?

maralbahari:
> > @leseb thank you for the PR. I think we can rename crates folder to src so the agentic-api/src is not left empty I think we should follow the rust frontend based on vllm https://github.com/vllm-project/vllm/tree/main/rust/src while each module is maintaining its own crate package like vllm-tokenizer, vllm-chat, etc but there's still under same source.
> 
> i think we discussed it in another pr but this is not how the broad rust ecosystem is doing things, crates/ is the Rust community standard for workspace members (ripgrep, bevy, nushell, deno all use it). vllm uses rust/src/ because Rust is a secondary language in that repo. the crates live inside a Python project. In our repo, Rust is the primary language, so crates/ is idiomatic. Also, src/ would be confusing since each crate already has its own src/ directory.

cool .. I might have missed that. thanks for clarifying .

franciscojavierarceo (approved):
i'm going to follow up on @maralbahari's feedback so she can incorporate this into https://github.com/vllm-project/agentic-api/pull/33

PR #30, opened 2026-05-20
feat: axum server with OGX-backed agentic loop and state hydration

franciscojavierarceo:
## Summary

- Replace Praxis filter architecture with a plain axum HTTP server
- Add trait-based store backend (`ResponseStore`, `VectorSearch`) with OGx as first implementation
- Implement state hydration: fetch conversation history from OGx when `previous_response_id` is set
- Implement agentic loop: execute `file_search` tool calls via OGx vector store API, feed results back to vLLM, loop until final response
- Reverse proxy to vLLM with streaming support and auth injection

## Test plan

- 15 tests pass (`cargo test`): 7 proxy, 4 agentic loop, 2 state hydration, 2 config unit tests
- `cargo fmt -- --check` clean
- Tests use mock axum servers for vLLM and OGx — no live services needed
- Integration test with live OGx pending


maralbahari:
@franciscojavierarceo could you please follow the same design as https://github.com/vllm-project/agentic-api/pull/28. agentic-core would not fully in rust native with standard rust package it handles response store and conversation as well. maybe rebase with https://github.com/vllm-project/agentic-api/pull/29 and implement part of `agentic-praxis`

PR #35, opened 2026-05-28
refactor: move gateway modules from agentic-core to agentic-server

franciscojavierarceo:
## Summary

- Moves `proxy.rs`, `readiness.rs`, `config.rs`, and `error.rs` from `agentic-core` to `agentic-server` — these are gateway-specific concerns, not framework-agnostic core abstractions
- Renames `Config` → `GatewayConfig` to clarify the struct's scope
- Migrates all crate `Cargo.toml` files to use `[workspace.dependencies]` declarations
- Addresses feedback from @maralbahari on #29

## Test Plan

- [x] `cargo build` — clean
- [x] `cargo test` — 25/25 pass
- [x] `cargo clippy --all-targets -- -D warnings` — clean
- [x] `cargo fmt -- --check` — clean


ashwing (commented):
Looks good — moving the proxy stuff to agentic-server makes sense since it's all gateway-specific right now.

One thing — PR #33 is adding the store layer which should probably land in agentic-core (it's the framework-agnostic stuff from ADR-03). With this emptying core out, just want to make sure the two PRs don't step on each other around `lib.rs`. Probably fine since yours just leaves the placeholder comment, but figured I'd flag it.

crates/agentic-core/src/lib.rs
ashwing, line 1: Makes sense to keep this alive. Might be worth updating the comment to hint at what's coming (store traits, executor, step functions) so future contributors know this isn't dead code.

franciscojavierarceo (commented):
Thanks for flagging — we expect this PR to merge before #33, so it should be a clean base for the store layer to land on top of.

crates/agentic-core/src/lib.rs
franciscojavierarceo: Good call — updated the comment to hint at what's coming.

maralbahari (approved):
LGTM, Thanks

leseb:
@maralbahari @franciscojavierarceo this goes against ADR3, we have agentic-core containing all functions, proxy_request is a framework agnostic request, takes a payload and passes to the inference backend, if we put in the server now praxis (or any other gw) has to bring the axum dep too and it’s not what we want. thanks!

can we revert?

franciscojavierarceo:
yeah sorry @leseb

franciscojavierarceo:
@maralbahari we can discuss more in slack (or here too of course)

PR #39, opened 2026-05-29
feat: add CORS support to gateway

### LINKED ISSUE #38: feat: add CORS support to gateway
The gateway currently returns `405 Method Not Allowed` for `OPTIONS` requests and doesn't include any `Access-Control-*` headers. This blocks browser-based clients (playground UIs, Swagger, JS apps) from calling the API directly.

**Reproduction:**
```bash
curl -v -X OPTIONS http://localhost:9000/v1/responses \
  -H "Origin: http://example.com" \
  -H "Access-Control-Request-Method: POST"
# Returns: 405 Method Not Allowed, no CORS headers
```

`tower-http` with the `cors` feature is already in `agentic-server/Cargo.toml` — just not wired into the router. A permissive default (allow any origin/method/header) would unblock local development; production deployments can tighten it via config.

Straightforward fix — happy to pick this up if nobody else grabs it.

cc @leseb @franciscojavierarceo

ashwing:
## Summary

Wire `tower-http` `CorsLayer` into the router with configurable origin restrictions via the `CORS_ALLOWED_ORIGINS` environment variable.

- **Unset or empty** → permissive (allow any origin) — suitable for local dev, playground UIs, Swagger
- **Set** → comma-separated list of allowed origins (e.g., `CORS_ALLOWED_ORIGINS=https://app.example.com,https://admin.example.com`)

Adds a `ServerConfig` struct that reads CORS configuration from the environment at startup and passes it to `build_router`.

Closes #38

## Configuration

| Env Var | Default | Effect |
|---------|---------|--------|
| `CORS_ALLOWED_ORIGINS` | (unset) | Allow any origin (permissive) |
| `CORS_ALLOWED_ORIGINS=https://app.example.com` | — | Only that origin allowed |
| `CORS_ALLOWED_ORIGINS=https://a.com,https://b.com` | — | Both origins allowed |

Methods and headers are always permissive (`Any`) regardless of origin setting.

## Test Plan

**Automated (2 integration tests):**
- `test_cors_preflight_returns_200` — OPTIONS with Origin + Access-Control-Request-Method → 200 with all CORS headers
- `test_cors_headers_on_regular_request` — POST with Origin → response includes `access-control-allow-origin`

```
cargo test --workspace
# 31 tests pass (23 unit + 2 main + 4 health + 2 CORS)
```

**Lint/format:**

**Discovery:** Found via live stress testing against deployed gateway on EC2 (see #38 for reproduction).

### DIFF
--- a/crates/agentic-server/src/app.rs
+++ b/crates/agentic-server/src/app.rs
@@ -1,13 +1,17 @@
 use agentic_core::proxy::ProxyState;
 use axum::Router;
 use axum::routing::{get, post};
+use tower_http::cors::{Any, CorsLayer};
 
 use crate::handler::{health, proxy_responses, ready};
 
 pub fn build_router(state: ProxyState) -> Router {
+    let cors = CorsLayer::new().allow_origin(Any).allow_methods(Any).allow_headers(Any);
+
     Router::new()
         .route("/health", get(health))
         .route("/ready", get(ready))
         .route("/v1/responses", post(proxy_responses))
+        .layer(cors)
         .with_state(state)
 }
--- a/crates/agentic-server/src/app.rs
+++ b/crates/agentic-server/src/app.rs
@@ -1,17 +1,57 @@
 use agentic_core::proxy::ProxyState;
 use axum::Router;
 use axum::routing::{get, post};
-use tower_http::cors::{Any, CorsLayer};
+use http::HeaderValue;
+use tower_http::cors::{AllowOrigin, Any, CorsLayer};
 
 use crate::handler::{health, proxy_responses, ready};
 
-pub fn build_router(state: ProxyState) -> Router {
-    let cors = CorsLayer::new().allow_origin(Any).allow_methods(Any).allow_headers(Any);
+/// Server-level configuration read from environment variables.
+pub struct ServerConfig {
+    pub cors_allowed_origins: Vec<String>,
+}
 
+impl ServerConfig {
+    /// Read `CORS_ALLOWED_ORIGINS` (comma-separated). Unset or empty = permissive.
+    #[must_use]
+    pub fn from_env() -> Self {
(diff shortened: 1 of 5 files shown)


crates/agentic-server/src/app.rs
```diff
@@ -1,11 +1,19 @@
...
 use axum::Router;
 use axum::routing::post;
+use tower_http::cors::{Any, CorsLayer};
 
 use crate::handler::proxy_responses;
 
 pub fn build_router(state: ProxyState) -> Router {
+    // Permissive CORS default — allows any origin for local dev and playground UIs.
+    // For production, this should be tightened. Two options:
+    //   1. Env-driven: read CORS_ALLOWED_ORIGINS (comma-separated), fall back to Any
```
franciscojavierarceo: let's create a `ServerConfig` struct and add there, yeah? we can pass it to the `build_router` function.

something like:

```rust
impl ServerConfig {
    /// Read `CORS_ALLOWED_ORIGINS` (comma-separated). Unset or empty = permissive.
    pub fn from_env() -> Self {
        let cors_allowed_origins = std::env::var("CORS_ALLOWED_ORIGINS")
            .ok()
            .map(|s| {
                s.split(',')
                    .map(str::trim)
                    .filter(|o| !o.is_empty())
                    .map(str::to_owned)
                    .collect::<Vec<_>>()
            })
            .unwrap_or_default();
        Self { cors_allowed_origins }
    }
}
```


would be good to update the docs with this as setting here would default to permissive.

ashwing: Done in 1738152 — added ServerConfig::from_env() that reads CORS_ALLOWED_ORIGINS (comma-separated). Empty/unset defaults to permissive. Updated build_router to take the config, and documented the env var in the PR body.

franciscojavierarceo (approved):

PR #40, opened 2026-05-29
Revert "refactor: move gateway modules from agentic-core to agentic-server"

franciscojavierarceo:
Reverts vllm-project/agentic-api#35


leseb (approved):
@maralbahari PTAL

PR #44, opened 2026-05-29
docs: core public API design — phased implementation plan

ashwing:
## Summary

Design reference for the `agentic-core` public API extensions. This doc defines what we're building on top of PR #46's base executor loop.

**Ownership split:**
- @maralbahari — base loop: `execute()`, `rehydrate_conversation()`, `call_inference()`, `persist_response()` (PR #46)
- @ashwing — tool dispatch, loop control, streaming tee, executor traits (this doc + follow-up PRs)

**Implementation phases (each = one PR):**

| Phase | Scope | Depends On |
|-------|-------|------------|
| 1 | SSEEventType expansion + accumulator FunctionCall detection | nothing (lands on main) |
| 2 | `LoopDecision` + `dispatch_tools()` + `execute_loop()` | PR #46 |
| 3 | Streaming tee (forward to client + accumulate) | PR #46 |
| 4 | Tool executor traits + mock impls (MCP, web_search, vector_store) | Phase 2 |

**What changed in this update:**
- Removed all code stubs (per @maralbahari's feedback — impl goes in follow-up PRs)
- Reframed around PR #46 as the foundation
- Reduced from 626 lines to ~230 — just enough to execute, not over-specify
- Added Praxis filter mapping table with ownership per step

## Test Plan

Design doc only — no code changes. Verification via follow-up implementation PRs with integration tests.

### DIFF
--- a/docs/design/core-public-api.md
+++ b/docs/design/core-public-api.md
@@ -1,0 +1,625 @@
+# Design: `agentic-core` Public API
+
+> Status: Draft — soliciting inline review
+> References: [ADR-03](../adr/ADR-03_gateway_integration.md), [Issue #42](https://github.com/vllm-project/agentic-api/issues/42), [Praxis #354](https://github.com/praxis-proxy/praxis/issues/354)
+> Owner: @ashwing
+
+---
+
+## Overview
+
+`agentic-core` exposes the agentic loop as composable step functions. Each function is:
+- Independently testable (no HTTP server needed)
+- Wrappable in a Praxis `HttpFilter` (1:1 mapping per ADR-03)
+- Callable directly in standalone mode via `execute()`
+
+This design incorporates @leseb's expanded proposal (14 functions, 8 traits) and organizes it into implementation phases.
+
+**Relationship to existing code:** PR #33 landed a concrete `ConversationStore` + `ResponseStore` with SQLx. This design defines the *orchestration-level* traits that wrap or delegate to those concrete stores. The implementation PRs will wire against PR #33's types directly.
+
+**Step numbering:** Steps are numbered sequentially within each phase (A1, A2, ... B1, B2, ... C1, C2, ...). The Praxis filter mapping table at the end shows how these map to @leseb's filter chain order.
+
+---
+
+## Shared State: `AgenticState`
+
+Mutable, request-scoped struct that flows through the entire loop. Each step reads some fields and writes others.
+
+```rust
+pub struct AgenticState {
+    // Identity (set once by validate_request, read-only thereafter)
+    pub response_id: String,
+    pub conversation_id: Option<String>,
+    pub tenant_id: Option<String>,
+    pub model: String,
+    pub previous_response_id: Option<String>,
+    pub request: Value,  // original request preserved for forwarding
+
+    // Routing flags (set once by validate_request)
+    pub store_enabled: bool,
+    pub stream_enabled: bool,
+    pub background: bool,
+
+    // Conversation (written by rehydrate, read by call_inference)
+    pub messages: Vec<Value>,
+
+    // Tools (written by dispatch_tools, read by call_inference on loop)
+    pub tools: Vec<Value>,
+    pub tool_choice: Value,
+    pub tool_calls: Vec<Value>,
+
+    // Output (written by transform_stream, read by persist)
+    pub output_items: Vec<Value>,
+    pub response_object: Value,
+    pub usage: Value,
+    pub status: ResponseStatus,
+
+    // Loop control
+    pub iteration: u32,
+
+    // Runtime
+    pub config: AgenticConfig,
+}
+
+pub struct AgenticConfig {
+    pub max_tool_iterations: u32,
+    pub compaction_enabled: bool,
+    pub compaction_threshold_tokens: u64,
+    pub reasoning_summary_enabled: bool,
+    pub default_model: Option<String>,
+}
+
+impl Default for AgenticConfig {
+    fn default() -> Self {
+        Self {
+            max_tool_iterations: 10,
+            compaction_enabled: false,
+            compaction_threshold_tokens: 100_000,
+            reasoning_summary_enabled: false,
+            default_model: None,
+        }
+    }
+}
+
+pub enum ResponseStatus {
+    Queued,
+    InProgress,
+    Completed,
+    Incomplete(String),
+    Failed(String),
+    Cancelled,
+}
+
+pub enum LoopDecision {
+    Continue,
+    Done,
+    Incomplete(String),
+}
+
+pub enum BackendFormat {
+    Responses,
+    ChatCompletions,
+}
+```
+
+---
+
+## Step Functions
+
+### Phase A: Core Loop
+
+These are the minimum set needed for a working agentic loop.
+
+#### A1: `validate_request`
+
+```rust
+pub fn validate_request(
+    body: &Value,
+    config: &AgenticConfig,
+) -> Result<AgenticState, AgenticError>
+```
+
+Parse the incoming request. Extract routing flags (`stream`, `store`, `background`), `model`, `previous_response_id`, `conversation_id`. Generate `response_id`. Validate constraints (`background=true && store=false` is invalid). Return fully initialized `AgenticState`.
+
+Does NOT validate inference-specific params (temperature, top_p, etc.) — those are the backend's responsibility.
+
+**Gateway filter:** `request_validate`
+
+---
+
+#### A2: `rehydrate_conversation`
+
+```rust
+pub async fn rehydrate_conversation(
+    state: &mut AgenticState,
+    store: &dyn ResponseStore,
+) -> Result<(), AgenticError>
+```
+
+Load conversation history from `previous_response_id` or `conversation_id`. Reconstruct message list. Append current input.
+
+- If `previous_response_id`: fetch stored response + messages. Returns `AgenticError::Validation` if the previous response is incomplete/in-progress/cancelled (cannot continue from an unfinished response).
+- If `conversation_id` only: fetch conversation messages
+- If neither: pass current input through (messages extracted from `state.request`)
+
+**Gateway filter:** `rehydrate`
+
+---
+
+#### A3: `call_inference`
+
+```rust
+pub async fn call_inference(
+    state: &AgenticState,  // intentionally &, not &mut — inference is read-only on state
+    client: &dyn InferenceClient,
+) -> Result<(InferenceStream, BackendFormat), AgenticError>
+```
+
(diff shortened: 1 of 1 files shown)


maralbahari:
@ashwing Thank you for the job on this PR , it clarifies general functions required in the core module howeverI have a concern with this is that the design implementations are based on agent assumptions on some input output which is not explicit and once the actual implementation is in place the design might not be working well and developer would need to make so much changes and the review process would become complex.
I feel we should focus on actual small feature implementation with benchmarks and a smoke test of each small feature so we know the actual usage of each module and their purpose.

ashwing:
Fair point — stubs that drift from implementation would just create rework. I'll shift the approach: instead of landing all the type definitions at once, I'll implement one complete slice end-to-end (e.g., `rehydrate_conversation` with a real test against your store layer from PR #33) so the types are validated by actual usage rather than assumptions.

The design doc stays as a reference for the overall direction, but the PRs will be small, tested features — one step function at a time with integration tests proving the interfaces work.

ashwing:
Updated — stripped the code stubs and kept this as a pure design doc for inline review. Implementation will follow as separate small PRs with integration tests against the store layer.

@leseb @franciscojavierarceo @maralbahari would appreciate your review on the design direction in `docs/design/core-public-api.md`. Happy to adjust based on feedback before starting implementation.

maralbahari:
> Updated — stripped the code stubs and kept this as a pure design doc for inline review. Implementation will follow as separate small PRs with integration tests against the store layer.
> 
> @leseb @franciscojavierarceo @maralbahari would appreciate your review on the design direction in `docs/design/core-public-api.md`. Happy to adjust based on feedback before starting implementation.

Thank you @ashwing for the proposal. I have a simple response/conversation flow in https://github.com/vllm-project/agentic-api/pull/46. this PR only implements the stateful agent loop for simple input text messages it might look big because of the integration tests and cassettes records against OpenAI's conversation and responses api for assessment. However the agentic-loop  is not complete with function_calls, instructions the default tool calls as  your proposal which contains a bigger skeleton to ship. as discussed on slack we can work on https://github.com/vllm-project/agentic-api/pull/46 to have a more explicit implementation and complete the agentic-loop. I think could start with implementation to complete the function_calls and handling reasoning tokens etc?

ashwing:
@maralbahari sounds good — I reviewed PR #46 and left some comments there. Makes sense for you to own the stateful text flow as the base, and I'll build function_calls, tool dispatch, and the looped execution on top once #46 lands. The design doc here stays as the reference for the broader skeleton.

ashwing:
@franciscojavierarceo @maralbahari @leseb Gentle nudge to review the updated implementation plan. The doc is now minimal with work items phased out.

docs/design/core-public-api.md
```diff
@@ -0,0 +1,239 @@
...
+
+This enables the real-time streaming requirement from ADR-01 §3 — events should reach the client as they arrive, interleaved with the tool loop, rather than buffered until completion.
+
+**Size:** ~300 lines | **Blocked by:** PR #46 merge | **Target:** feature PR
+
+---
+
+### Phase 4: Tool Executor Traits + Mock Implementations (depends on Phase 2)
+
+**PR scope:** `executor/tools/` module.
```
maralbahari: @ashwing I think for the tools it should be on just agentic-core/tools as ADR-03
we need to keep the modules in agentic-core as small as possible to avoid circular imports and dependencies so it is easier to maintain as well.

docs/design/core-public-api.md
```diff
@@ -0,0 +1,239 @@
...
+}
+```
+
+**Size:** ~400 lines | **Blocked by:** PR #46 merge | **Target:** first feature PR (Phase 2 of committer track)
+
+---
+
+### Phase 3: Streaming Tee (depends on PR #46)
+
+**PR scope:** `executor/stream_tee.rs`, refactor `run_stream` path.
```
maralbahari, line 148: @ashwing I think as SSE streaming grows we can also maintain them in their own small module in `agentic-core/sse`.

docs/design/core-public-api.md
```diff
@@ -0,0 +1,239 @@
...
+| 10 | `file_search` | `VectorStoreClient::search()` | Phase 4 | @ashwing |
+| 11 | `compact` | `compact_context()` | Future | — |
+| 12 | `reasoning` | `summarize_reasoning()` | Future | — |
+| 13 | `response_store` (resp) | `persist_response()` | PR #46 | @maralbahari |
+
+---
+
+## Open Questions
+
+1. **`execute_loop` vs refactoring `execute`:** Should the loop wrapper be a new function or replace PR #46's `execute()`? Pending maralbahari's response on PR #46 review.
```
maralbahari, line 257: the `execute` in https://github.com/vllm-project/agentic-api/pull/46 is temporary as the entire loop is not complete with whole functionality since it is only validating text-only responses.  so that we can test each small component and verify their correctness as we implement each composable pieces.
we shall write the last `execute_loop` once we have all the pieces in.

maralbahari:
> @franciscojavierarceo @maralbahari @leseb Gentle nudge to review the updated implementation plan. The doc is now minimal with work items phased out.

Thank you @ashwing I have responded to your questions in https://github.com/vllm-project/agentic-api/pull/46#discussion_r3352104210
for the SSE events and handling them during streaming in more details would need your feedback on it thanks.

docs/design/core-public-api.md
```diff
@@ -0,0 +1,259 @@
...
+        store_id: &str,
+        query: &str,
+        max_results: u32,
+    ) -> impl Future<Output = Result<Vec<Value>, ExecutorError>> + Send;
+}
+```
+
+This PR includes mock implementations for integration testing (in-memory tool executors that return canned responses). Real implementations (MCP client, Brave search, Qdrant) come in later PRs.
+
+**Note:** These traits must be compatible with @franciscojavierarceo's OGX integration (PR #34). The trait-based approach allows OGX to be one implementation behind `McpToolExecutor` — the dispatch layer doesn't care whether tools run via OGX or native Rust.
```
franciscojavierarceo: @ashwing Can you clarify what you mean here? OGX is a vector store search service—the existing integration (vector_search/ogx.rs) implements a VectorSearch trait and hits OGX's REST API directly. It's not an MCP server.

Did you mean OGX would be an implementation behind `VectorStoreClient` (not `McpToolExecutor`)? That would make sense and is basically what we already have. 

If you're proposing routing OGX through MCP as an intermediary, I'd push back on that—it adds indirection for something that's a straightforward REST call.

docs/design/core-public-api.md
```diff
@@ -0,0 +1,259 @@
...
+## Praxis Filter Mapping
+
+How the complete pipeline maps to @leseb's proposed filter chain:
+
+| # | Praxis Filter | Core Function | Phase | Owner |
+|---|---------------|---------------|-------|-------|
+| 0 | `request_validate` | `validate_request()` | Future | — |
+| 1 | `response_store` (init) | `init_store()` | Future | — |
+| 2 | `rehydrate` | `rehydrate_conversation()` | PR #46 | @maralbahari |
+| 3 | `file_resolve` | `resolve_files()` | Future | — |
```
franciscojavierarceo: happy to take this one on

docs/design/core-public-api.md
```diff
@@ -0,0 +1,259 @@
...
+| 4 | `tool_parse` | `parse_tools()` | Future | — |
```
franciscojavierarceo: and this

docs/design/core-public-api.md
```diff
@@ -0,0 +1,259 @@
...
+| 5 | `responses_proxy` | `call_inference()` | PR #46 | @maralbahari |
+| 5.5 | `event_normalize` | `normalize_sse_line()` | Phase 1 | @ashwing |
+| 6 | `stream_events` | `transform_stream()` / tee | Phase 3 | @ashwing |
+| 7 | `tool_dispatch` | `dispatch_tools()` | Phase 2 | @ashwing |
+| 8 | `mcp_tool` | `McpToolExecutor::execute()` | Phase 4 | @ashwing |
+| 9 | `web_search` | `WebSearchProvider::search()` | Phase 4 | @ashwing |
+| 10 | `file_search` | `VectorStoreClient::search()` | Phase 4 | @ashwing |
+| 11 | `compact` | `compact_context()` | Future | — |
```
franciscojavierarceo, line 249: and this

docs/design/core-public-api.md
```diff
@@ -0,0 +1,259 @@
...
+| 12 | `reasoning` | `summarize_reasoning()` | Future | — |
```
franciscojavierarceo, line 250: same here

docs/design/core-public-api.md
```diff
@@ -0,0 +1,259 @@
...
+
+---
+
+## Praxis Filter Mapping
+
+How the complete pipeline maps to @leseb's proposed filter chain:
+
+| # | Praxis Filter | Core Function | Phase | Owner |
+|---|---------------|---------------|-------|-------|
+| 0 | `request_validate` | `validate_request()` | Future | — |
```
franciscojavierarceo, line 237: i can do this one too

franciscojavierarceo (approved):
some small nits but otherwise this lgtm

docs/design/core-public-api.md
ashwing: You're right — that was sloppy wording on my part. OGX is a `VectorStoreClient` implementation, not MCP. The dispatch layer routes by tool type: function calls → `McpToolExecutor`, file_search → `VectorStoreClient` (your OGX impl), web_search → `WebSearchProvider`. No MCP indirection for direct REST calls.

Won't push a fix since you've already approved — the correct mapping is in the Praxis filter table below and will be reflected in the code PRs.

docs/design/core-public-api.md
ashwing: Awesome — updated ownership split in my head:

**You:**
- `file_resolve` / `resolve_files()`
- `tool_parse` / `parse_tools()`
- `web_search` / `WebSearchProvider` (Brave or similar)
- `file_search` / `VectorStoreClient` (OGX)

**Me:**
- `event_normalize` (PR #49 — submitted)
- `stream_events` / streaming tee
- `tool_dispatch` / `dispatch_tools()` + `LoopDecision`
- `mcp_tool` / `McpToolExecutor`

The dispatch layer calls your providers via the trait interface so we can develop in parallel once trait definitions are agreed. Let me know if you want to define the trait signatures for your pieces or if you'd rather I propose them in a follow-up PR and you iterate.

docs/design/core-public-api.md
maralbahari: @ashwing just nit the tools would be in `agentic-core/tools`  not `executor/tools`

ashwing: Makes sense — keeps tool traits at the crate top level rather than coupling them to the executor. `crates/agentic-core/src/tools/` it is. The executor just imports and calls them.

ashwing:
@maralbahari @franciscojavierarceo Updated — tools path, OGX mapping, and ownership table all reflect your feedback. Thanks!

PR #45, opened 2026-06-03
fix: storage CRUD follow-up — schema placeholders, metadata, unknown item types

franciscojavierarceo:
## Summary

Follow-up to #33 addressing outstanding review feedback:

- **Schema placeholders**: New migration `0002` adds `tenant_id`, `metadata`, and `raw_tokens` columns to avoid future schema migrations when multi-tenancy, conversation metadata, and token pass-through land
- **Conversation metadata**: `metadata: Option<String>` added to `Conversation` model and `ConversationData` domain type with `From` impl coverage
- **Unknown item type fallback**: `#[serde(other)] Unknown` variant on `InputItem` and `OutputItem` so unrecognized item types (web search, MCP, reasoning, compaction, etc.) don't silently fail deserialization
- **Observability**: `tracing::warn!` in `as_inout()` when items can't be deserialized — tries `InputItem` first, then `OutputItem`, only warns if both produce `Unknown`

## Test Plan

- All 84 existing tests pass (`cargo test`)
- `cargo fmt -- --check` clean
- Pre-commit hooks pass

### DIFF
--- a/crates/agentic-core/src/storage/models/conversation.rs
+++ b/crates/agentic-core/src/storage/models/conversation.rs
@@ -12,6 +12,9 @@ pub struct Conversation {
     /// Unique conversation identifier.
     pub id: String,
 
+    /// Optional metadata as JSON string.
+    pub metadata: Option<String>,
+
     /// Creation timestamp as Unix timestamp in seconds.
     pub created_at: i64,
 }
@@ -69,10 +72,12 @@ mod tests {
     fn test_conversation_basic() {
         let conversation = Conversation {
             id: "conv_1".to_string(),
+            metadata: None,
             created_at: 1_704_067_200,
         };
 
         assert_eq!(conversation.id, "conv_1");
+        assert!(conversation.metadata.is_none());
         assert_eq!(conversation.created_at, 1_704_067_200);
     }
 }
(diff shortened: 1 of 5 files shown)


crates/agentic-core/src/storage/models/item.rs
```diff
@@ -44,9 +46,18 @@ impl Item {
     /// Deserialize data column as either `InputItem` or `OutputItem`.
     #[must_use]
     pub fn as_inout(&self) -> Option<InOutItem> {
-        self.as_input()
-            .map(InOutItem::Input)
-            .or_else(|| self.as_output().map(InOutItem::Output))
+        if let Some(input) = self.as_input() {
```
maralbahari: it would be cleaner to just use a single `match` expression rather than if then match.

crates/agentic-core/src/storage/models/conversation.rs
```diff
@@ -12,6 +12,9 @@ pub struct Conversation {
     /// Unique conversation identifier.
     pub id: String,
 
+    /// Optional metadata as JSON string.
+    pub metadata: Option<String>,
```
maralbahari, line 16: @franciscojavierarceo is this metadata different than the one stored in `response` table? if not I think we dont need it here cause the information already exist in `response` table

franciscojavierarceo: it is part of OpenAI spec: https://developers.openai.com/api/reference/resources/conversations/methods/create

franciscojavierarceo:
@maralbahari Good call on the match — refactored the if-let chain into a single match expression in 3a2caf6.

leseb (approved):

PR #49, opened 2026-06-10
feat: add SSE event normalizer module

ashwing:
## Summary

Adds `events/` module to `agentic-core` — a pure parsing library that normalizes raw SSE data lines into typed `EventFrame` structs. This is the foundation for tool dispatch, streaming tee, and loop control (Phases 2–4 of the core API design in PR #44).

- **20-variant `SSEEventType` enum** covering all Responses API event types (text, function_call, reasoning, file_search, web_search)
- **`EventPayload` enum** with typed extraction per event (no downstream JSON access needed)
- **`normalize_sse_line(&str) → Option<EventFrame>`** — pure function, no state, no async
- Handles both vLLM (`response.done`) and OpenAI (`response.completed`) wire formats
- `#[non_exhaustive]` on enums for forward compatibility
- No dependency on the executor module — lands on main independently of PR #46

Per discussion with @maralbahari on [PR #46](https://github.com/vllm-project/agentic-api/pull/46#discussion_r3383789521): this is a separate core module to avoid bloating the accumulator. Once PR #46 merges, a follow-up refactors the accumulator to consume `EventFrame` instead of inline JSON parsing.

**Validated against live vLLM** (google/gemma-4-26B-A4B-it, v0.21.0) — cassettes recorded from real streaming responses including function_call tool use.

## Test Plan

- 33 integration tests in `tests/event_normalizer_test.rs`:
  - Per-event-type parsing (text, function_call, reasoning, response lifecycle, content_part, file/web search)
  - Edge cases: `[DONE]`, empty lines, malformed JSON, unknown events, empty deltas, unicode
  - Full streaming sessions: text-only, function_call, parallel calls, mixed text+tool
  - Real vLLM cassette replay (recorded from live gemma-4-26B-A4B-it)
  - YAML cassette-driven tests (2 cassette files)
- `cargo fmt --check` — clean
- All 117 workspace tests pass (no regressions)


crates/agentic-core/src/events/normalize.rs
```diff
@@ -0,0 +1,174 @@
...
+/// Returns `None` for non-data lines, empty lines, and the `data: [DONE]`
+/// sentinel.
+#[must_use]
+pub fn normalize_sse_line(line: &str) -> Option<EventFrame> {
+    let data_str = line.strip_prefix("data: ")?;
+    if data_str == "[DONE]" {
+        return None;
+    }
+
+    let json: Value = serde_json::from_str(data_str).ok()?;
```
maralbahari: @ashwing could use `deserialize_from_str_opt` from `utils/common.rs`

crates/agentic-core/tests/cassettes/events/function-call-vllm-gemma4.yaml
```diff
@@ -0,0 +1,37 @@
+# Recorded from google/gemma-4-26B-A4B-it via vLLM v0.21.0
+# Request: {"model":"google/gemma-4-26B-A4B-it","input":"What is the weather in San Francisco?","tools":[...],"tool_choice":"required","stream":true}
+# Date: 2026-06-09
```
maralbahari, line 3: I think for the cassettes you can record whole response not only the sse. so later after on function_call feature support you can use them directly to verify the model output. what do you think? unless you are planning to test `function_call` with the multi-turn stateful features then we dont need it here we add new cassettes record for non text output for multi-turns?

crates/agentic-core/src/events/normalize.rs
ashwing: Good call — fixed, using `deserialize_from_str_opt` from utils. Will push shortly.

crates/agentic-core/tests/cassettes/events/function-call-vllm-gemma4.yaml
ashwing: Makes sense for full-response cassettes when testing the executor end-to-end (your format in PR #46 already handles that well). For this module the input is just the raw `data:` line — the normalizer has no concept of HTTP headers or request/response envelope. I'll record full cassettes in your format when we add function_call support to the executor integration tests.

maralbahari (approved):
LGTM

franciscojavierarceo (approved):

ashwing:
@maralbahari @franciscojavierarceo looks good to merge?

PR #50, opened 2026-06-10
fix: preserve response hydration checkpoints

franciscojavierarceo:
## Summary
- Materialize ADR-02 response checkpoints by carrying prior history item IDs forward.
- Hydrate previous responses/conversations even when the new turn uses `store=false`, while keeping `store=false` turns unpersisted.
- Reject requests that provide both `conversation_id` and `previous_response_id`, and assert hydrated upstream request bodies in stateful tests.

## Test Plan
- `cargo test -p agentic-core`
- `cargo clippy -p agentic-core --all-targets -- -D warnings`


maralbahari (approved):
@franciscojavierarceo thank you for catching the these. LGTM

Issue #42, opened 2026-05-29
feat: agentic-core public API — composable step functions per ADR-03


ashwing:
# Proposal: `agentic-core` public API — composable step functions per ADR-03

## Context

ADR-03 defines the core crate's public API as a set of composable async functions that implement the agentic loop. Each function is a standalone step that can be:
- Called directly in standalone mode (`execute()` composes them)
- Wrapped in a Praxis `HttpFilter` for gateway mode
- Tested in isolation without HTTP infrastructure

The ADR intentionally left the function parameters as `(...)` placeholders. This issue proposes concrete type signatures, informed by:
- The ADR-02 store schema (three-table model: Conversation/Response/Item)
- PR praxis-proxy/praxis#34's working agentic loop implementation (tool dispatch pattern)
- vLLM's own `ConversationContext` pattern in `responses/serving.py`
- RFC praxis-proxy/praxis#18's session-aware KV cache requirements (future extensibility)

## Design Goals

1. **Each step is independently testable** — no hidden coupling between steps
2. **`serde_json::Value` at boundaries, typed internally** — the gateway is model-agnostic; we don't enforce a fixed item schema at the API level
3. **Trait-based backends** — store, inference client, and tool dispatch are swappable
4. **Streaming-first** — inference returns a stream, not a collected response
5. **Tool loop is explicit** — the caller decides iteration policy, not the library
6. **Future-proof for RFC praxis-proxy/praxis#18** — `ExecutionContext` carries optional cache hints without breaking the API

### `ExecutionContext` — shared state across steps

```rust
pub struct ExecutionContext {
    /// Pluggable response/conversation store (ADR-02).
    pub store: Arc<dyn ResponseStore>,

    /// Inference backend caller.
    pub inference: Arc<dyn InferenceClient>,

    /// Tool registry for dispatch.
    pub tools: Arc<dyn ToolRegistry>,

    /// Gateway configuration (LLM base URL, timeouts, etc.)
    pub config: CoreConfig,
}
```

### Store trait (ADR-02 alignment)

```rust
#[async_trait]
pub trait ResponseStore: Send + Sync {
    /// Load a response by ID, returning its metadata and history_item_ids.
    async fn get_response(&self, response_id: &str) -> Result<StoredResponse, Error>;

    /// Bulk-fetch items by IDs. Caller is responsible for re-ordering.
    async fn get_items(&self, item_ids: &[String]) -> Result<Vec<StoredItem>, Error>;

[...]

PR #51, opened 2026-06-11
feat: add tool dispatch layer — ToolContext, traits, and LoopDecision

ashwing:
## Summary

Server-side tool dispatch infrastructure for `agentic-core`. Adds `execute_loop` — an agentic loop orchestrator that executes tool calls and re-enters inference, so the gateway can resolve tools without client round-trips.

This is a **library-only PR** (Step 1 in the sequencing below). It adds tested, callable functions but does NOT wire them into the server handler yet. The client-side passthrough flow (vLLM returns `function_call` → client executes → client re-submits) continues to work as before via `execute()` directly.

**What this enables (once wired):**
```
Client request → LLM responds with FunctionCall → gateway executes tool →
injects result → LLM responds with final answer → return to client
```

All within a single request. The client never sees the intermediate tool calls.

## Sequencing

Agreed with @noobHappylife and @maralbahari in the comments below:

| Step | PR | Status | What |
|------|-----|--------|------|
| 0 | #59 | **Merged** | function_call accumulation — accumulator produces `OutputItem::FunctionCall` from SSE |
| 1 | #51 (this) | Ready for review | Server-side loop infrastructure — `execute_loop` + `dispatch_tools` + `ToolContext` + traits |
| 2 | TBD | Future | Type-aware dispatch — pass `tools` array into `dispatch_tools`, route by `type` field |
| 3 | TBD | Future | Tool normalization — MCP → function-compatible schemas before vLLM sees them |
| 4 | TBD | Future | Server handler wiring — `handler.rs` calls `execute_loop` when tools present |

**Key design point (from @noobHappylife):** vLLM only speaks function calling. Other tool types (`mcp`, `web_search`, `file_search`) must be normalized into function-compatible schemas before inference, while keeping metadata to route the output back to the correct executor. That normalization is Step 3.

## What's Added

| Module | What | Lines |
|--------|------|-------|
| `tools/mcp.rs` | `McpToolExecutor` trait | 18 |
| `tools/web_search.rs` | `WebSearchProvider` trait | 15 |
| `tools/vector_store.rs` | `VectorStoreClient` trait | 17 |
| `executor/dispatch.rs` | `dispatch_tools()` → `LoopDecision` | 101 |
| `executor/tool_context.rs` | `ToolContext` + parallel `execute_all()` | 163 |
| `executor/execute_loop.rs` | Orchestrator: execute → dispatch → loop | 184 |

## How It Works

```
execute_loop(request, exec_ctx, tool_ctx)
  │
  ├── Capture original IDs (prev_resp_id, conv_id)
  ├── Clear persistence triggers (store=false, IDs=None)
  │
  ├── for each iteration (capped at 128):
  │     │
  │     ├── execute(request) → ResponsePayload  [30s timeout]
  │     │     └── store=false, no IDs → internal execute() skips persist
  │     │
  │     ├── dispatch_tools(output) → LoopDecision
  │     │     ├── Done → restore IDs, return payload to caller
  │     │     ├── Incomplete → set status, restore IDs, return
  │     │     └── Continue(results) → extend input in-place, loop
  │     │
  │     └── tool_ctx.execute_all(calls)  [30s per call, parallel]
  │           └── individual failures → error JSON (not fatal)
  │
  └── Caller owns persistence. Loop never writes to DB.
```

## Safety Guarantees

- **Hard loop cap:** 128 iterations (defense-in-depth)
- **Soft tool cap:** `max_iterations` default 10 (via `dispatch_tools`)
- **LLM timeout:** 30s per inference call
- **Tool timeout:** 30s per tool call
- **No partial persistence:** All 3 triggers cleared (`store`, `previous_response_id`, `conversation_id`) — prevents intermediate tool-call-only responses from being written to store (required since PR #56 persists when ANY of the three is set)
- **ID restoration:** Both `previous_response_id` and `conversation_id` are captured before the loop and restored on the final payload (Done and Incomplete paths)
- **In-place input growth:** Tool results extend the input vec in-place — no quadratic clone-and-reassign

## Test Plan

32 tests covering happy paths, error paths, boundaries, persistence invariants, and a cassette recorded from live vLLM (gemma-4-26B-A4B-it). 3 rounds of adversarial review passed. All workspace tests green.

<details>
<summary>Full test list (32 tests)</summary>

**dispatch_tools (10):** no calls→Done, single/parallel→Continue, max iterations→Incomplete, failing tool→error output, no provider→error, mixed results, message+FC mixed, iteration boundary

**execute_loop (22):** streaming rejection, no tools, store=true persistence, single tool loop, parallel tools, mixed output, max iterations (status=incomplete), boundary, tool failure continues, no providers, empty output, multi-hop (3 iterations), LLM error propagated, Items input extended correctly, previous_response_id=None preserved, cassette replay from live vLLM, **previous_response_id=Some restored on Done**, **conversation_id restored on Done**, **store suppressed in internal iterations**, **all 3 persistence triggers cleared (multi-hop)**, **Incomplete path restores both IDs**, **Items input + conversation_id**

</details>

## MVP Routing (Step 1 limitation)

Currently `dispatch_tools` receives `(output, tool_ctx, iteration)` but NOT the request's `tools` array. This means:
- ALL `FunctionCall` items are treated as server-executable (routed to first configured provider)
- No distinction between `type: "function"` (client-side) and `type: "mcp"` (server-side)
- Client-side passthrough still works — just don't call `execute_loop` (use `execute()` directly)

Step 2 changes the signature to include the tools array and adds a `LoopDecision::Partial` variant for mixed client+server tool responses.

## Known Limitations

- **Non-streaming only** — StreamTee PR needed for streaming + tools
- **`InputItem` lacks FunctionCall variant** — model doesn't see its own tool-call items in re-entry context (type system change needed)
- **`request.clone()` per iteration** — `execute()` takes ownership; refactoring to `&RequestPayload` is out of scope (touches engine.rs internals)
- **MVP routing** — priority-order (MCP → web_search → vector_store), not by tool type from request

## References

- Issue #42 (core API proposal)
- PR #44 Phase 2 (design doc)
- PR #59 (function_call accumulation — prerequisite, merged)


crates/agentic-core/tests/cassettes/tool_loop/function-call-loop-vllm-gemma4.yaml
```diff
@@ -0,0 +1,43 @@
+# Recorded from google/gemma-4-26B-A4B-it via vLLM v0.21.0
+# Date: 2026-06-12
+# Scenario: User asks weather → model calls get_weather → tool returns result → model answers
+#
+# This cassette drives execute_loop through a complete 2-iteration tool loop:
+#   Iteration 0: model returns FunctionCall (get_weather)
+#   Iteration 1: model returns Message (final answer using tool result)
+
```
maralbahari, line 8: @ashwing is it possible to generate cassettes by serving openai and record the requests responses like https://github.com/vllm-project/agentic-api/blob/main/crates/agentic-core/tests/cassettes/text_only/conversation/conv-two-turn-gpt-4o-streaming.yaml
but instead of just text use function calls like `get_weather`.
the recording scripts can be found here https://github.com/vllm-project/agentic-api/blob/main/crates/agentic-core/tests/cassettes/record_cassette.py

crates/agentic-core/src/executor/execute_loop.rs
```diff
@@ -0,0 +1,198 @@
...
+                    ))
+                })??
+        };
+
+        // execute() returns Either<ResponsePayload, BoxStream>.
+        // We only support non-streaming in execute_loop (streaming requires StreamTee).
+        let mut payload = match result {
+            Either::Left(payload) => payload,
+            Either::Right(_stream) => {
+                return Err(crate::executor::ExecutorError::InvalidRequest(
```
maralbahari: why not importing on top of the file `use crate::executor::{ExecutorError}`

crates/agentic-core/src/executor/execute_loop.rs
```diff
@@ -0,0 +1,198 @@
...
+                );
+
+                // Build the input for the next iteration:
+                // existing context + tool results appended.
+                let mut items = match &request.input {
+                    // Already structured items — clone and extend.
+                    ResponsesInput::Items(existing) => existing.clone(),
+                    // First iteration with plain text — convert to structured item.
+                    ResponsesInput::Text(t) => {
+                        vec![crate::types::io::InputItem::Message(crate::types::io::InputMessage {
```
maralbahari: `crate::types::io::InputItem::Message(crate::types::io::InputMessage)` is making the line so long. could reduce the line by importing the objects on top of the file.

crates/agentic-core/src/executor/execute_loop.rs
```diff
@@ -0,0 +1,198 @@
...
+    // Capture before the loop mutates the request. Restored on final payload
+    // so the response chain (previous_response_id linkage) remains correct
+    // for the caller's persistence logic.
+    let original_previous_response_id = request.previous_response_id.clone();
+
+    // Suppress persistence for ALL iterations inside the loop. The caller
+    // (server handler) owns final persistence with the correct RequestContext.
+    // Without this, the first iteration would persist a partial response
+    // (containing only the tool-call output, not the final answer) to the DB.
+    request.store = false;
```
maralbahari, line 91: why is this hardcoded false? I see the comment 
```
// Without this, the first iteration would persist a partial response
 // (containing only the tool-call output, not the final answer) to the DB.
```
could you clarify what do you mean by partial response?

the contract is as following:
The `store` value from request payload coming in from server initially if is `false` we are passing into vllm proxy for stateless inference.
if is set to true we would invoke `execute_loop` which should not set this to false. from second turn onwards if the `store=False` context ID is given meaning either of (`conversation_id `or `previouse_response_id` is present) we need to still persist. if context ID not provided then we proxy to vLLM server stateless.

crates/agentic-core/src/executor/execute_loop.rs
```diff
@@ -0,0 +1,198 @@
...
+                    iteration,
+                    results = tool_results.len(),
+                    "tool results received, re-entering"
+                );
+
+                // Build the input for the next iteration:
+                // existing context + tool results appended.
+                let mut items = match &request.input {
+                    // Already structured items — clone and extend.
+                    ResponsesInput::Items(existing) => existing.clone(),
```
maralbahari: When existing is cloned, Rust preserves capacity but then extend may still trigger reallocation if tool_results.len() exceeds the slack. Can pre-size:

```
let mut items = Vec::with_capacity(existing.len() + tool_results.len());

File: crates/agentic-core/src/executor/execute_loop.rs [hunk 6]
```diff
@@ -0,0 +1,198 @@
...
+            LoopDecision::Incomplete(reason) => {
+                debug!(iteration, %reason, "loop incomplete");
+                payload.status = "incomplete".to_string();
+                payload.incomplete_details =
+                    Some(crate::types::request_response::IncompleteDetails { reason: Some(reason) });
+                payload.previous_response_id = original_previous_response_id;
+                return Ok(payload);
+            }
+            // Tools were executed — inject results and re-enter inference.
+            LoopDecision::Continue(tool_results) => {
```
maralbahari, line 157: Each iteration: clone → extend → reassign. On iteration 2, you clone the vec that already contains iteration 1's results. On iteration 3, you clone a vec with iterations 1+2's results. The allocation grows linearly and is thrown away each time.

The in-place alternative:

```
LoopDecision::Continue(tool_results) => {
    // Convert Text → Items on first tool call, then just extend in-place
    if let ResponsesInput::Text(t) = &request.input {
        request.input = ResponsesInput::Items(vec![InputItem::Message(...)]);
    }
    if let ResponsesInput::Items(ref mut v) = request.input {
        v.extend(tool_results);   // ← no clone, mutates the existing Vec
    }
}

File: crates/agentic-core/src/executor/execute_loop.rs [hunk 7]
```diff
@@ -0,0 +1,198 @@
...
+        }
+
+        debug!(iteration, "execute_loop iteration");
+
+        // --- Step 1: Call the LLM ---
+        // Timeout prevents hanging on unresponsive LLM backends.
+        // Duration::ZERO = no timeout (provider/reqwest manages its own).
+        let inference_timeout = exec_ctx.streaming_timeout;
+        let result = if inference_timeout.is_zero() {
+            execute(request.clone(), Arc::clone(&exec_ctx)).await?
```
maralbahari, line 111: to avoid frequent `request.clone()` let's change the `execute()` signature to borrow the RequestPayload `execute(request: &RequestPayload,...)`
this change would require also change in `rehydrate_conversation(request: &RequestPayload, ...)`

crates/agentic-core/src/executor/tool_context.rs
```diff
@@ -0,0 +1,164 @@
...
+    ///
+    /// # Retry Policy
+    ///
+    /// This layer does NOT retry. Providers handle their own retries internally
+    /// (transient network errors, 503s, etc.). By the time an error reaches here,
+    /// the provider already exhausted its retry budget. The agentic loop itself
+    /// serves as a higher-level retry — the model can re-call a failed tool on
+    /// the next iteration if it chooses to.
+    pub async fn execute_all(&self, calls: &[&FunctionToolCall]) -> Vec<InputItem> {
+        let futures: Vec<_> = calls.iter().map(|call| self.execute_one(call)).collect();
```
maralbahari: great choice using `future` just to optimize memory and performance here we can directly use:

```
futures::future::join_all(calls.iter().map(|call| self.execute_one(call))).await
```
since `join_all` accepts iterator. the intermediate step to create a vector is not needed.

crates/agentic-core/src/executor/dispatch.rs
```diff
@@ -0,0 +1,101 @@
...
+/// Returns `ExecutorError` only on internal failures (e.g. serialization).
+/// Individual tool execution failures are captured as error output strings
+/// in the returned `InputItem` list — they do NOT propagate as errors.
+pub async fn dispatch_tools(
+    output: &[OutputItem],
+    tool_ctx: &ToolContext,
+    iteration: usize,
+) -> ExecutorResult<LoopDecision> {
+    // Step 1: Extract FunctionCall items. Messages and Unknown are ignored.
+    let function_calls: Vec<_> = output
```
maralbahari: there is allocation before empty check which makes the allocation unnecessary.

we could optimize this by

```
let mut function_calls = output.iter().filter_map(|item| match item {
    OutputItem::FunctionCall(fc) => Some(fc),
    _ => None,
}).peekable();

if function_calls.peek().is_none() {
    return Ok(LoopDecision::Done);
}
if iteration >= tool_ctx.max_iterations {
    return Ok(LoopDecision::Incomplete(...));
}

let calls: Vec<_> = function_calls.collect();  // only alloc when we'll actually use it
tool_ctx.execute_all(&calls).await
```
This skips the `Vec` allocation on the two early-exit paths entirely.

crates/agentic-core/tests/cassettes/tool_loop/function-call-loop-vllm-gemma4.yaml
ashwing: Good call — yeah we can do that. I'll add a cassette recording mode in a follow-up PR that proxies to a real vLLM instance and dumps the request/response pairs. For this one I hand-crafted from an actual gemma4 session to keep the test deterministic without needing a live server.

crates/agentic-core/src/executor/execute_loop.rs
ashwing: Done — moved all crate imports to the top. Pushed.

crates/agentic-core/src/executor/execute_loop.rs
ashwing: Done — imported InputItem, InputMessage, InputMessageContent, and ResponsesInput at the top. Much cleaner now.

crates/agentic-core/src/executor/execute_loop.rs
ashwing: The loop suppresses persistence for intermediate iterations — otherwise each internal execute() call would write a partial response (just tool-call output, not the final answer) to the store. We clear all three persistence triggers (store, previous_response_id, conversation_id) because PR #56 persists when any of them is set. The caller (server handler) does the final persist once with the complete response.

Updated the doc comment to make this clearer.

crates/agentic-core/src/executor/execute_loop.rs
ashwing: Done — switched to in-place mutation with items.reserve(tool_results.len()) + items.extend(tool_results). No clone at all now, just grows the existing vec.

crates/agentic-core/src/executor/execute_loop.rs
ashwing: Fixed — now mutates request.input in-place. Text gets converted to Items once on the first tool call, then subsequent iterations just reserve+extend on the same vec. No cloning.

crates/agentic-core/src/executor/execute_loop.rs
ashwing: Agree this is the right long-term fix. The request.clone() is still there because execute() takes ownership today (it passes the payload into rehydrate_conversation which moves it). Changing that signature touches engine.rs internals (rehydrate, inject_ids, persist) which are outside our PR scope.

I'll open a follow-up to refactor execute() to borrow — it'll be a clean diff once this lands. For now the clone cost is bounded by the input size which doesn't grow per-iteration anymore (tool results are just appended, not cloned on each loop).

crates/agentic-core/src/executor/tool_context.rs
ashwing: Yep simplified — just passes the iterator directly into join_all now. No intermediate collect.

crates/agentic-core/src/executor/dispatch.rs
ashwing: Good catch — switched to a .peekable() iterator. peek() checks for emptiness without allocating, and we only collect into a Vec when we know we'll actually execute. Pushed.

crates/agentic-core/src/executor/execute_loop.rs
ashwing: Happy to discuss further, esp. the requirement to persist intermediate tool call/results and the layer responsible to persist.

noobHappylife:
Hi @ashwing, 

Thank you for this PR. I'm wondering, would it be easier to split the function call into another PR? Maybe this PR can focus on the server-side execution loop, and we handle the basic function call support separately.

For function call support, I think the flow should be:

* client sends request with tools
* `agentic-api` passes the tools to upstream vLLM
* upstream vLLM returns the `tool_calls` signature
* `agentic-api` returns those `tool_calls` back to the client
* client executes the tools and sends back the tool results
* `agentic-api` passes the tool results back to upstream vLLM
* upstream returns the final response

Then we can record cassettes to test the complete function call cycle:

`client sends request with tool(s)` → `agentic-api` returns `tool_call(s)` → client executes and sends `tool_call_result(s)` → `agentic-api` returns final result.

Once this is properly set up, the server-side execution loop should be mostly the same, except `agentic-api` becomes the executor for the `tool_call(s)` instead of the client.

Additionally, I think there might be some semantic confusion around how LLM function calling works.

For most OSS models, especially when served through vLLM, the model-facing tool interface is basically function calling (`type: "function"`). So for other tool types like `mcp`, `web_search`, etc., I don't think we should pass them directly to the model/template unless the backend explicitly supports it. `agentic-api` should internally normalize or lower those tools into function-compatible schemas so the LLM can understand them, while keeping the metadata needed to route the tool call back to the correct executor.

For example:

* `type: "function"` can be passed through or lightly normalized
* `type: "mcp"` should probably be resolved into MCP server tools, then exposed to the model as function-compatible tools
* hosted tools like `web_search` should either be implemented by `agentic-api`, proxied, or rejected as unsupported (if we haven't implemented it)

I think it might be good to first align on the function call loop and semantics, then build the server-side execution on top of that. This should make the implementation easier to test and reason about.

ashwing:
Hey @noobHappylife, thanks for the detailed breakdown — really helpful to see the full picture laid out like this.

I think we're actually in agreement on the architecture. Let me map your flow to what's here:

**Client-side passthrough already works today.** When a request hits `execute()` in engine.rs without an `execute_loop` caller, vLLM returns `function_call` output items and they pass straight through in the ResponsePayload to the client. The client executes, sends results back as `function_call_output` input items in the next request, and the cycle continues. That path doesn't need `execute_loop` at all — it's just the stateless proxy doing its job.

**What this PR adds is the server-side layer on top.** When `agentic-api` itself needs to execute tools (MCP servers, web_search, code_interpreter), `execute_loop` orchestrates that. The key is: `dispatch_tools` inspects the output and decides whether to execute server-side or stop. If there are no server-executable tool calls → `LoopDecision::Done` → the response goes back to the client as-is (same as your passthrough flow).

So the two flows are:
- No `execute_loop` (or `ToolContext` with no providers): pure passthrough, client does everything
- With `execute_loop` + configured providers: gateway executes what it can, returns the rest to client

**On type-aware routing — fully agree.** The dispatch code has an explicit MVP note (line 53-56) saying "the distinction between client-side functions and gateway-executed tools requires access to the request's tools array — deferred to a follow-up." That's exactly what you're describing. The follow-up would:
1. Pass the tools array into `dispatch_tools`
2. Classify each FunctionCall by its original tool type
3. Gateway types (web_search, code_interpreter, mcp) → execute server-side → `Continue`
4. Client types (function, namespace, tool_search, custom) → don't execute → `Done` with `status: requires_action`

**On normalization before vLLM** — also agree. MCP tools should be resolved into function-compatible schemas before the model sees them. vLLM only speaks function calling. That normalization layer sits upstream of `execute_loop` (during request enrichment) and is independent of this PR's dispatch mechanics.

I think the sequencing works as:
1. **This PR** — server-side loop infrastructure (execute_loop + dispatch + ToolContext)
2. **Next** — type-aware dispatch (classify by tool type, split client vs gateway)
3. **Then** — tool normalization (MCP → function schema, namespace → flatten)

Each layer builds on the previous. Does this framing make sense to you? Happy to discuss further — especially the normalization step since you've clearly thought through the Codex integration side of this.

maralbahari:
> I think the sequencing works as:
> 
> 1. **This PR** — server-side loop infrastructure (execute_loop + dispatch + ToolContext)
> 2. **Next** — type-aware dispatch (classify by tool type, split client vs gateway)
> 3. **Then** — tool normalization (MCP → function schema, namespace → flatten)
> 
> Each layer builds on the previous. Does this framing make sense to you? Happy to discuss further — especially the normalization step since you've clearly thought through the Codex integration side of this.

@ashwing I think we could have one PR to add built-in `function_call` s first?.
 now we have https://github.com/vllm-project/agentic-api/pull/57 which is handling reasnoning output we need to support other output types which are result of `function_call` like:
```
 input prompt : "how is the weather in France "             
    "type": "function",
    "name": "get_weather"
```
or
```
  input prompt: "what is the answer to 2+2"
  "type":"function",
  name:"calculator"
```
where model's output would contain a json like
```
    "output": [
        {
            "arguments": "",
            "call_id": "chatcmpl-tool-aaff0fbfaabeb6e2",
            "name": "function_name ",
            "type": "function_call",
            "id": "fc_94af73a42c3d05c0",
            "namespace": null,
            "status": "completed"
        }
    ],
 ``` 

after we have the PR supporting function_call and the reasnoning in  https://github.com/vllm-project/agentic-api/pull/57 then we can go with 
> 1. **This PR** — server-side loop infrastructure (execute_loop + dispatch + ToolContext)
> 2. **Next** — type-aware dispatch (classify by tool type, split client vs gateway)
> 3. **Then** — tool normalization (MCP → function schema, namespace → flatten)

cause I feel for the server-side loop we should at least add the function_call described as above would make the loop more solid since we are at least supporting the built-in calls. 
let me know your thoughts, thanks

ashwing:
@maralbahari Good point — you're right that the accumulator doesn't actually produce `OutputItem::FunctionCall` from the SSE stream today. The normalizer classifies `function_call_arguments.delta/done` events and extracts the payload, but the accumulator's `process_event` falls through to the wildcard for those event types. So in streaming mode, function calls would never appear in `payload.output` for dispatch to find.

I can add function_call accumulation to this PR — it follows the same pattern as PR #57's reasoning items:
- Track `current_function_call: Option<FunctionToolCall>` + `accumulated_arguments: String` on the accumulator
- `OutputItemAdded` with `item_type == "function_call"` → start a new in-flight function call (using `name`/`call_id` from the event)
- `FunctionCallArgumentsDelta` → append to accumulated_arguments
- `FunctionCallArgumentsDone` → finalize with full arguments string, push `OutputItem::FunctionCall` to output
- `finalize_current_function_call()` for the partial-disconnect case

That way the loop has real function_call items to dispatch on in both blocking and streaming paths. Should be ~50 lines following the existing accumulation pattern.

Does that work for you, or would you prefer it as a separate PR that lands first?

maralbahari:
> @maralbahari Good point — you're right that the accumulator doesn't actually produce `OutputItem::FunctionCall` from the SSE stream today. The normalizer classifies `function_call_arguments.delta/done` events and extracts the payload, but the accumulator's `process_event` falls through to the wildcard for those event types. So in streaming mode, function calls would never appear in `payload.output` for dispatch to find.
> 
> I can add function_call accumulation to this PR — it follows the same pattern as PR #57's reasoning items:
> 
> * Track `current_function_call: Option<FunctionToolCall>` + `accumulated_arguments: String` on the accumulator
> * `OutputItemAdded` with `item_type == "function_call"` → start a new in-flight function call (using `name`/`call_id` from the event)
> * `FunctionCallArgumentsDelta` → append to accumulated_arguments
> * `FunctionCallArgumentsDone` → finalize with full arguments string, push `OutputItem::FunctionCall` to output
> * `finalize_current_function_call()` for the partial-disconnect case
> 
> That way the loop has real function_call items to dispatch on in both blocking and streaming paths. Should be ~50 lines following the existing accumulation pattern.
> 
> Does that work for you, or would you prefer it as a separate PR that lands first?

@ashwing thanks I think separate PR would be easier to review and land in first.

ashwing:
@maralbahari Opened #59 with the function_call accumulation as a separate PR per your suggestion. Once that lands, this PR's execute_loop will have `FunctionCall` items available in the streaming path for dispatch.

ashwing:
@maralbahari PR #59 is merged — rebased this on top of it. Ready for re-review. Updated the PR description with the sequencing we discussed.

maralbahari:
> @maralbahari PR #59 is merged — rebased this on top of it. Ready for re-review. Updated the PR description with the sequencing we discussed.

@ashwing Thank you for changes. as per discussed in community meeting today. we should hold on this PR since the  agentic `execution_loop` cannot be completed as off now. the tools are not implemented in this PR they're just place stub functions. first step would be get MCP in.

ashwing:
Closing this in favor of https://github.com/vllm-project/agentic-api/pull/67

PR #52, opened 2026-06-11
refactor: accumulator consumes EventFrame instead of inline JSON

ashwing:
## Summary

Refactors `ResponseAccumulator::process_sse_line` to use the typed `EventFrame` from the `events/` module (PR #49) instead of inline JSON parsing. Pure behavioral refactor — same output for same input.

**Changes:**
- `process_sse_line` is now a thin wrapper: calls `normalize_sse_line()` → `process_event()`
- New `pub(crate) fn process_event(&EventFrame)` — typed matching on `(SSEEventType, EventPayload)` pairs
- Removes inline `serde_json::Value` field access, old `SSEEventType` import, `deserialize_from_value` import

**Why:** Enables `StreamTee` (future) to call `process_event` directly with pre-normalized frames — avoiding double-parsing when forwarding SSE to client while accumulating for tool detection.

**No behavioral change** — existing cassette-based integration tests pass unchanged.

## Test Plan

- All existing tests pass unchanged (cassette-based streaming + non-streaming)
- 5 new unit tests for `process_event` directly:
  - `ResponseCreated` sets response_id
  - `ResponseCreated` with empty id doesn't overwrite
  - `TextDelta` accumulates and attaches to message
  - `ResponseCompleted` extracts usage
  - Unknown events silently ignored
- `cargo clippy -- -D warnings` clean
- 160+ workspace tests pass

### DIFF
--- a/crates/agentic-core/src/executor/accumulator.rs
+++ b/crates/agentic-core/src/executor/accumulator.rs
@@ -325,4 +325,144 @@ mod tests {
         assert_eq!(MessageStatus::Completed.as_str(), "completed");
         assert_eq!(MessageStatus::InProgress.as_str(), "in_progress");
     }
+
+    // --- process_event tests (exercises the refactored path directly) ---
+
+    /// Feeding a `ResponseCreated` `EventFrame` sets the `response_id` on the accumulator.
+    #[test]
+    fn test_process_event_response_created_sets_id() {
+        let mut acc = ResponseAccumulator::new("resp_old".into(), None);
+        let frame = EventFrame {
+            event_type: SSEEventType::ResponseCreated,
+            payload: EventPayload::Response {
+                id: "resp_new".into(),
+                status: "in_progress".into(),
+                usage: None,
+            },
+            sequence_number: Some(0),
+        };
+
+        acc.process_event(&frame);
+        assert_eq!(acc.response_id, "resp_new");
+    }
+
+    /// `ResponseCreated` with empty id should NOT overwrite the existing `response_id`.
+    #[test]
+    fn test_process_event_response_created_empty_id_no_overwrite() {
+        let mut acc = ResponseAccumulator::new("resp_keep".into(), None);
+        let frame = EventFrame {
+            event_type: SSEEventType::ResponseCreated,
+            payload: EventPayload::Response {
+                id: String::new(),
+                status: "in_progress".into(),
+                usage: None,
+            },
+            sequence_number: Some(0),
+        };
+
+        acc.process_event(&frame);
+        assert_eq!(acc.response_id, "resp_keep");
+    }
+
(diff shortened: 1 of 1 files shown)


crates/agentic-core/src/executor/accumulator.rs
```diff
@@ -178,45 +179,43 @@ impl ResponseAccumulator {
...
-                }
+    /// Processes a typed [`EventFrame`], updating accumulator state.
+    ///
+    /// This is the core state machine — callers that already have a normalized
+    /// frame (e.g. [`StreamTee`](future)) can call this directly without
+    /// re-parsing from a raw line.
+    pub(crate) fn process_event(&mut self, frame: &EventFrame) {
+        match (&frame.event_type, &frame.payload) {
+            (SSEEventType::ResponseCreated, EventPayload::Response { id, .. }) if !id.is_empty() => {
+                self.response_id.clone_from(id);
```
maralbahari, line 195: is the `cloning` necessary? can it be avoided?

crates/agentic-core/src/executor/accumulator.rs
```diff
@@ -178,45 +179,43 @@ impl ResponseAccumulator {
...
+            (SSEEventType::OutputItemAdded, EventPayload::OutputItemAdded { item_id, .. }) => {
                 self.finalize_current_message();
-                let item_id = json["item"]["id"]
-                    .as_str()
-                    .map_or_else(|| uuid7_str("msg_"), str::to_string);
-                self.current_message = Some(OutputMessage::new(&item_id, MessageStatus::InProgress.as_str()));
+                let id = if item_id.is_empty() {
+                    uuid7_str("msg_")
+                } else {
+                    item_id.clone()
```
maralbahari, line 202: is `item_id.clone()` necessary here ? also could write in a single line instead of if/else braces.

maralbahari (approved):
@ashwing thank you for the refactor. just left some nit comments.

crates/agentic-core/src/executor/accumulator.rs
ashwing: Yeah the clone is unavoidable here since we borrow `&EventFrame` (so `StreamTee` can inspect the frame after). `clone_from` at least reuses the existing buffer. If we later find accumulator is the sole consumer we can switch to taking the frame by value and destructure instead.

crates/agentic-core/src/executor/accumulator.rs
ashwing: Same deal — need an owned String since we borrow the frame. Removed the `&id` in the `OutputMessage::new` call though (72e430a) so the String moves in directly instead of re-allocating from a ref. Re single line — rustfmt expands it back at our max_width so left it as-is.

franciscojavierarceo (approved):

Issue #53, opened 2026-06-15
`agentic-core/storage` handle concurrent access in DB transaction in `persist`.


maralbahari:
two simultaneous requests with the same `conversation_id` both rehydrate the same `item_ids` snapshot, both infer, both persist. The second `persist` overwrites — first request's items get lost from the ordered list.

Not a bug today (single-user, low concurrency). Worth a TODO for when multi-user or high-concurrency scenarios arise. Options: optimistic locking (version column), or append-only with conflict detection.

_Originally posted by @ashwing in https://github.com/vllm-project/agentic-api/pull/48#discussion_r3398502552_

PR #55, opened 2026-06-16
fix: `agentic-core` concurrency in db layer include benchmarks.

maralbahari:
# Summary

Fixes a sequence-number race in `ConversationStore::persist` mentioned in https://github.com/vllm-project/agentic-api/issues/53 and adds concurrent storage benchmarks to `agentic-core`. 
**Race fix (`storage/conversation.rs` + `storage/models/item.rs`):**

`conversation_item_count` previously ran outside the transaction, so two simultaneous requests with the same `conversation_id` could both snapshot the same `MAX(seq)` and overwrite each other's items. The fix moves the count read inside the transaction so the DB serializes the snapshot and the second writer always sees the first writer's committed seq.

**SQLite pool fix (`storage/pool.rs`):**

SQLite only supports one writer at a time. Setting `max_connections(1)` for SQLite pools serializes writes at the pool-queue level. Postgres/MySQL keep `max_connections(10)`.

**New benchmarks (`benches/storage_concurrent.rs`):**

| Group | What it measures |
|---|---|
| `concurrent_conversation_persist` | N writers to the **same** `conversation_id` |
| `concurrent_conversation_persist_independent` | N writers each to their own `conversation_id` |
| `concurrent_conversation_rehydrate` | N readers of the **same** `conversation_id` |
| `concurrent_conversation_rehydrate_independent` | N readers each on their own `conversation_id` |
| `concurrent_response_persist` | N writers each to their own `response_id` |
| `concurrent_response_rehydrate` | N readers of the **same** `response_id` |

Concurrency levels default to `[2, 4, 8, 16]`, configurable via `BENCH_CONCURRENCY=2,4,8,16,32`.

## Test Plan

`test_conversation_concurrent_turns` (existing, now fixed): two tasks persist to the same conversation concurrently; asserts both succeed and the rehydrated item count is correct. Runs on a `multi_thread` tokio runtime with `worker_threads = 2`.

```bash
cargo test -p agentic-core
cargo bench --bench benches -- concurrent_
BENCH_CONCURRENCY=2,4,8,16 cargo bench -p agentic-core --bench benches -- concurrent_
```

All 152 tests pass across the workspace.

## Benchmark Results

> SQLite · `sample-size=100` · concurrency 2-16

### `concurrent_conversation_persist` (same `conversation_id`)

| N | Median | Throughput |
|:-:|-------:|----------:|
| 2  | 56.6 ms | 35.3 elem/s |
| 4  | 200 ms  | 20.0 elem/s |
| 8  | 163 ms  | 49.1 elem/s |
| 16 | 569 ms  | 28.1 elem/s |

### `concurrent_conversation_persist_independent` (N distinct conversations)

| N | Median | Throughput |
|:-:|-------:|----------:|
| 2  | 70.7 ms | 28.3 elem/s |
| 4  | 229 ms  | 17.5 elem/s |
| 8  | 507 ms  | 15.8 elem/s |
| 16 | 756 ms  | 21.2 elem/s |

### `concurrent_conversation_rehydrate` (same `conversation_id`)

| N | Median | Throughput |
|:-:|-------:|----------:|
| 2  | 305 µs | 6.6 Kelem/s |
| 4  | 624 µs | 6.4 Kelem/s |
| 8  | 1.30 ms | 6.2 Kelem/s |
| 16 | 2.74 ms | 5.8 Kelem/s |

### `concurrent_conversation_rehydrate_independent`

| N | Median | Throughput |
|:-:|-------:|----------:|
| 2  | 299 µs | 6.7 Kelem/s |
| 4  | 610 µs | 6.6 Kelem/s |
| 8  | 1.33 ms | 6.0 Kelem/s |
| 16 | 2.71 ms | 5.9 Kelem/s |

### `concurrent_response_persist` (independent response IDs)

| N | Median | Throughput |
|:-:|-------:|----------:|
| 2  | 35.8 ms | 55.8 elem/s |
| 4  | 176 ms  | 22.7 elem/s |
| 8  | 163 ms  | 48.9 elem/s |
| 16 | 372 ms  | 43.0 elem/s |

### `concurrent_response_rehydrate` (same `response_id`)

| N | Median | Throughput |
|:-:|-------:|----------:|
| 2  | 591 µs | 3.4 Kelem/s |
| 4  | 1.19 ms | 3.4 Kelem/s |
| 8  | 2.50 ms | 3.2 Kelem/s |
| 16 | 4.81 ms | 3.3 Kelem/s |

### Analysis

**Reads scale linearly with N.** Both rehydrate groups show near-constant throughput (~6 Kelem/s for conversations, ~3.3 Kelem/s for responses) across all concurrency levels, confirming reads do not contend on SQLite's single-connection pool.

**Writes show high variance at N>=4.** The large confidence intervals on persist groups (especially `independent/16`: 557 ms-1.01 s) reflect SQLite pool-queue serialization cost accumulating as N grows. This is expected on SQLite; numbers will improve on Postgres where writers do not serialize.

**Same-target vs independent write throughput is comparable on SQLite** because the pool already serializes all writers through one connection. The distinction matters on Postgres where independent writes can proceed in parallel but same-`conversation_id` writes must still serialize to maintain seq ordering.

> **Note:** Concurrent write correctness has only been validated on SQLite. Postgres testing is out of scope for this PR. A follow-up should run `test_conversation_concurrent_turns` and the `concurrent_conversation_persist` benchmark groups against a real Postgres instance to verify the transaction fix holds under true parallel writers.

### DIFF
--- a/crates/agentic-core/benches/benches.rs
+++ b/crates/agentic-core/benches/benches.rs
@@ -1,6 +1,11 @@
 mod executor_throughput;
+mod storage_concurrent;
 mod storage_crud;
 
 use criterion::criterion_main;
 
-criterion_main!(storage_crud::storage_benches, executor_throughput::executor_benches);
+criterion_main!(
+    storage_crud::storage_benches,
+    storage_concurrent::storage_concurrent_benches,
+    executor_throughput::executor_benches
+);
--- a/crates/agentic-core/benches/storage_concurrent.rs
+++ b/crates/agentic-core/benches/storage_concurrent.rs
@@ -1,0 +1,421 @@
+//! Concurrent storage benchmarks for [`ConversationStore`] and [`ResponseStore`].
+//!
+//! Each benchmark group spawns N tasks simultaneously to measure throughput and
+//! contention under concurrent load.
+//!
+//! | Group                                        | What it measures                                     |
+//! |----------------------------------------------|------------------------------------------------------|
+//! | `concurrent_conversation_persist`            | N writers to the **same** conversation (seq contention) |
+//! | `concurrent_conversation_persist_independent`| N writers each to their **own** conversation         |
+//! | `concurrent_conversation_rehydrate`          | N readers of the **same** conversation               |
+//! | `concurrent_conversation_rehydrate_independent` | N readers each on their **own** conversation      |
+//! | `concurrent_response_persist`               | N independent response writes in parallel            |
+//! | `concurrent_response_rehydrate`             | N readers of the **same** response                  |
+//!
+//! # Configuring concurrency levels
+//!
+//! Set `BENCH_CONCURRENCY` to a comma-separated list of integers before running.
+//! Defaults to `2,4,8,16` when unset.
+//!
+//! ```bash
+//! # Run with default concurrency levels (2, 4, 8, 16)
+//! cargo bench --bench storage_concurrent
+//!
+//! # Run with custom concurrency levels
+//! BENCH_CONCURRENCY=2,4,8,16,32 cargo bench --bench storage_concurrent
+//!
+//! # Run a single group
+//! cargo bench --bench storage_concurrent -- concurrent_conversation_persist
+//! ```
+
+use std::sync::Arc;
+
+use criterion::{BenchmarkId, Criterion, Throughput, black_box, criterion_group};
+use tokio::runtime::Runtime;
+
+use agentic_core::storage::{ConversationStore, InOutItem, ResponseMetadata, ResponseStore, create_pool_with_schema};
+use agentic_core::types::io::{InputItem, InputMessage, InputMessageContent, OutputItem, OutputMessage};
+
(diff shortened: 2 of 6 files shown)


leseb (approved):
that

> so two simultaneous requests with the same conversation_id

seems very unlikely?

maralbahari:
> that
> 
> > so two simultaneous requests with the same conversation_id
> 
> seems very unlikely?

yes it is. since @ashwing been trying to catch concurrency issues. would be fine to catch those issues in case.
also the fix of single connection pool for SQLite case in this PR is proper fix to not run into problems later on.

leseb:
> > that
> > > so two simultaneous requests with the same conversation_id
> > 
> > 
> > seems very unlikely?
> 
> yes it is. since @ashwing been trying to catch concurrency issues. would be fine to catch those issues in case. also the fix of single connection pool for SQLite case in this PR is proper fix to not run into problems later on.

agreed on the concurrency 1 for sqlite

PR #57, opened 2026-06-16
feat: support reasoning output items

franciscojavierarceo:
**Summary**
- Add typed reasoning input/output item support for vLLM Responses payloads.
- Accumulate streamed `response.reasoning_text.*` events into `reasoning` output items.
- Preserve reasoning items during stateful rehydration so follow-up requests sent through Agentic API keep the full vLLM Responses history.
- Add storage disambiguation for input vs output items that share the same wire shape.

**Context**
We found this while testing Agentic API against a local vLLM server through the Responses API. Models that emit reasoning return output like `{ "type": "reasoning", ... }` before the assistant message. The previous typed model did not represent those items, so they could be dropped or misclassified during parsing, streaming accumulation, and rehydration.

For stateful flows, this matters on the next request: Agentic API reconstructs the prior response history and forwards it to vLLM as `input`. vLLM expects the reasoning items, including their `content`, to be present in that reconstructed history. Without preserving them, follow-up requests can lose model reasoning context or fail compatibility with vLLM's Responses implementation.

Example captured from local vLLM at `http://10.0.0.99:8000/v1/responses` with `openai/gpt-oss-20b`:

```json
{
  "status": "completed",
  "output": [
    {
      "id": "rs_b4371fd8b4767c3f",
      "type": "reasoning",
      "status": null,
      "content": [
        {
          "type": "reasoning_text",
          "text": "The user asks: \"What is 2+2? Answer briefly.\" We can answer succinctly: \"4.\""
        }
      ]
    },
    {
      "id": "msg_944a5458f4b5be23",
      "type": "message",
      "status": "completed",
      "content": [
        {
          "type": "output_text",
          "text": "4.",
          "annotations": []
        }
      ]
    }
  ]
}
```

This PR makes reasoning items first-class in the core types, handles both non-streaming JSON and streaming SSE forms, and adds regression coverage for done-only reasoning text, message/reasoning item ordering, and storage round-tripping.

**Test Plan**
- `cargo fmt -- --check`
- `cargo test`
- Manually tested against local vLLM Responses API with reasoning output, including non-streaming, streaming, and stateful follow-up rehydration

### DIFF
--- a/crates/agentic-core/src/executor/accumulator.rs
+++ b/crates/agentic-core/src/executor/accumulator.rs
@@ -15,7 +15,9 @@ use futures::{Stream, StreamExt};
 use crate::events::{EventFrame, EventPayload, SSEEventType, normalize_sse_line};
 use crate::executor::error::{ExecutorError, ExecutorResult};
 use crate::types::event::{MessageStatus, ResponseStatus};
-use crate::types::io::{OutputItem, OutputMessage, OutputTextContent, ResponseUsage};
+use crate::types::io::{
+    OutputItem, OutputMessage, OutputTextContent, ReasoningOutput, ReasoningTextContent, ResponseUsage,
+};
 use crate::types::request_response::{IncompleteDetails, ResponsePayload};
 use crate::utils::common::{deserialize_from_str, deserialize_from_value_opt};
 use crate::utils::uuid7_str;
@@ -32,6 +34,9 @@ pub struct ResponseAccumulator {
     // In-flight message state — owned here so process_sse_line takes only &mut self.
     current_message: Option<OutputMessage>,
     accumulated_text: String,
+    // In-flight reasoning state.
+    current_reasoning: Option<ReasoningOutput>,
+    accumulated_reasoning_text: String,
 }
 
 impl ResponseAccumulator {
@@ -47,6 +52,8 @@ impl ResponseAccumulator {
             incomplete_details: None,
             current_message: None,
             accumulated_text: String::new(),
+            current_reasoning: None,
+            accumulated_reasoning_text: String::new(),
         }
     }
 
@@ -90,6 +97,8 @@ impl ResponseAccumulator {
             incomplete_details: None,
             current_message: None,
             accumulated_text: String::new(),
+            current_reasoning: None,
+            accumulated_reasoning_text: String::new(),
         })
     }
 
@@ -141,6 +150,7 @@ impl ResponseAccumulator {
         for line in rx {
             acc.process_sse_line(&line);
         }
+        acc.finalize_current_reasoning();
         acc.finalize_current_message();
         if acc.status == ResponseStatus::InProgress {
             acc.status = ResponseStatus::Completed;
@@ -159,8 +169,22 @@ impl ResponseAccumulator {
         for line in lines {
             acc.process_sse_line(&line);
         }
+        acc.finalize_current_reasoning();
         acc.finalize_current_message();
         acc
+    }
+
+    /// Closes the in-flight reasoning item, pushing it to `output` with accumulated text.
+    fn finalize_current_reasoning(&mut self) {
+        if let Some(mut reasoning) = self.current_reasoning.take() {
+            if !self.accumulated_reasoning_text.is_empty() {
+                reasoning
+                    .content
+                    .push(ReasoningTextContent::new(&self.accumulated_reasoning_text));
+            }
+            self.output.push(OutputItem::Reasoning(reasoning));
+        }
+        self.accumulated_reasoning_text.clear();
     }
 
     /// Closes the in-flight message, pushing it to `output` with accumulated text.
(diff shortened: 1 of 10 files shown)


ashwing (commented):
nice work — left inline comments on storage compat, accumulator invariants, and enum consolidation.

crates/agentic-core/src/storage/types/item.rs
```diff
@@ -3,19 +3,38 @@
...
+impl ItemKind {
+    pub(crate) const fn as_str(self) -> &'static str {
+        match self {
+            Self::Input => "input",
+            Self::Output => "output",
+        }
+    }
+
+    pub(crate) fn from_stored_str(value: &str) -> Option<Self> {
+        match value {
```
ashwing, line 30: nice approach with `_agentic_item_kind` — solves the deser ambiguity cleanly.

one thing: existing rows in the DB won't have this tag, so `stored_item_kind()` returns None and we fall through to the old "try input, then try output" heuristic. that works for messages and function_call_outputs since they don't have this ambiguity. but if someone migrates the DB forward and then rolls back, you'd have rows with the tag that old code doesn't understand — the `_agentic_item_kind` key would pass through as an unrecognized field in serde. i think that's fine (serde default is deny_unknown_fields=false), just want to confirm that's intentional.

also — worth a `#[cfg(test)]` that round-trips a pre-tag row (just serialized without the tag injection) through `as_inout()` to prove backward compat in the test suite?

crates/agentic-core/src/executor/accumulator.rs
```diff
@@ -159,10 +169,24 @@ impl ResponseAccumulator {
         for line in lines {
             acc.process_sse_line(&line);
         }
+        acc.finalize_current_reasoning();
         acc.finalize_current_message();
         acc
     }
 
+    /// Closes the in-flight reasoning item, pushing it to `output` with accumulated text.
```
ashwing, line 177: edge case worth confirming: when the backend sends deltas AND a non-empty `reasoning_text.done`, the done-text is used and the accumulated delta buffer is cleared. so done-text is authoritative — deltas are for streaming display only. that matches openai's spec where the done event carries the full final text.

the partial-disconnect case also looks good: if vLLM drops mid-reasoning without sending done, `finalize_current_reasoning()` catches whatever was accumulated via deltas. partial reasoning is better than silent loss.

looks correct — just calling it out since it's a non-obvious correctness invariant.

crates/agentic-core/src/types/event.rs
```diff
@@ -98,6 +98,12 @@ pub enum SSEEventType {
     /// Text delta; incremental token content added to the current message.
     ResponseOutputTextDelta,
 
+    /// Reasoning text delta; incremental reasoning content.
+    ResponseReasoningTextDelta,
+
```
ashwing: noticed this adds `ResponseReasoningTextDelta` and `ResponseReasoningTextDone` to this enum, but the accumulator uses the canonical `events::types::SSEEventType` (the one with the full set of variants + `EventPayload` dispatch).

this `types::event::SSEEventType` is now only exercised by its own module tests + `FromStr` impl — nothing else imports it for event classification. might be worth a follow-up to consolidate these two enums so we don't need to keep them in sync. not blocking for this PR though.

crates/agentic-core/src/storage/types/item.rs
franciscojavierarceo: Yeah, the unknown-field behavior is intentional: older code will ignore `_agentic_item_kind` under serde’s default behavior.

I don’t think we need a pre-tag row compatibility test here since we haven’t cut a dev release yet. The marker is mainly for current correctness: `InputItem::Reasoning` and `OutputItem::Reasoning` have the same wire shape, so new stored rows need an explicit direction marker to avoid ambiguous `as_inout()` classification.

crates/agentic-core/src/executor/accumulator.rs
franciscojavierarceo: Yep, that’s the intended invariant: `reasoning_text.done` is authoritative when present, and accumulated deltas are only used as the fallback if the stream ends before done arrives.

We have coverage for both paths: delta+done, done-only, and finalizing partial accumulated reasoning.

crates/agentic-core/src/types/event.rs
franciscojavierarceo: Agreed. I removed the new reasoning variants from the legacy `types::event::SSEEventType`; the accumulator now only relies on the canonical `events::types::SSEEventType` path for reasoning events.

I'll consolidate the remaining duplicate enum in a follow up, but this PR no longer expands the legacy one.

franciscojavierarceo:
> nice work — left inline comments on storage compat, accumulator invariants, and enum consolidation.

thanks for the review! responded to your feedback

maralbahari:
@franciscojavierarceo nice work. code is clean nothing on that side. I just thought we should keep cassettes from vLLM upstream so that in future we can test everything from unit-test without manually testing everything during development. so I opened this PR https://github.com/vllm-project/agentic-api/pull/58 I recorded cassettes from vLLM upstream for two models `gpt-oss` and `qwen3` so that you can use them in your PR unit test.

ashwing (approved):
Looks good — clean implementation, all three points addressed well. The enum consolidation follow-up makes sense as separate scope.

One note: @maralbahari's cassettes in #58 (qwen3 with reasoning) could be a nice addition to your test suite here if you want to add a cassette-driven reasoning accumulation test before merge. Not blocking though.

franciscojavierarceo:
Will incorporate that as a follow up 👍

leseb (approved):

PR #58, opened 2026-06-17
test support: add reasoning cassettes for vLLM responses mode

maralbahari:
## Summary
- Adds `record_reasoning_cassettes.sh` to record single-turn cassettes from a vLLM backend via `--mode responses --vllm`
- Captures two cassette variants: non-streaming (reasoning embedded inline as `<think>` in `output_text`) and streaming (reasoning as a separate `response.reasoning_text.delta` SSE stream with its own `reasoning` output item)
- Recorded against `Qwen/Qwen3-30B-A3B-FP8` and `openai/gpt-oss-20b` at `http://0.0.0.0:5050`

This recorded cassettes in this PR assist to test https://github.com/vllm-project/agentic-api/pull/57

## How to Record

**1. Start vLLM:**

```bash
# gpt-oss-20b
vllm serve openai/gpt-oss-20b --port 5050 > server.log 2>&1

# Qwen3 (with reasoning parser)
vllm serve Qwen/Qwen3-30B-A3B-FP8 \
    --reasoning-parser deepseek_r1 \
    --port 5050 > server.log 2>&1
```

**2. Run the cassette recorder:**

```bash
VLLM_URL=http://0.0.0.0:5050 MODEL=openai/gpt-oss-20b bash tests/cassettes/record_reasoning_cassettes.sh
VLLM_URL=http://0.0.0.0:5050 MODEL=Qwen/Qwen3-30B-A3B-FP8 bash tests/cassettes/record_reasoning_cassettes.sh
```

### DIFF
--- a/crates/agentic-core/tests/cassettes/reasoning/responses/reasoning-single-Qwen-Qwen3-30B-A3B-FP8-nonstreaming.yaml
+++ b/crates/agentic-core/tests/cassettes/reasoning/responses/reasoning-single-Qwen-Qwen3-30B-A3B-FP8-nonstreaming.yaml
@@ -1,0 +1,99 @@
+turns:
+- filename: t1
+  request:
+    body:
+      input: 'Reply with exactly one word: HELLO'
+      model: Qwen/Qwen3-30B-A3B-FP8
+      store: true
+      stream: false
+    headers:
+      accept: '*/*'
+      content-type: application/json
+      user-agent: python-httpx/0.28.1
+    method: POST
+    path: /v1/responses
+    query_params: {}
+  response:
+    body:
+      background: false
+      created_at: 1781669660
+      frequency_penalty: 0.0
+      id: resp_b2761c3ff936af2e
+      incomplete_details: null
+      input_messages: null
+      instructions: null
+      kv_transfer_params: null
+      max_output_tokens: 40944
+      max_tool_calls: null
+      metadata: null
+      model: Qwen/Qwen3-30B-A3B-FP8
+      object: response
+      output:
+      - content:
+        - text: '
+
+            Okay, the user sent "Reply with exactly one word: HELLO". So they want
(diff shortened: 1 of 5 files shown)


ashwing (approved):
A few observations:

1. **Format divergence from existing cassettes**: The `events/` cassettes use a flat `sse: [lines]` format where each entry is a raw `data:` line. These new cassettes use a `turns` structure with separate `event:` and `data:` entries plus request/response metadata. Whoever writes tests against these will need a different loader — worth noting in the recording script or a README so future contributors know which format to use when.

2. **Consider adding an `expected_reasoning` field** (like the existing `expected_function_call` in `function-call-vllm-gemma4.yaml`) so cassette-driven tests can assert the accumulated reasoning text without parsing the full response body. Not blocking — can be a follow-up.

maralbahari:
> A few observations:
> 
> 1. **Format divergence from existing cassettes**: The `events/` cassettes use a flat `sse: [lines]` format where each entry is a raw `data:` line. These new cassettes use a `turns` structure with separate `event:` and `data:` entries plus request/response metadata. Whoever writes tests against these will need a different loader — worth noting in the recording script or a README so future contributors know which format to use when.
> 2. **Consider adding an `expected_reasoning` field** (like the existing `expected_function_call` in `function-call-vllm-gemma4.yaml`) so cassette-driven tests can assert the accumulated reasoning text without parsing the full response body. Not blocking — can be a follow-up.

1. we must follow the format as vLLM and OpenAI we cannot alter the request and response payload to stay as true and close possible to real cases. I will add a readme file to know how to run the cassettes recorder.  although currently there is instruction in python cassettes recorder and the bash scripts are to run different cases invoking python recorder. 
2.  we cannot add in custom fields into yaml recorders it has to be pure request/response payloads that is sent and received to vLLM server or OpenAI server.

franciscojavierarceo (approved):

PR #59, opened 2026-06-17
feat: accumulate function_call events in streaming path

ashwing:
## Summary

The streaming path's `ResponseAccumulator` handles text message events but didn't yet have handlers for `function_call` events — they fell through the wildcard arm. This adds the missing accumulation so streaming responses include `FunctionCall` output items, matching the blocking JSON path.

Prerequisite for #51 (`execute_loop` needs `FunctionCall` items available in the streaming path to dispatch tools).

### Before

A client sends a streaming request with tools:
```json
{"model": "meta-llama/...", "stream": true, "input": "What's the weather?", "tools": [{"type": "function", "name": "get_weather", ...}]}
```

vLLM responds with `function_call` SSE events, but the accumulator ignores them. The client gets back:
```json
{"status": "completed", "output": []}
```

### After

The same request now produces:
```json
{
  "status": "completed",
  "output": [{
    "type": "function_call",
    "id": "fc_1",
    "call_id": "call_abc",
    "name": "get_weather",
    "arguments": "{\"location\":\"Paris\"}",
    "status": "completed"
  }]
}
```

### What's handled

- `OutputItemAdded` with `item_type == "function_call"` — starts a new in-flight `FunctionToolCall`
- `FunctionCallArgumentsDelta` — appends to argument buffer
- `FunctionCallArgumentsDone` — finalizes with authoritative arguments, pushes to output
- Multiple function calls in one response (parallel tool use)
- Interleaved message → function_call → message ordering preserved
- Stream disconnect mid-arguments — partial args retained via forced finalize
- Orphaned deltas (no active function call) — safely cleared before next call
- Coexists with reasoning accumulation (PR #57) — unified `match` with three-way type dispatch

## Test Plan

**Unit tests** (8 in `accumulator.rs`):
- Full lifecycle, parallel tool use, interleaved with messages, orphaned deltas, forced finalize on disconnect

**Cassette integration tests** (12 in `accumulator_cassette_test.rs`):
- `tool_choice=auto` streaming + non-streaming (parallel tool calls)
- `tool_choice=required` streaming + non-streaming
- `tool_choice=named` streaming + non-streaming
- `tool_choice=none` streaming + non-streaming (zero function calls)
- Reasoning streaming — Qwen3 (reasoning + message)
- Reasoning streaming — GPT-oss (reasoning only, see note below)
- Legacy Gemma4 function_call + text-only regression guard

Non-streaming tests exercise the `from_json` path against the same cassettes.

All tests pass, `cargo fmt --check` + `cargo clippy -- -D warnings` clean.

### Note: GPT-oss streaming gap

While testing the reasoning cassettes, I found that GPT-oss doesn't emit `response.output_item.added` for the message item after reasoning — it jumps from `reasoning_text.done` → `output_text.done` → `output_item.done`. The streaming accumulator can't capture that message because no `output_item.added` creates the in-flight state. Qwen3 emits the event correctly. Not blocking — the full output is in the `response.completed` payload. Filed as #62 for follow-up.

### DIFF
--- a/crates/agentic-core/src/executor/accumulator.rs
+++ b/crates/agentic-core/src/executor/accumulator.rs
@@ -235,6 +235,7 @@ impl ResponseAccumulator {
     /// This is the core state machine — callers that already have a normalized
     /// frame (e.g. [`StreamTee`](future)) can call this directly without
     /// re-parsing from a raw line.
+    #[allow(clippy::too_many_lines)]
     pub(crate) fn process_event(&mut self, frame: &EventFrame) {
         match (&frame.event_type, &frame.payload) {
             (SSEEventType::ResponseCreated, EventPayload::Response { id, .. }) if !id.is_empty() => {
@@ -283,8 +284,7 @@ impl ResponseAccumulator {
                         self.finalize_current_reasoning();
                         self.finalize_current_function_call();
                         self.finalize_current_message();
-                        self.current_message =
-                            Some(OutputMessage::new(item_id, MessageStatus::InProgress.as_str()));
+                        self.current_message = Some(OutputMessage::new(item_id, MessageStatus::InProgress.as_str()));
                     }
                 }
             }
--- a/crates/agentic-core/src/executor/accumulator.rs
+++ b/crates/agentic-core/src/executor/accumulator.rs
@@ -16,7 +16,8 @@ use crate::events::{EventFrame, EventPayload, SSEEventType, normalize_sse_line};
 use crate::executor::error::{ExecutorError, ExecutorResult};
 use crate::types::event::{MessageStatus, ResponseStatus};
 use crate::types::io::{
-    OutputItem, OutputMessage, OutputTextContent, ReasoningOutput, ReasoningTextContent, ResponseUsage,
+    FunctionToolCall, OutputItem, OutputMessage, OutputTextContent, ReasoningOutput, ReasoningTextContent,
+    ResponseUsage,
 };
 use crate::types::request_response::{IncompleteDetails, ResponsePayload};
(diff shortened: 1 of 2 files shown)


ashwing:
@maralbahari This is the function_call accumulation PR we discussed in #51. Ready for review when you get a chance.

Covers the full streaming lifecycle: `OutputItemAdded(function_call)` → argument deltas → `FunctionCallArgumentsDone` → finalized `OutputItem::FunctionCall` in the response payload. Also handles parallel tool calls, interleaving with text messages, and stream disconnect with partial args.

Once this lands, #51's `dispatch_tools` will have real `FunctionCall` items to work with in the streaming path.

maralbahari:
@ashwing thank you for the quick PR.
I have added https://github.com/vllm-project/agentic-api/pull/60 to capture all the scenarios of request payload for variety of function_call options as tool_choices to vLLM upstream and recorded them. so we can cover comprehensive tests. could you please try them out.

franciscojavierarceo (approved):

ashwing:
@maralbahari thanks! I rebased on top of #60 and added cassette tests that exercise all 4 `tool_choice` modes through the accumulator:

- `auto` streaming + non-streaming (parallel tool calls — `get_stock_price` + `search_web`)
- `required` streaming + non-streaming
- `named` streaming + non-streaming
- `none` streaming + non-streaming (asserts zero function calls)

Also added reasoning cassette tests (Qwen3 + GPT-oss) as a regression guard — found an interesting edge case with GPT-oss where it skips `output_item.added` for the message after reasoning. Filed #62 for follow-up.

12 cassette tests total, all passing.

PR #64, opened 2026-06-18
refactor: split io types, add SSEItemType enum, reduce allocations

maralbahari:
## Summary

- `types/io.rs` split into four focused modules: `input`, `output`, `usage`, `tools`. All re-exported from `io/mod.rs` so no import paths changed.
- `TryFrom<EventPayload>` added for `FunctionToolCall`, `OutputMessage`, and `ReasoningOutput`, with uuid generation moved into each impl. `process_event` now constructs in-flight items via `try_from` instead of inline struct literals.
- `ApplyDone` trait added on `ReasoningOutput` and `FunctionToolCall`. Collapses the `ReasoningTextDone` and `FunctionCallArgumentsDone` match arms from ~10 lines each to 3.
- `process_event` takes owned `EventFrame`. `OutputItemAdded` arm dispatches on the new `SSEItemType` enum (`Reasoning`, `FunctionCall`, `Message`) instead of hardcoded strings, then passes the owned payload to `try_from` with no further allocation.
- `SSEItemType` added to `events/types.rs` with `From<&str>`, `From<String>`, and `PartialEq<str>` impls. `EventPayload::OutputItemAdded` and `OutputItemDone` now carry `item_type: SSEItemType` instead of `String`. The normalize layer converts the wire string via `SSEItemType::from(json_str(...))`.
- `ResponseCompleted` usage deserialization replaced with `deserialize_from_value_opt` from `utils::common`, removing the last direct `serde_json::from_value` call in the accumulator.
- `from_json` uses `json[...].take()` + `deserialize_from_value_opt` to move values out of the parsed JSON instead of cloning them.
- Finalize helpers `finalize_all`, `finalize_except_*` extracted to remove repetition across match arms.
- Status strings replaced with `MessageStatus::Completed/InProgress.as_str().into()` throughout.

## Benchmarks

Measured with `run_accumulator_bench.sh` (100 samples, `BENCH_CONCURRENCY=2,4,8,16,32`) against a cassette containing reasoning + function_call + message output items.

### Concurrent stream (tokio tasks, 582 SSE lines × N)

| Threads | main thrpt | RFC thrpt | delta |
|---|---|---|---|
| 2 | 734 K/s | 779 K/s | +6% |
| 4 | 858 K/s | 929 K/s | +8% |
| 8 | 723 K/s | 749 K/s | +4% |
| 16 | 709 K/s | 727 K/s | +3% |
| 32 | 783 K/s | 749 K/s | -4% |

### Concurrent JSON (threads, 1 body × N)

| Threads | main thrpt | RFC thrpt | delta |
|---|---|---|---|
| 2 | 4.40 K/s | 4.42 K/s | ~0% |
| 4 | 5.83 K/s | 6.26 K/s | +7% |
| 8 | 7.88 K/s | 7.95 K/s | +1% |
| 16 | 9.10 K/s | 9.12 K/s | ~0% |
| 32 | 8.17 K/s | 7.95 K/s | -3% |

## Test Plan

- All 186 unit tests pass with `--test-threads=8`
- `cargo clippy --all-targets` clean with no warnings


crates/agentic-core/src/executor/accumulator.rs
```diff
@@ -235,120 +220,61 @@ impl ResponseAccumulator {
...
-                    item_id.clone()
-                };
-                match item_type.as_str() {
-                    "reasoning" => {
-                        self.finalize_current_function_call();
-                        self.finalize_current_message();
-                        self.finalize_current_reasoning();
-                        self.current_reasoning = Some(ReasoningOutput::new(item_id));
+            (SSEEventType::OutputItemAdded, payload @ EventPayload::OutputItemAdded { item_type, .. }) => {
+                self.finalize_all();
```
franciscojavierarceo: i think there's a bug here for parallel tool-call streams.

the accumulator now has one `in_flight` item and each `OutputItemAdded` finalizes/replaces it. That's a problem when the stream emits multiple `output_item.added` events before the deltas finish arriving. we have this shape in `event_normalizer_test::test_parallel_function_calls`.

i think we need to keep in-flight state keyed by `item_id` or `output_index`, then apply deltas/done to the matching item and preserve ordering separately.

franciscojavierarceo (changes requested):
holding here, i think there's a regression introduced in the `finalize_all()` in `accumulator.rs`

let's resolve that and then i think we're good to go.

crates/agentic-core/src/executor/accumulator.rs
```diff
@@ -235,120 +220,61 @@ impl ResponseAccumulator {
...
+                self.in_flight = match item_type {
```
ashwing: when this moves to keyed in-flight, `output_index` is available on the event payloads but not persisted on InFlight. if you key the map by `item_id`, you'll need `output_index` stored on the struct for ordered drain at finalize. if you key by `output_index` directly, ordering comes free.

crates/agentic-core/src/executor/accumulator.rs
```diff
@@ -235,120 +220,61 @@ impl ResponseAccumulator {
...
+                    SSEItemType::Reasoning => ReasoningOutput::try_from(payload).ok().map(|item| InFlight::Reasoning {
```
ashwing, line 232: if validation guards land in these TryFrom impls later, failure mode becomes "malformed event silently produces no output item" with no log. might want a `debug_assert!` or tracing span here so that's visible in dev.

crates/agentic-core/src/types/io/output.rs
```diff
@@ -163,96 +155,85 @@ impl ReasoningOutput {
...
-    #[serde(rename = "function")]
-    Function {
-        name: String,
-    },
+/// Applies a `*Done` event payload onto an in-flight output item.
+///
+/// `buffer` holds accumulated delta text/arguments. If the payload's own field
+/// is empty the buffer is used as the final value and then cleared; otherwise
+/// the buffer is discarded and the payload value is used directly.
+pub trait ApplyDone {
```
ashwing, line 193: for StreamTee, `impl ApplyDone for OutputMessage` handling `OutputTextDone` would slot in here without changing the accumulator's state machine. the accumulator deliberately skips `OutputTextDone` since `finalize_all()` flushes at item boundary — downstream consumers will want that signal.

ashwing (commented):
three notes inline.

crates/agentic-core/src/executor/accumulator.rs
maralbahari: @ashwing got addressed by using indexmap which preserves the order of the items and benefit of hashmap.

crates/agentic-core/src/executor/accumulator.rs
maralbahari: @ashwing By the time execution reaches TryFrom, a malformed event is already impossible. There are two upstream validation layers:

1. `normalize_sse_line` parses and validates the raw SSE JSON; any malformed line returns None and never reaches process_event.
2. The match arm payload @ EventPayload::OutputItemAdded { .. } guarantees the correct variant at the type level before `try_from` is called.

Even after these two checks, TryFrom is still there to catch any parse error if the event is malformed. therefore there is no silent ignorance of malformed events.

crates/agentic-core/src/executor/accumulator.rs
ashwing: Indexmap handles it — insertion order preserved with O(1) lookup. Works.

crates/agentic-core/src/executor/accumulator.rs
ashwing: Makes sense — the two upstream layers make the TryFrom path effectively unreachable. My concern was forward-looking if someone adds a new code path that bypasses normalize, but agreed that's not worth defensive code today.

crates/agentic-core/src/executor/accumulator.rs
maralbahari: @franciscojavierarceo storing the inflight items into an indexmap to ensure the order of receiving items and benefiting from map structure for item_id referencing and retrieval.

franciscojavierarceo (approved):

PR #67, opened 2026-06-19
docs: tool framework design — multi-type tool support

ashwing:
## Summary

Design proposal for a generic tool framework that handles heterogeneous tool types (`function`, `mcp`, `web_search`, `file_search`, `code_interpreter`) through a single pipeline with type-specific handlers.

Key architectural decisions:
- `ResponsesTool` becomes a `#[serde(tag = "type")]` enum (backward-compatible)
- Request-scoped `ToolRegistry` routes post-inference `function_call` items by name lookup
- `ToolHandler` trait — new types implement it without touching the executor loop
- `function` type is client-owned (`requires_action`); all other types are gateway-executed
- `LoopDecision::ContinuePartial` handles mixed requests (gateway + client tools in one call)

References ADR-01 D7 (MCP as primary tool interface) and ADR-03 D3 (tool registry in agentic-core).

Looking for feedback on:
1. The `function` type as client-owned passthrough vs gateway-executed — is this the right split?
2. `ContinuePartial` semantics for mixed tool requests
3. The `ToolHandler` trait surface area — too much? too little?
4. PR decomposition and ordering

## Test Plan

- Doc-only change, no code
- `cargo build --workspace && cargo test --workspace` passes (no functional changes)


ashwing:
@franciscojavierarceo @noobhappylife @maralbahari — would appreciate your eyes on this design proposal. 

Key questions I'd love feedback on:
1. Is `function` as client-owned passthrough (`requires_action`) the right split? (see Alternatives section)
2. Does the `ToolHandler` trait surface area feel right for extensibility?
3. Does the `ContinuePartial` loop decision make sense for mixed tool requests?

This is meant to be the framework that MCP, web_search, file_search all plug into.

maralbahari:
@ashwing thank you for the design document. 
The overall flow sounds correct. user controls the tool choices then `agentic-api` need to normalize the tools into `function_call`s the way that the model sees them and then inference results in which `function_call` to be execution either on `agentic-api` or the client  side. 
to determine how to execute the `function_call` s would depend on how we design.  let's say execute them in parallel in case of multiple tools being executed in parallel would need to consider a timeout if any of the job was not completed in given time to drop them.

there is a case we need two `function_call`  executions in parallel let's say one execution would need to take place in  agentic-api (gateway) that is bound by timeout and one needs to be executed on client side (referred in this design as `function`)  which requires action by user then we have several options here either user declines and all tools stops or user responses with another prompt and doesnt complete the execution of the current `function` would this case be handled in `continuePartial`? or the `continuePartial` is only when user confirms the `function`.

there is some complication there which I feel we could implement in small stages without partial complication and leave that last stage. Note that the tool calls are not actually part of MVP. with that being said we do not need to rush to get into this stage we need to plan the stages to reach to full agentic-loop. 

I think one starting point to get  the loop sequence right we could observe and record a multi-turn conversation cassettes from OpenAI with some tool choices to clarify the map of `execution_loop` with simple cases. then build from there.  
The steps I could think of that would benefit us to implement first  to enable us get closer to a wholesome agentic-loop iteration with tool options :
- record multi-turn conversation with `function_calls` from OpenAI (cassettes).
- filtering tool options and tool normalizer
- implement a simple `execution_loop` with simple `LoopDecision` based on the recorded cassettes.
- mcp executor
- support of other tools like file_search (currently in progress using OGX)
- refine `execution_loop` and `LoopDecision` with new tools

ashwing:
@maralbahari On `ContinuePartial` — fair point about the decline/redirect edge case. I'll defer that to a follow-up and start with the clean two-way split: gateway-owned tools loop, client-owned tools halt with `requires_action`. Mixed case is additive once both paths work.

Cassettes with tool calls are already part of the plan — we'll record multi-turn OpenAI sessions with function + MCP tools as the first PR, same approach as #66.

On ordering — I'd push back on full sequential. PR A (types + registry + normalize) is pure interfaces with no execution logic. Once that lands, PR B (dispatch) and PR C (MCP handler) can develop in parallel — B tests against mock handlers, C tests against a mock MCP server, neither imports the other. They only meet at the integration layer after both land. Serializing them adds wait time for no technical reason — decoupling "how to route" from "how to execute" is the whole point of the trait-based design.

Proposed: cassettes → types/registry/normalize → dispatch + MCP in parallel → integration.

ashwing:
@maralbahari @franciscojavierarceo Can we merge this if there are no further comments?

maralbahari:
> @maralbahari @franciscojavierarceo Can we merge this if there are no further comments?

looks good to me.
@franciscojavierarceo have any comments on this?

ashwing:
Refreshed this to match what actually shipped — it started as a proposal and the framework's since landed across #80/#82/#84/#85/#91 (with #83/#89 in review), so keeping it as a forward-looking sketch didn't make sense.

What changed vs the original:
- `ToolHandler` split into `ToolHandler` + `GatewayExecutor` (execute() only belongs on gateway-owned types), `Pin<Box<dyn Future>>` over `#[async_trait]`
- `ToolType` gains `CodexNamespace`; `ResponsesTool` shown as the shipped `#[non_exhaustive]` enum with the `namespace` + `Unknown` variants
- dispatch is now two layers — per-call `registry.dispatch()` (#82) under the multi-turn `classify_round`/`LoopDecision` loop (#83) — with a diagram
- `LoopDecision` trimmed to 4 variants; mixed gateway+client turns resolved by precedence, not the old `ContinuePartial`

Added an Implementation Status table (PR mapping) and a Future Work section (layering ADR, GatewayAccumulator, per-type config, the two unshipped handlers). Rationale sections are untouched. Each divergence from the original is annotated inline so the history's clear.

docs/design/tool-framework.md
```diff
@@ -0,0 +1,461 @@
...
+---
+
+## Pipeline Stages
+
+Every request with tools passes through 7 stages. Stages 1–4 run once at request start. Stages 5–7 repeat per inference iteration.
+
+| # | Stage | Generic (framework) | Type-Specific (handler) |
+|---|-------|---------------------|-------------------------|
+| 1 | **Parse** | Deserialize `tools[]`, classify by `type` | Validate required fields per type |
+| 2 | **Discover** | Iterate handlers, collect discovered tools | MCP: `tools/list`. Others: no-op |
```
franciscojavierarceo, line 86: I'm fine with this for now but it's worth noting that this can be slow proportional to the discovery. 

Probably we can consider doing discovery at either an explicit registry step or server start.

franciscojavierarceo (approved):
one small point to consider but otherwise this lgtm

Issue #54, opened 2026-06-15
RFC: Integration with Codex CLI


noobHappylife:
### Problem statement / motivation

This RFC discussed and present findings about Codex CLI. This aims to let `agentic-api` be able to bridge vLLM and Codex CLI. 

A) The built-in tool types that Codex CLI uses, for `agentic-api` to work with Codex CLI. Will need to properly handle these tool types. 
B) The Codex CLI auto-approval feature required an auto-approval model. So, for `agentic-api` to be compatible with this feature, will need to handle the auto-approval model properly. 

### A) Codex CLI built-in tool types

Codex CLI will call these tool types:
| Type | Execution |
|------------------|-----------|
| `type: "function"` | Codex CLI executes |
| `type: "namespace"` | Codex CLI executes |
| `type: "tool_search"` | Codex CLI executes |
| `type: "custom"` | Codex CLI executes |
| `type: "code_interpreter"` | `agentic-api` executes |
| `type: "web_search"` | `agentic-api` executes |

- type: function is essentially the common function tool call
- type code_interpreter and web_search are both OpenAI built-in tools that are executed server-side. 
- while type "namespace", "tool_search", "custom" are function type that Codex CLI might send in a request. And the model is supposed to return the tool call signature and let Codex CLI execute and return the tool call result

Solutions: 
- Likely `agentic-api` needs to recognize these tool type and internally mapped to a function tool call to pass to upstream vLLM. 

---

Here are some examples to what the namespace", "tool_search", "custom" are. 

-  `namespace` — Tool Container

```json
{
    "type": "namespace",
    "name": "mcp__github",
    "description": "GitHub tools",
    "tools": [
        {"name": "create_issue", "description": "Create issue", "parameters": {...}},
        {"name": "search_code", "description": "Search code", "parameters": {...}}
    ]
}
```

- `tool_search` — Deferred Tool Discovery

```json
{
    "type": "tool_search",
    "description": "Search for deferred tools",
    "parameters": {
        "type": "object",
        "properties": {
            "query": {"type": "string", "description": "Search query"},
            "limit": {"type": "number", "description": "Max results"}
        },
        "required": ["query"]
    }
}
```

- `custom` — Custom/Freeform Tool

```json
{
    "type": "custom",
    "name": "lark_parser",
    "description": "Parse Lark grammar files",
    "format": {
        "type": "grammar",
        "syntax": "lark",
        "definition": "..."
    }
}
```

---

[...]

PR #68, opened 2026-06-19
docs: Codex integration design

haoshan98:


ashwing (commented):
The Codex compatibility concepts here — tool registry, client vs gateway ownership, normalize for upstream, loop decision with client action — overlap significantly with #67 which proposes a generic tool framework covering all tool types. The registry, normalize pipeline, and `RequiresAction` loop variant are the same infrastructure. Worth coordinating so we don't end up with two parallel type systems solving the same routing problem.

Separately, this reads better as a standalone `docs/design/codex-integration.md`. The `core-public-api.md` doc is specifically about the executor loop phases and its ownership is already defined in the header.

docs/design/core-public-api.md
```diff
@@ -252,6 +254,249 @@ How the complete pipeline maps to @leseb's proposed filter chain:
...
+The current `ResponsesTool = FunctionTool` alias is too narrow for Codex. Replace it with a tagged tool enum that
+preserves unknown shapes while giving Codex-used variants first-class names.
+
+Do not implement this as `#[serde(tag = "type")]` wrapping the existing `FunctionTool` struct directly, because
+`FunctionTool` already stores the wire `type` field. Either use variant-specific payload structs that omit the
+already-consumed tag, as sketched below, or write manual deserialization that preserves the raw object.
+
+```rust
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(untagged)]
```
ashwing: `untagged` with a `Known`/`Unknown` wrapper is fragile — serde tries variants in declaration order and takes the first successful parse. Debugging deserialization mismatches becomes painful. `#[serde(tag = "type")]` with a catch-all variant is more explicit and gives clearer errors when a new type appears. See #67 for that approach.

docs/design/core-public-api.md
```diff
@@ -252,6 +254,249 @@ How the complete pipeline maps to @leseb's proposed filter chain:
...
+pub struct ResponsesFunctionTool {
+    pub name: String,
+    pub description: Option<String>,
+    pub parameters: Option<Value>,
+    pub strict: Option<bool>,
+    #[serde(default)]
+    pub defer_loading: bool,
+}
+
+pub struct CodexNamespaceTool {
```
ashwing: `namespace`, `tool_search`, and `custom` are Codex CLI internals — they are not part of the OpenAI Responses API spec. Baking them as first-class types in `agentic-core` couples us to one client's implementation details. Alternative: accept via a catch-all variant, normalize to function tools for inference, and preserve the raw shape in the registry for passthrough on response. That way we don't break if Codex changes their internal wire format.

docs/design/core-public-api.md
```diff
@@ -252,6 +254,249 @@ How the complete pipeline maps to @leseb's proposed filter chain:
...
+The key distinction is execution owner, not just wire type. `function` is a shared wire type: Codex can own local
+functions, while future gateway integrations may also expose server-executed functions. The request normalizer must
+classify every model-visible tool before inference and carry that registry through response handling.
+
+### Public Type Additions
+
+The current `ResponsesTool = FunctionTool` alias is too narrow for Codex. Replace it with a tagged tool enum that
+preserves unknown shapes while giving Codex-used variants first-class names.
+
+Do not implement this as `#[serde(tag = "type")]` wrapping the existing `FunctionTool` struct directly, because
```
ashwing: The issue described here — "`FunctionTool` already stores the wire `type` field" — is a fixable implementation detail, not a reason to avoid tagged enums. The fix is to remove the redundant `type_: String` from the payload struct (serde consumes the tag during deserialization, so it shouldn't also be a field). #67 takes this approach and it works cleanly with `#[serde(tag = "type")]`.

docs/design/core-public-api.md
```diff
@@ -252,6 +254,249 @@ How the complete pipeline maps to @leseb's proposed filter chain:
...
+still emits `function_call` with an optional `namespace`. Any backend-specific flattening is an adapter detail,
+not the public semantics of the tool.
+
+```rust
+pub struct NormalizedTools {
+    pub upstream_tools: Vec<Value>,
+    pub registry: ToolRegistry,
+}
+
+pub struct ToolName {
```
ashwing: Namespace splitting adds complexity the gateway likely doesn't need. vLLM emits flat function names — the model doesn't know about namespaces. If Codex organizes tools into namespaces client-side, the gateway can preserve the original declaration verbatim in the registry and return it on `requires_action` without parsing or reconstructing namespace structure. Simpler and less coupled to Codex's naming convention.

docs/design/core-public-api.md
```diff
@@ -252,6 +254,249 @@ How the complete pipeline maps to @leseb's proposed filter chain:
...
+the call to Codex, and expects Codex to continue the conversation with the corresponding tool output item after
+local execution.
+
+The loop needs an explicit client-action decision so this path is not confused with either `Done` or
+`Continue`:
+
+```rust
+pub enum LoopDecision {
+    Continue(Vec<InputItem>),
+    RequiresClientAction(Vec<OutputItem>),
```
ashwing: `Vec<OutputItem>` is too broad here. The only items that require client action are function calls — not messages, not reasoning. Using `Vec<FunctionToolCall>` makes the type system enforce this constraint rather than relying on callers to only put the right variants in.

docs/design/core-public-api.md
```diff
@@ -252,6 +254,249 @@ How the complete pipeline maps to @leseb's proposed filter chain:
...
+    Custom(CodexCustomTool),
+}
+
+pub struct ResponsesFunctionTool {
+    pub name: String,
+    pub description: Option<String>,
+    pub parameters: Option<Value>,
+    pub strict: Option<bool>,
+    #[serde(default)]
+    pub defer_loading: bool,
```
ashwing: `defer_loading` appears as a named field but there's no description of what the gateway does with it. If it's a client hint the gateway ignores, it should live in the catch-all raw `Value` rather than as a typed field — otherwise it implies the gateway acts on it.

docs/design/core-public-api.md
```diff
@@ -252,6 +254,249 @@ How the complete pipeline maps to @leseb's proposed filter chain:
...
+}
+```
+
+### Auto-Approval Model Alias
+
+[Issue #54](https://github.com/vllm-project/agentic-api/issues/54) also notes Codex's auto-approval request path. MVP support should add a simple model alias map in
+configuration:
+
+```toml
+[model_aliases]
```
ashwing: Model aliases are practical, but the issue #54 context suggests this also changes approval behavior (auto-review = skip human confirmation). Conflating model routing with approval semantics in a single alias is surprising — a request to `codex-auto-review` silently changes safety behavior. If both are needed, they should be separate config knobs.

maralbahari:
> The Codex compatibility concepts here — tool registry, client vs gateway ownership, normalize for upstream, loop decision with client action — overlap significantly with #67 which proposes a generic tool framework covering all tool types. The registry, normalize pipeline, and `RequiresAction` loop variant are the same infrastructure. Worth coordinating so we don't end up with two parallel type systems solving the same routing problem.
> 
> Separately, this reads better as a standalone `docs/design/codex-integration.md`. The `core-public-api.md` doc is specifically about the executor loop phases and its ownership is already defined in the header.

@haoshan98 it would be good to have separate `docs/design/codex-integration.md` for codex integration.

@ashwing note that the codex integration is from tool angle using codex for tool calls. regarding the overlap what is shared between any tools are execution in the `execution_loop`.  however at this stage for tool implementation we can have a generic interface `Tool` where each would expose common trait like `normalize` and `execute` this way each contributors/developer they can implement their tools without interfering and overlapping then after all components are in we add them into `execution_loop`.
for instance currently we are handling the stream events by categorizing the items into `reasoning`, `function_call` and `message` since they're being added into a buffer we can use the interface I just mentioned above to identify the tool type then call `tool.execute()`.
and we would need tool.normalize() after receiving the `RequestPayload` to get user selected tools filter the tool options normalize before `call_inference()`. 
while the `execution_loop` and `LoopDecision` involves orchestrating the all tools on flight and live streaming. while currently we are buffering the inflight item responses for us to add support each tool types at a time and verify their execution and correctness in `execution()` (in `agentic-core/executor/engine.rs`) and `ResponsesAcumulator.process_event()`  one all that done then we implement the whole live orchestration with `execution_loop` and `LoopDecision`.
codex as tool call would be mostly use to bridge the gap of `agentic-api` tool execution and we do not own it is considered as custom tool. since we do not have any support of any function_call execution for fastest prototype we use use codex to demo `agentic-api` with tool call until we start implementing `agentic-api` built-in tools for `file_search`, `code_interpreter` etc.

@haoshan98 @ashwing let me know your thoughts and feedback on this approach. then @haoshan98 in a separate `docs/design/codex-integration.md` would write a more specific implementation that would be implement by her for CustomCodexTool and what components would be shared with design in https://github.com/vllm-project/agentic-api/pull/67

ashwing:
@maralbahari @haoshan98 Agree with this layered approach — it maps cleanly to what #67 proposes. A few thoughts from the tool framework design side:

**On the `Tool` trait with `normalize()` and `execute()`:** This is exactly the `ToolHandler` trait in #67. The interface is `validate → discover → normalize → execute → emit`. Each tool type (Codex custom tool, MCP, web_search, etc.) implements that trait independently without touching the executor. Haoshan's `CustomCodexTool` would be one `ToolHandler` impl — discovery returns Codex's tool manifest, normalize flattens to function defs for vLLM, execute dispatches to Codex's sandbox.

**On phased implementation:** Agree we don't need `ContinuePartial` or live orchestration in phase 1. The order I'd propose (aligns with what you described):
1. Tool types + registry (PR A from #67's shipping plan) — `ResponsesTool` enum, `ToolRegistry`, `ToolHandler` trait skeleton, `FunctionHandler` as the trivial passthrough
2. Simple `execution()` dispatch — registry lookup determines gateway vs client routing, returns `requires_action` for function tools
3. Codex handler (haoshan) + MCP handler — first real `execute()` impls against the trait
4. Full `execution_loop` + `LoopDecision` with streaming — the live orchestration layer

**On Codex as "custom tool" / bridge:** Makes sense as a stopgap. From #67's perspective, Codex is just another `ToolHandler` that happens to route `execute()` to an external agent runtime. It slots in at step 3 above without any special-casing in the executor.

**On the shared interface:** The pieces both PRs need are identical — `ToolRegistry` for routing, `normalize()` for the vLLM boundary, `execute()` for dispatch. Happy to have #67 land the trait + registry skeleton, then haoshan builds `CustomCodexTool` against it. Avoids two parallel type systems.

The multi-turn tool call cassettes are already recorded (PR #77, now merged) — 12 cassettes covering linear, streaming, branching, parallel, and tool-output-only patterns against both vLLM and OpenAI. Those can feed directly into the simple `execution_loop` implementation at step 2.

haoshan98:
Thanks for the thoughtful feedbacks @ashwing @maralbahari and for clarifying the overlap with #67.

I agree that the generic tool framework should own the formal normalization, registry, ownership classification, execution, and loop-decision abstractions. I do not want this PR to create a parallel tool framework.

I updated the Codex design to frame this PR as a small MVP compatibility slice: accept and preserve Codex-used Responses tool shapes, preserve namespaced/function/custom/tool-search call items through request/response/storage paths, support `previous_response_id` continuation, and add model alias routing for vLLM-backed models.

The helper types in this PR are intentionally lightweight. They document the Codex requirements and let us verify the wire-shape/continuation behavior now, but #67 should formalize the canonical `ToolHandler`, registry, normalization lifecycle, ownership model, and requires-action loop decision. Once #67 lands, this slice can plug into or be refactored onto those shared abstractions.

docs/design/codex-integration.md
```diff
@@ -0,0 +1,153 @@
...
+- **PR #67:** formalize generic tool normalization, execution, registry, ownership, and loop decisions.
+
+---
+
+## Current PR Scope
+
+This PR should do only the minimum needed for Codex compatibility:
+
+- Add this standalone design doc.
+- Accept Codex-used tool declarations without rejecting requests.
```
franciscojavierarceo, line 28: probably log the tool call so we can create an issue for it

franciscojavierarceo (approved):

docs/design/codex-integration.md
maralbahari: @franciscojavierarceo by `log` do you mean like record a cassettes with some request, response payloads using codex cli?

franciscojavierarceo: oh no i meant we can emit a log on the server so the inference request can re read and reused for debugging.

PR #75, opened 2026-06-23
refactor: split `executor/engine.rs` and eliminate `ExecutionContext` clone per request

maralbahari:
## Summary

`ExecutionContext` owns long-lived shared state: DB connection pool, HTTP client, and server config. Cloning it per request means each clone holds independent references to those resources, making it easy for future fields to silently diverge between the shared instance and the per-request copy. This refactor removes `client_auth` (the only per-request field) from `ExecutionContext`, making the struct purely shared and immutable for the server lifetime.

- `client_auth: Option<String>` removed from `ExecutionContext`. Supplied per-request via `ExecuteRequest::with_auth()` instead, threaded as a parameter through `run_blocking` and `run_stream`.
- `ExecuteRequest` builder added: `ExecuteRequest::new(payload, exec_ctx).with_auth(token).run().await`. The boundary between shared config and per-request state is now explicit in the type.
- `resolve_exec_ctx` in `agebtic-server/handlers/common.rs` deleted. Replaced by `extract_bearer(headers, config_key)` and a direct `ExecuteRequest` construction, zero clones per request. `extract_bearer` returns the request-level bearer token if present, and falls back to `openai_api_key` from `AppState
- `execute()` kept as a one-liner shim over `ExecuteRequest` so existing call sites compile without changes.
- `engine.rs` split into four focused modules: `inference.rs` (HTTP transport), `rehydrate.rs` (history rehydration), `persist.rs` (response persistence), `engine.rs` (orchestration only). All public names re-exported from `executor/mod.rs`, no import paths changed.

## Test Plan

- `cargo test -- --test-threads=$(nproc)`: 201 tests pass, 0 failed


crates/agentic-server/src/handler.rs
```diff
@@ -94,22 +94,14 @@ async fn proxy_responses(state: &AppState, parts: Parts, body: Bytes) -> Respons
     convert_response(proxy_request(proxy_req, &state.proxy_state).await)
 }
 
-fn resolve_exec_ctx(state: &AppState, parts: &Parts) -> Arc<ExecutionContext> {
-    let request_auth = parts
+fn extract_bearer(parts: &Parts) -> Option<String> {
```
ashwing: `extract_bearer` returns `None` when there's no `Authorization` header, and that `None` flows all the way through to `call_inference` — so requests with no header hit the upstream with no bearer token. The old `resolve_exec_ctx` fell back to `exec_ctx.client_auth` in that case. Also the `execute()` shim goes through `ExecuteRequest::new(...).run()` which defaults `client_auth: None`, same result.

The deeper issue is that `Config.openai_api_key` is now completely dead — `from_config` no longer reads it into `ExecutionContext` — so even if you tried to fall back there's nothing to fall back to. Deployments that set `OPENAI_API_KEY` server-side without injecting `Authorization` on every client request will start 401ing.

crates/agentic-core/src/executor/request.rs
```diff
@@ -108,7 +106,6 @@ impl ExecutionContext {
             resp_handler,
```
ashwing, line 106: `Config.openai_api_key` is still accepted here and set via CLI in `main.rs`, but it's never actually used — the field was dropped from `ExecutionContext`. Either wire it back in as a server-level fallback or remove it from `Config` entirely so the dead code is visible.

ashwing (commented):
`run_blocking` and `run_stream` are private which is fine — `rehydrate_conversation`, `call_inference`, and `persist_response` are `pub` so Praxis filters can still compose steps individually. One thing worth thinking about before tool dispatch lands: `ExecuteRequest::run()` is currently a straight pipeline. Tool calls need to re-enter inference after execution, so `run()` will eventually need to loop or accept some kind of decision callback. Might be worth leaving that door open in the shape of `ExecuteRequest` now rather than having to reshape a stable API later.

crates/agentic-server/tests/common/mod.rs
```diff
@@ -28,7 +28,6 @@ pub fn test_state(config: &Config) -> AppState {
         ResponseHandler::new(ResponseStore::disabled()),
         Arc::new(reqwest::Client::new()),
         config.llm_api_base.clone(),
-        config.openai_api_key.clone(),
     ));
```
ashwing, line 32: `openai_api_key: Some("test-key")` is set here but never reaches `ExecutionContext` anymore (see `from_config`). Once the auth fallback is fixed, worth adding a test that a no-`Authorization`-header request still forwards the server-configured key upstream — otherwise this silently regresses again.

crates/agentic-server/src/handler.rs
maralbahari: @ashwing when app state is created now the `openai_api_key` from config is also stored in app state similar to `llm_api_base`. then per request if the token is not found via `extract_bearer` during authorization. we fall back to `config.openai_api_key`  since the config is created first then app state then per request api would be received. note that it is ok and intended that user would not pass api key and we don't necessarily need it as this is the same behavior as vLLM upstream (vLLM can run without auth, so None being passed upstream when neither the client nor config provides a key is the correct behavior ). 

another note the config in `agentic-core` struct that is shared among different modules it is not mean to be only for execution context etc. the `openai_api_key`  is not a dead code  as it is used from config in other modules like `proxy.rs` and `readiness.rs` . `ExecutionContext.from_config` to build execution context does not mean it must use all the config filed.

crates/agentic-core/src/executor/request.rs
maralbahari: @ashwing `from_config` to build execution context does not mean it must use all the config filed. note that core config is shared between other modules. we cannot drop `openai_api_key` as it is used from config in other modules like `proxy.rs` and `readiness.rs`.

maralbahari:
> `run_blocking` and `run_stream` are private which is fine — `rehydrate_conversation`, `call_inference`, and `persist_response` are `pub` so Praxis filters can still compose steps individually. One thing worth thinking about before tool dispatch lands: `ExecuteRequest::run()` is currently a straight pipeline. Tool calls need to re-enter inference after execution, so `run()` will eventually need to loop or accept some kind of decision callback. Might be worth leaving that door open in the shape of `ExecuteRequest` now rather than having to reshape a stable API later.

@ashwing  Thank you for the feedback. I think  we should modify APIs as needed when the feature actually lands. Right now we have no tool calls, so any shape we add to `ExecuteRequest` would be based on guesses about how tool dispatch will work in practice. Those guesses will almost certainly be wrong once we see the real requirements.
Adding a callback or loop shape today means future implementors have to work around our speculation rather than design from real constraints. The "open door" would end up being dead weight or worse, the wrong door entirely.
When tool execution is actually implemented, reshaping `run()` is cheap. `ExecuteRequest` is not a public stable API with many external callers. We can make any structural changes the real implementation needs at that point.

crates/agentic-server/src/handler.rs
ashwing: Missed that — `extract_bearer` on the branch takes `config_key` as a fallback parameter and `AppState` carries `openai_api_key`. The fallback is already there. Disregard.

crates/agentic-core/src/executor/request.rs
ashwing: Fair point — `Config` is shared and the field is used in `proxy.rs` and `readiness.rs`. Disregard.

ashwing:
Makes sense — no point adding shape we'll likely have to undo. Will revisit when tool dispatch is real.

maralbahari:
@franciscojavierarceo @leseb ready for review.

leseb (approved):

PR #76, opened 2026-06-23
feat: add Responses WebSocket support

### LINKED ISSUE #72: Add Responses WebSocket support for Codex continuations
Parent tracker: #69
Design context: #65

## Summary

Add a Responses WebSocket adapter in `agentic-api` so Codex-style continuations can use `previous_response_id` plus marginal input and reach the same replay path as HTTP continuations.

## Scope

- Accept Codex-style `response.create` frames.
- Map `previous_response_id` into the response-store continuation path.
- Normalize HTTP and WebSocket requests into the same internal Responses request model.
- Emit Responses stream events back over the socket.
- Preserve Codex fallback semantics when a request is not a valid continuation.

## Acceptance criteria

- First-turn and continuation WebSocket flows are covered by tests.
- WebSocket continuation reaches the same replay planner as HTTP continuation.
- Invalid continuation requests fall back or fail in a documented way.
- vLLM itself does not need to expose a Responses WebSocket endpoint for the first implementation.

franciscojavierarceo:
## Summary
- Fixes #72 by adding `GET /v1/responses` WebSocket handling for `response.create` frames and local `previous_response_id` continuations.
- Streams upstream Responses events over the socket while preserving gateway response IDs and storing completed turns for continuation reuse.
- Enforces WebSocket text/size limits, documents the HTTP and WebSocket Responses API surfaces, and adds CI coverage for the integration tests.

## Test Plan
- `cargo test -p agentic-server --test responses_websocket_test`
- `cargo fmt -- --check`
- `git diff --check`
- `cargo test`
- `cargo test --workspace`
- `cargo build`
- `SKIP=no-commit-to-branch uvx pre-commit run --all-files`
- `uv run --with-requirements docs/requirements.txt mkdocs build --strict`


.github/workflows/rust.yml
```diff
@@ -48,3 +48,6 @@ jobs:
 
       - name: Run tests
         run: cargo test
+
+      - name: Run Responses WebSocket integration tests
+        run: cargo test -p agentic-server --test responses_websocket_test
```
maralbahari: the `cargo test`  runs all tests including the `agentic-server` test why need this separately?

crates/agentic-server/src/handler.rs
```diff
@@ -183,3 +194,195 @@ pub async fn responses(State(state): State<AppState>, req: Request) -> Response
...
 }
+
+pub async fn responses_ws(State(state): State<AppState>, headers: HeaderMap, ws: WebSocketUpgrade) -> Response {
+    ws.max_message_size(MAX_BODY_SIZE)
+        .max_frame_size(MAX_BODY_SIZE)
+        .on_upgrade(move |socket| responses_ws_loop(socket, state, headers))
+}
+
+async fn responses_ws_loop(mut socket: WebSocket, state: AppState, headers: HeaderMap) {
+    while let Some(message) = socket.recv().await {
```
maralbahari: The while `let Some(message) = socket.recv().await`  loop has no shutdown/cancellation path. If the server needs to shut down gracefully, there is no way to interrupt this loop. Consider adding a `CancellationToken` branch via tokio::select!:


`tokio::select! {
    msg = socket.recv() => { /* handle */ }
    _ = cancel_token.cancelled() => break,
}`

crates/agentic-server/src/handler.rs
```diff
@@ -183,3 +194,195 @@ pub async fn responses(State(state): State<AppState>, req: Request) -> Response
...
+            Ok(Message::Pong(_)) => {}
+            Err(e) => {
+                warn!("responses websocket receive error: {e}");
+                break;
+            }
+        }
+    }
+}
+
+async fn handle_ws_text(socket: &mut WebSocket, state: &AppState, headers: &HeaderMap, text: &str) -> bool {
```
maralbahari: I think this function is too long. would be good to do some refactoring. like extracting the streaming + persistence block into a helper function.

and the `handler.rs` has become large maybe could add a WebSocket handler file?

crates/agentic-server/src/handler.rs
```diff
@@ -183,3 +194,195 @@ pub async fn responses(State(state): State<AppState>, req: Request) -> Response
...
+    let mut lines = Vec::new();
+    let mut stream = Box::pin(call_inference(
+        upstream_json,
+        exec_ctx.responses_url(),
+        Arc::clone(&exec_ctx.client),
+        exec_ctx.client_auth.clone(),
+        exec_ctx.streaming_timeout,
+    ));
+
+    while let Some(line) = stream.next().await {
```
maralbahari: This stream loop has no cancellation path. If the client disconnects mid-stream, the server keeps streaming from the LLM and accumulating lines.

crates/agentic-server/src/handler.rs
```diff
@@ -183,3 +194,195 @@ pub async fn responses(State(state): State<AppState>, req: Request) -> Response
...
+}
+
+async fn send_executor_error(socket: &mut WebSocket, err: ExecutorError) -> bool {
+    let status = err.http_status();
+    let code = err.error_code();
+    let message = err.to_string();
+    send_ws_error(socket, status, code, &message).await
+}
+
+async fn send_ws_error(socket: &mut WebSocket, status: StatusCode, code: &str, message: &str) -> bool {
```
maralbahari: looking at all the error handling  `handler.rs` now having ws errors. I think it would be cleaner and easier if we maintain an enum server error for `agentic-server` crate as following:
```
#[derive(Debug, Error)]
pub enum ServerError {
    /// Executor-layer failure, forwarded from agentic-core.
    #[error(transparent)]
    Executor(#[from] ExecutorError),

    /// WebSocket send failed (connection dropped by client).
    #[error("websocket send failed")]
    WsSendFailed,

    /// Client sent a binary frame where JSON text was expected.
    #[error("websocket binary frames are not supported")]
    WsBinaryFrame,

    /// WebSocket message was not valid JSON.
    #[error("invalid JSON: {0}")]
    WsInvalidJson(#[source] serde_json::Error),

    /// WebSocket message had unexpected `type` field.
    #[error("unexpected message type: expected response.create")]
    WsUnexpectedType,
}

impl ServerError {
    /// Serialize to the WS error envelope sent to the client.
    pub fn to_ws_frame(&self) -> Value { ... }
}
```

this gives us several benefits:
- `From<ExecutorError>` so `?` works in `handle_ws_text` without `send_executor_error` adapter
- `From<serde_json::Error>` for the JSON parse case
- A single `to_ws_frame()` method replacing the ad-hoc `send_ws_error` / `send_executor_error` pair

Clean separation: `agentic-core` stays transport-agnostic, `agentic-server `owns transport errors
`send_ws_error` and `send_executor_error` would collapse into one function: `send_ws_err(socket, ServerError) -> bool`.

franciscojavierarceo: I did 

```rust
#[derive(Debug, thiserror::Error)]
enum WsError {
    #[error(transparent)]
    Executor(#[from] ExecutorError),

    #[error("invalid JSON: {0}")]
    InvalidJson(#[source] serde_json::Error),

    #[error("websocket message type must be response.create")]
    UnexpectedType,

    #[error("websocket messages must be JSON text frames")]
    BinaryFrame,

    #[error("websocket send failed")]
    SendFailed,

    #[error("websocket client disconnected")]
    ClientDisconnected,
}
```

and kept it local to `WsError` for now. in the future we can broaden it into a crate-level `ServerError` and move HTTP response mapping over

crates/agentic-server/src/handler.rs
franciscojavierarceo: done

.github/workflows/rust.yml
franciscojavierarceo: nice catch, thanks

crates/agentic-server/src/handler.rs
franciscojavierarceo: adjusted

crates/agentic-server/src/handler.rs
franciscojavierarceo: thanks for catching this

crates/agentic-server/src/handler.rs
maralbahari: @franciscojavierarceo sorry I think there was no change for this the function to breakdown into smaller functions and `handler.rs` became even larger now after adding the error enums. is it ok to move WebSocket related functions to to a separate `ws_handler.rs`?

leseb (approved):

PR #82, opened 2026-07-02
docs: MCP gateway integration design

maralbahari:
## Summary

Adds `docs/design/mcp-gateway-integration.md` outlining the design for integrating MCP into the agentic-api gateway. Covers `McpClient`, `McpClientPool` (HTTP and stdio transports), `McpHandler`/`McpHandlerKind`, `ToolRegistry` dispatch, `ResponseAccumulator` extension for tool dispatch, and the `execute_with_mcp()` agentic loop. Includes a roadmap for stdio server support and a two-PR split for infrastructure vs execution loop.

## Test Plan

- Design doc only, no code changes.

### DIFF
--- a/docs/design/mcp-gateway-integration.md
+++ b/docs/design/mcp-gateway-integration.md
@@ -1,0 +1,411 @@
+# MCP Gateway Integration
+
+Target: `crates/agentic-core/`
+Reference: [mcp handlers](https://github.com/openai/codex/tree/main/codex-rs/core/src/tools/handlers), [rmcp-client](https://github.com/openai/codex/tree/main/codex-rs/rmcp-client), [turn.rs](https://github.com/openai/codex/blob/main/codex-rs/core/src/session/turn.rs), [rust-sdk README](https://github.com/modelcontextprotocol/rust-sdk#readme)
+
+---
+
+## Goal
+
+This design uses [`rmcp`](https://github.com/modelcontextprotocol/rust-sdk) as the protocol layer for connecting to remote MCP servers. The MCP spec in rmcp gives us two patterns for implementing built-in tools inside agentic-api. Each pattern becomes a `GatewayExecutor` registered in `ToolRegistry` under a fixed tool name:
+
+1. **MCP Resources** ([rust-sdk Resources](https://github.com/modelcontextprotocol/rust-sdk#resources)): for tools that read or fetch content by URI. `read_mcp_resource` is the first example of this pattern.
+2. **MCP Tools** ([rust-sdk `#[tool]` macro](https://github.com/modelcontextprotocol/rust-sdk#tools)): for standalone MCP server tools that expose computation or actions. The `#[tool]` macro auto-generates the JSON schema, validation, and MCP wire format for a server that agentic-api connects to via `McpClientPool`. Examples:
+   - **stdio MCP server** (`connect_stdio`): a locally spawned process (command + args) the gateway connects to over stdin/stdout. Users wire up their own MCP server process and the gateway talks to it via rmcp's `TokioChildProcess` transport.
+   - **calculator** (reference example from [rust-sdk Tools](https://github.com/modelcontextprotocol/rust-sdk#tools)): simplest possible `#[tool]` server, useful as a template.
+
+### First built-in tool: `read_mcp_resource`
+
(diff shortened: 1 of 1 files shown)


franciscojavierarceo (approved):
this lgtm as a first cut, let's iterate and see

docs/design/mcp-gateway-integration.md
```diff
@@ -0,0 +1,411 @@
...
+```
+
+The rest of the stack (`McpHandler`, `build_mcp_registry`, dispatch loop) is unchanged. Stdio servers are just another entry in the pool.
+
+---
+
+## Implementing MCP and the First Built-in Tool
+
+### Step 1: `McpClient` (`mcp/client.rs`)
+
```
ashwing, line 69: This embeds `handler: Option<Arc<dyn GatewayExecutor>>` directly in `ToolEntry` and adds `ToolRegistry::dispatch()`. PR #83 (currently open) goes the other direction: a separate `HashMap<ToolType, Arc<dyn GatewayExecutor>>` built at the call site and passed into `dispatch_tools`. Both are landing simultaneously and they're incompatible — one puts executor ownership in the registry, the other keeps it external. Before either the MCP loop PR or any downstream work can proceed cleanly, we need to pick one model and have the other converge to it. What's the intended resolution here?

docs/design/mcp-gateway-integration.md
```diff
@@ -0,0 +1,411 @@
...
+Thin async wrapper around `rmcp::service::RunningService<RoleClient>`. Connects over HTTP/SSE (streamable HTTP transport).
+
+```rust
+pub struct McpClient {
+    inner: Arc<rmcp::service::RunningService<RoleClient>>,
+    tool_timeout: Duration,
```
ashwing, line 75: Adding `handler` to `ToolEntry` means every existing construction site in `ToolRegistry::build()` — for `Function`, `WebSearch`, `FileSearch`, `CodeInterpreter` — needs `handler: None`. Not a design concern, just a practical heads-up that this touches more of the codebase than the doc implies.

docs/design/mcp-gateway-integration.md
```diff
@@ -0,0 +1,411 @@
...
+One `McpClient` per server, keyed by `server_label`. Has two constructors: one built from the client request (`from_params`), one built from gateway config at startup (`from_config`, for gateway-managed stdio/HTTP servers). The `ReadResource` handler holds an `Arc<McpClientPool>` and uses `get()` at call time to route to the server named in the model's arguments.
+
+```rust
+pub struct McpClientPool {
+    clients: HashMap<String, Arc<McpClient>>,
+}
+
+impl McpClientPool {
+    pub async fn from_params(params: &[McpToolParam]) -> Self
+    pub async fn from_config(servers: HashMap<String, McpServerEntry>) -> Self
```
ashwing, line 101: `build_mcp_registry()` constructs an `McpHandler` for `ReadResource` using `McpClient::placeholder()` because that path routes through the pool and never uses the client directly. If `placeholder()` is ever called on a method that isn't expected to be a no-op, it silently misbehaves rather than failing. Could the `client` field on `McpHandler` be `Option<Arc<McpClient>>` for variants that don't need it, so the absence is explicit in the type rather than hidden behind a sentinel?

docs/design/mcp-gateway-integration.md
```diff
@@ -0,0 +1,411 @@
...
+                McpHandlerKind::ReadResource { pool } => {
+                    let args: ReadResourceArgs = serde_json::from_str(arguments)?;
+                    let client = pool.get(&args.server)
+                        .ok_or_else(|| ToolError::Execution(format!("unknown server: {}", args.server)))?;
+                    let result = client.read_resource(&args.uri).await?;
+                    Ok(ToolOutput { call_id: String::new(), output: read_result_to_text(&result) })
+                }
+            }
+        })
+    }
```
ashwing, line 148: The ordering guarantee needs to be explicit: if the LLM emits parallel tool calls in one turn, tool B blocks on tool A completing because dispatch is sequential. If that's intentional for this phase, document it explicitly so it doesn't get treated as a bug when someone later assumes concurrent dispatch.

docs/design/mcp-gateway-integration.md
```diff
@@ -0,0 +1,411 @@
...
+    // existing: HTTP/SSE transport
+    pub async fn connect(server_url: &str, headers: Option<HashMap<String, String>>) -> Result<Self, McpError>
+
+    // new: spawns command and connects over stdin/stdout
+    pub async fn connect_stdio(command: &str, args: &[&str]) -> Result<Self, McpError>
+}
+```
+
+`McpClientPool` gains a config-based constructor for gateway-managed servers:
+
```
ashwing, line 38: When `rmcp` is added to `Cargo.toml`, pin the version and flag that `modelcontextprotocol/rust-sdk` is pre-1.0 — API stability isn't guaranteed across minor bumps, and reviewers should evaluate that risk consciously at add time.

ashwing (commented):
The dispatch model in this doc conflicts with what PR #83 just introduced, and that's the main thread to pull before this moves forward. The other points are smaller — construction site impact, a fragile placeholder pattern, and the new `rmcp` dependency — but they're all downstream of resolving which dispatch interface wins.

PR #84, opened 2026-07-02
feat: codex integration

haoshan98:
# Typed Codex Namespace Support For Responses Gateway

## Summary

This PR adds typed, stateful Codex namespace tool support to the Responses gateway path.

Codex can send `type: "namespace"` tool declarations, while vLLM expects ordinary function-tool declarations. The
gateway now keeps the Codex-facing tool shape at the request/storage boundary, flattens namespace members only when
building the upstream request, and restores upstream flat function calls back to Codex's public `{ namespace, name }`
shape in final and streaming responses.

Main changes:

- Add `ResponsesTool::Namespace` and Codex namespace member types.
- Add `CodexNamespaceHandler` and `NamespaceMap` for request-scoped flatten/restore.
- Flatten namespace member names inside `RequestPayload::to_upstream_request()`.
- Rewrite namespaced `tool_choice` values to the same flattened names sent upstream.
- Restore flat namespace function calls in final payloads and SSE/WebSocket events.
- Preserve effective tools, tool choice, instructions, and continuation metadata for stateful follow-up requests.
- Keep the raw proxy path transparent; namespace normalization lives in the typed executor path.
- Remove gateway-side model aliasing from this PR.
- Keep `SKIP_LLM_READY_CHECK` and `scripts/codex-start-gateway.sh` for manual demo runs.

The intended runtime path is:

```text
Codex CLI -> agentic-api /v1 -> vLLM /v1 -> served model
```

## Gateway Flow

Codex can declare a namespace tool:

```json
{
  "type": "namespace",
  "name": "mcp__agentic_fixture",
  "tools": [
    { "type": "function", "name": "add_numbers" }
  ]
}
```

The request and storage layers keep that public shape as `ResponsesTool::Namespace`.

When the gateway builds the upstream request, `RequestPayload::to_upstream_request()` calls
`CodexNamespaceHandler::resolve_namespace_members()` and turns namespace members into model-visible function names:

```text
agentic_ns__mcp__agentic_fixture__add_numbers
```

The upstream request receives only `FunctionTool` declarations. The namespace container itself is not sent to vLLM.

When the model calls the flattened function, the gateway uses the request's `NamespaceMap` to restore the Codex-facing
shape:

```json
{
  "type": "function_call",
  "namespace": "mcp__agentic_fixture",
  "name": "add_numbers"
}
```

The same mapping is used for:

- explicit namespaced `tool_choice`
- final response payload output
- streaming SSE event payloads
- WebSocket response events

## Tool Registry Integration

Namespace support is wired through the existing tool abstractions rather than a separate request-side tool model.

- `ResponsesTool` remains the public request/storage enum.
- `CodexNamespaceHandler::resolve_namespace_members()` performs the pure request-scoped rename pass.
- `ResponsesTool::to_function_tools()` converts the already-renamed typed tools into upstream `FunctionTool` values.
- `ToolRegistry::build_with_handlers()` resolves namespace members before registering model-visible tool names.
- `ToolRegistry::restore_final_payload_output()` restores completed response payloads.
- `ToolRegistry::restore_stream_event_value()` restores streaming event payloads.

This keeps the special Codex namespace behavior in one handler while preserving the registry/handler convention used by
the rest of the gateway.

## Stateful Continuation

Stateful Responses execution stores the effective request metadata needed by later turns:

- tools
- tool choice
- instructions
- previous response linkage
- conversation linkage

Follow-up requests using `previous_response_id` or `conversation_id` rehydrate prior context and inherit effective tool
metadata unless the client explicitly overrides it.

`tool_choice` is represented as `Option<ToolChoice>` on `RequestPayload`, so rehydration can distinguish a missing
choice from an explicitly provided choice. Missing request-side tool choice resolves to upstream `auto`; explicit values
remain explicit.

Instructions stay as top-level Responses `instructions`. The gateway does not synthesize an extra system message.

## Proxy And Routing Scope

This PR intentionally keeps raw proxy behavior small and transparent.

HTTP `/v1/responses` can still use the proxy path for stateless, function-only requests that do not need gateway
execution:

- `store = false`
- no `previous_response_id`
- no `conversation_id`
- no non-function request tools

Requests that need storage, continuation, WebSocket handling, gateway-owned tools, or Codex namespace tools go through
the typed executor path.

Out of scope for this PR:

- raw proxy namespace flatten/restore
- gateway-side model aliasing
- a gateway-side Codex runtime

Codex still owns client-side tool execution.

## Collision Handling

The model-visible namespace member format is:

```text
agentic_ns__{namespace}__{member}
```

The `agentic_ns__` prefix avoids provider-reserved prefixes such as `mcp__` and marks generated names as gateway-owned
flattened namespace members.

The namespace map builder avoids flattening a namespace member over an existing top-level tool name. It also warns when
two namespace/member pairs flatten to the same model-visible string, because MCP-style names often contain `__` and can
otherwise become ambiguous.

## Cassettes And Tests

This PR adds Codex cassette coverage under `crates/agentic-core/tests/cassettes/codex/` for:

- gateway HTTP/SSE function tools
- gateway HTTP/SSE namespace tools
- gateway WebSocket function tools
- gateway WebSocket namespace tools
- direct vLLM flattened namespace tools
- direct OpenAI namespace baseline behavior

It also adds recorder/support scripts for regenerating those cassettes.

## Change Size Analytics

| Category | Files | Added | Deleted |
|---|---:|---:|---:|
| Production + bench code | 34 | 1,789 | 135 |
| Test code | 8 | 1,212 | 67 |
| YAML cassettes | 10 | 7,608 | 0 |
| JSON fixtures | 4 | 74 | 0 |
| Scripts | 4 | 1,135 | 34 |
| Docs | 1 | 13 | 24 |
| **Total** | **61** | **11,831** | **260** |

Most of the PR size is recorded cassette data: YAML cassettes account for 7,608 added lines, about 64% of total
additions. The production and benchmark code portion is 1,789 added lines and 135 deleted lines.

Bucket rules used for this snapshot:

- `*.yaml` / `*.yml` files are counted as YAML cassettes.
- `*.json` cassette support files are counted as JSON fixtures.
- `scripts/*`, `*.sh`, and `*.py` files are counted as scripts.
- Remaining files under `tests/` are counted as test code unless they match a cassette/fixture/script rule above.
- `docs/*` files are counted as docs.
- Everything else is counted as production + bench code.

## Test Plan

```bash
cargo fmt --check
cargo check -p agentic-core --tests
cargo check -p agentic-server --tests
cargo test -p agentic-core codex --quiet
cargo test -p agentic-core --test tool_normalization_test --quiet
cargo test -p agentic-core --test stateful_responses_integration --quiet
cargo test -p agentic-server --test responses_websocket_test --quiet
cargo clippy -p agentic-core --tests -- -D warnings
cargo clippy -p agentic-server --tests -- -D warnings
bash -n scripts/codex-start-gateway.sh
git diff --check
```

Important coverage:

- Namespace request tools flatten to `agentic_ns__...` upstream function declarations.
- Namespaced `tool_choice` values are rewritten before upstream serialization.
- Unqualified function `tool_choice` values are not accidentally rebound to namespace members.
- Final and streaming upstream flat tool calls are restored to Codex-facing `{ namespace, name }` calls.
- Namespace member collisions against top-level tool names are avoided.
- Flattened-name collisions between namespace/member pairs are detected and warned.
- Stateful continuation rehydrates effective tools and tool choice correctly.
- WebSocket Responses sessions restore namespace tool call events.


franciscojavierarceo:
awesome @haoshan98 let me know when this is ready for review

crates/agentic-core/src/tool/codex.rs
```diff
@@ -0,0 +1,812 @@
...
+    #[must_use]
+    pub fn restore_response_value(&self, value: &mut Value, tools: Option<&[ResponsesTool]>) -> bool {
+        let Some(map) = namespace_map_from_tools(tools) else {
+            return false;
+        };
+        restore_response_value_with_map(value, &map)
+    }
+
+    #[must_use]
+    pub fn flatten_raw_tools_for_upstream(
```
ashwing: The raw path mirrors the typed one — same flatten/restore rule, but on `serde_json::Value` + SSE-line string rewriting instead of `ResponsesTool` (restore itself is a clean map lookup — that's fine; it's the outbound SSE-line surgery that's fragile). Two impls of one rule will drift — worth centralizing on one core with thin typed/raw adapters over it.

crates/agentic-core/src/tool/codex.rs
```diff
@@ -0,0 +1,812 @@
...
+use super::registry::ToolType;
+
+// Upstream Responses-compatible backends only see flat function names. Prefix
+// flattened Codex namespace members so generated names are recognizable,
+// unlikely to collide with user functions, and can be restored to
+// `{ namespace, name }` on the way back to the client.
+pub const MODEL_VISIBLE_NAMESPACE_MEMBER_PREFIX: &str = "agentic_ns__";
+
+#[must_use]
+pub fn model_visible_namespace_member_name(namespace: &str, member: &str) -> String {
```
ashwing, line 18: `agentic_ns__{namespace}__{member}` joined on `__` — the builder catches flattened-vs-top-level collisions but not two namespace+member pairs flattening to the same string (ns `a__b`+member `c` vs ns `a`+member `b__c` both → `agentic_ns__a__b__c`); `record_flat_member` just inserts, silent last-write-wins. MCP names use `__` heavily, so not hypothetical. Worth detecting the flattened-name collision at build — even a warn — before this hardens.

ashwing (commented):
Took an early look since this is on the critical path for the tool-loop work in #83/#89 — flagging structure/scoping now while it's cheap to reshape, rather than after it's built out. cc @maralbahari

Big enough (and draft) that it's worth asking whether it wants to land as one PR or a few. From the outside it reads as ~three separable concerns: (a) the typed namespace model — `ResponsesTool::Namespace` + `flatten_tools_for_upstream`/`restore_output_items`; (b) the raw `store=false` proxy path; (c) model aliasing (`resolve_model_alias` + the proxy rewrite), which has no namespace dependency at all. Each is independently reviewable, and landing them in sequence would make this far more tractable than one 12k diff — aliasing especially looks like it could go first or separately with almost no coupling. Is there coupling between these that makes slicing harder than it looks?

maralbahari:
> Took an early look since this is on the critical path for the tool-loop work in #83/#89 — flagging structure/scoping now while it's cheap to reshape, rather than after it's built out. cc @maralbahari
> 
> Big enough (and draft) that it's worth asking whether it wants to land as one PR or a few. From the outside it reads as ~three separable concerns: (a) the typed namespace model — `ResponsesTool::Namespace` + `flatten_tools_for_upstream`/`restore_output_items`; (b) the raw `store=false` proxy path; (c) model aliasing (`resolve_model_alias` + the proxy rewrite), which has no namespace dependency at all. Each is independently reviewable, and landing them in sequence would make this far more tractable than one 12k diff — aliasing especially looks like it could go first or separately with almost no coupling. Is there coupling between these that makes slicing harder than it looks?

agree we can separate out the support case on proxy path into separate PR. main focus of this PR to land in first for gateway support. however I am not sure for model aliasing would be needed for gateway @haoshan98 ?

haoshan98:
> > Took an early look since this is on the critical path for the tool-loop work in #83/#89 — flagging structure/scoping now while it's cheap to reshape, rather than after it's built out. cc @maralbahari
> > Big enough (and draft) that it's worth asking whether it wants to land as one PR or a few. From the outside it reads as ~three separable concerns: (a) the typed namespace model — `ResponsesTool::Namespace` + `flatten_tools_for_upstream`/`restore_output_items`; (b) the raw `store=false` proxy path; (c) model aliasing (`resolve_model_alias` + the proxy rewrite), which has no namespace dependency at all. Each is independently reviewable, and landing them in sequence would make this far more tractable than one 12k diff — aliasing especially looks like it could go first or separately with almost no coupling. Is there coupling between these that makes slicing harder than it looks?
> 
> agree we can separate out the support case on proxy path into separate PR. main focus of this PR to land in first for gateway support. however I am not sure for model aliasing would be needed for gateway @haoshan98 ?

Will separate proxy changes so this one stays focused on typed/stateful gateway namespace support. Model aliasing is not strictly needed, just an optional Codex UX/routing convenience, can keep it later as well.

haoshan98:
@franciscojavierarceo @leseb  Proxy-path namespace support will be split out and land in a follow-up PR; this PR keeps namespace handling on the executor path only. Ready for review.

crates/agentic-core/src/tool/codex.rs
haoshan98: Solid catch, added a build-time warning for flattened-name collisions in `record_flat_member` with a test covering the `a__b+c` vs `a+b__c `case. Note the behaviour on collision is still last-write-wins. A proper fix needs a collision-free encoding, which I'd like to do as a follow-up.

franciscojavierarceo:
took another pass with the Rust/API-shape lens. i think the core implementation is in pretty good shape, but i’d revise a few things before merging:

1. `docs/design/codex-integration.md` still reads like the older compatibility-only plan. It says namespace flatten/restore lives in `tool::normalize`, and the “Current PR Scope” section says this PR should mostly accept/preserve Codex shapes and not build more framework behavior. The actual PR now ships typed namespace flatten/restore, tool-choice rewriting, stream restoration, and continuation metadata. I think the doc should describe the current landed behavior so future readers don’t have to infer it from the tests.

2. The namespace collision fallback in `tool/codex.rs` deserves a clearer invariant. When a generated namespace member name would collide with a top-level function, the code leaves that namespace member unrenamed. That avoids generating the duplicate flat name, but it also seems like the upstream model may see a bare function name that won’t restore to `{ namespace, name }` later. I’d rather see this path explicitly rejected, skipped with a clear warning, or covered by a focused test proving the returned public shape is intentional.

3. `ToolChoice::Function { name }` is still a raw `String`, while tool declarations use `NonEmptyToolName`. It would be cleaner to validate `tool_choice` the same way, or otherwise reject an empty function tool choice at deserialize/normalization time.

I checked the proposed `cargo test -p agentic-core codex --quiet` command on the PR head and it does run a real filtered slice, so no concern there.

franciscojavierarceo (approved):
this is great! 

I'm going to cut a follow up PR to handle my feedback so we can land this.

ashwing:
Thanks for folding in the feedback. Since it's already merged, quick status on where the two points landed:

- **Slicing**: the proxy path and model-aliasing both came out, so the typed/stateful namespace support stands on its own — that's the shape that makes the #83 loop consolidation clean.
- **Flat-name collision**: the build-time `warn!` catches it at construction, and as you noted it's still last-write-wins on `agentic_ns__{ns}__{member}`. Francisco's pre-merge note lands in the same place and he's already cutting a follow-up — one concrete case to cover there: `mcp__a`+`b` and `mcp`+`a__b` both flatten to `agentic_ns__mcp__a__b`, silently dropping a tool. A collision-free encoding closes it.

I'll pick up the `RequiresClientAction` consolidation from the #83 side against the landed client path.

PR #85, opened 2026-07-02
feat: add web search gateway tool

franciscojavierarceo:
## Summary

- Adds gateway-managed `web_search_preview` support backed by You.com via `YOU_API_KEY` and `YOU_API_BASE_URL`, normalizing the tool to the `web_search` function shape expected by the model backend.
- Executes web search calls server-side, feeds `function_call_output` back into follow-up model rounds, and returns OpenAI-compatible `web_search_call` output items to clients.
- Emits web search lifecycle SSE events during streaming tool loops, executes multiple gateway tool calls concurrently, and aligns the gateway tool loop limit to 10 rounds.
- Adds recorded gpt-oss/vLLM gateway cassettes for blocking and streaming web search turns, plus cassette replay assertions for `web_search_call` output and lifecycle events.
- Splits the You.com HTTP integration behind an internal `WebSearchProvider` abstraction so additional search providers can be added without changing the OpenAI tool adapter.

## Test Plan

- `cargo test`
- `cargo fmt -- --check`
- `git diff --check`
- Live smoke: `openai/gpt-oss-20b` on vLLM via local `agentic-server`, with `YOU_API_BASE_URL=https://ydc-index.io`, successfully executed `web_search_preview` for `potato` and returned a completed `web_search_call` plus final answer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,12 @@ cargo build
 cargo test
 ```
 
+## Web search
+
+The stateful `/v1/responses` executor supports OpenAI-compatible `web_search`
+tool declarations by normalizing them into a `web_search` function call for
+vLLM. Set `YOU_API_KEY` to enable execution through You.com's Search API.
+
 ## Lint and format
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -27,7 +27,8 @@ cargo test
 
 The stateful `/v1/responses` executor supports OpenAI-compatible `web_search`
 tool declarations by normalizing them into a `web_search` function call for
-vLLM. Set `YOU_API_KEY` to enable execution through You.com's Search API.
+vLLM. Set `YOU_API_KEY` and `YOU_API_BASE_URL` to enable execution through
+You.com's Search API.
 
 ## Lint and format
 
--- a/crates/agentic-core/src/events/normalize.rs
+++ b/crates/agentic-core/src/events/normalize.rs
@@ -57,6 +57,7 @@ fn classify_event_type(type_str: &str) -> SSEEventType {
         "response.reasoning_summary_text.done" => SSEEventType::ReasoningSummaryTextDone,
         "response.file_search_call.searching" => SSEEventType::FileSearchCallSearching,
         "response.file_search_call.completed" => SSEEventType::FileSearchCallCompleted,
+        "response.web_search_call.in_progress" => SSEEventType::WebSearchCallInProgress,
         "response.web_search_call.searching" => SSEEventType::WebSearchCallSearching,
         "response.web_search_call.completed" => SSEEventType::WebSearchCallCompleted,
         _ => SSEEventType::Other,
@@ -90,6 +91,7 @@ fn extract_payload(event_type: SSEEventType, json: &Value) -> EventPayload {
         | SSEEventType::ReasoningPartDone
         | SSEEventType::FileSearchCallSearching
         | SSEEventType::FileSearchCallCompleted
+        | SSEEventType::WebSearchCallInProgress
         | SSEEventType::WebSearchCallSearching
         | SSEEventType::WebSearchCallCompleted
         | SSEEventType::Other => EventPayload::Raw(json.clone()),
--- a/crates/agentic-core/src/events/types.rs
+++ b/crates/agentic-core/src/events/types.rs
@@ -88,6 +88,7 @@ pub enum SSEEventType {
     // Built-in tool calls
     FileSearchCallSearching,
     FileSearchCallCompleted,
+    WebSearchCallInProgress,
     WebSearchCallSearching,
     WebSearchCallCompleted,
 
--- a/crates/agentic-core/src/executor/engine.rs
+++ b/crates/agentic-core/src/executor/engine.rs
@@ -9,6 +9,9 @@ use std::sync::Arc;
 
 use async_stream::stream;
 use either::Either;
+use futures::future::join_all;
+use serde_json::Value;
+use tokio::sync::mpsc;
 use tracing::warn;
 
 use crate::executor::accumulator::ResponseAccumulator;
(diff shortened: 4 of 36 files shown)


maralbahari:
@franciscojavierarceo  Thanks for the PR. I just have several concerns. the `engine.rs` has become too large that is handling so many things in one file. I think we could reduce the code as follow:

`engine.rs` is accumulating tool-specific logic and conversion logic. For this PR, I think we should move the web-search-specific pieces behind the existing gateway tool interface and let the registry own dispatch, so the executor stays mostly orchestration-only.

First, `WebSearchHandler` in `tool/web_search.rs` should implement `GatewayExecutor` fully. Its `execute()` should own the web-search-specific behavior: parsing web search arguments, validating query/config, calling the provider, serializing the result, and shaping web-search output.

Example of what should move out of `engine.rs` / executor-adjacent code into `tool/web_search.rs`:

```rust
impl GatewayExecutor for WebSearchHandler {
    fn execute(
        &self,
        call_id: &str,
        tool_name: &str,
        arguments: &str,
        config: &Value,
    ) -> Pin<Box<dyn Future<Output = Result<ToolOutput, ToolError>> + Send + '_>> {
        Box::pin(async move {
            if tool_name != "web_search" {
                return Err(ToolError::Config(...));
            }

            let args = WebSearchArguments::from_json(arguments)?;
            let config = serde_json::from_value::<WebSearchToolParam>(config.clone())?;

            let response = self.provider.search(&args, &config).await?;

            let output = serde_json::to_string(&json!({
                "query": response.query,
                "results": response.results,
                "metadata": response.metadata,
            }))?;

            Ok(ToolOutput {
                call_id: call_id.to_owned(),
                output,
            })
        })
    }
}
```

Web-search-specific public output conversion can also live there:

```rust
pub fn web_search_output_item(
    call: &FunctionToolCall,
    output: &ToolOutput,
    status: WebSearchCallStatus,
) -> OutputItem {
    // parse output.output
    // extract query
    // extract web/news sources
    // return OutputItem::WebSearchCall(...)
}
```

Then `engine.rs` only asks the web search module to build the public output item, instead of knowing how to parse web search results.

Second, dispatch should be owned by `ToolRegistry` instead of manually assembled in `engine.rs`. Right now `engine.rs` is doing lookup, executor selection, config extraction, and execution. That can be hidden behind `registry.dispatch(call)`.

Example shape:

```rust
pub struct ToolEntry {
    pub tool_type: ToolType,
    pub config: Value,
    pub server_label: Option<String>,
    pub handler: Option<Arc<dyn GatewayExecutor>>,
}

impl ToolRegistry {
    pub async fn dispatch(
        &self,
        call: &FunctionToolCall,
    ) -> Option<Result<ToolOutput, ToolError>> {
        let entry = self.entries.get(&call.name)?;
        let handler = entry.handler.as_ref()?;

        Some(
            handler
                .execute(
                    &call.call_id,
                    &call.name,
                    &call.arguments,
                    &entry.config,
                )
                .await,
        )
    }
}
```

Then `engine.rs` becomes generic:

```rust
let Some(result) = registry.dispatch(call).await else {
    // client-owned function tool
    continue;
};

let output = match result {
    Ok(output) => output,
    Err(ToolError::Execution(message) | ToolError::Config(message)) => {
        execution_error_output(call, &message)?
    }
};
```

If gateway dispatch is extracted, `engine.rs` also does not need its own `function_calls()` helper. The accumulator already recognizes function calls and finalizes them into `OutputItem::FunctionCall`; the gateway layer can own scanning finalized output for dispatchable calls:

```rust
let gateway_results = gateway::execute_output_calls(&current_output, &registry).await?;
```

That keeps output inspection next to gateway dispatch instead of leaving another low-level helper in `engine.rs`.

Third, `append_output_items_to_input()` in `engine.rs` duplicates the same `OutputItem -> InputItem` conversion that already exists inside `InOutItem::into_input_items()` in `storage/types/item.rs`. We should probably factor that into a reusable single-item helper so these match statements do not drift.

Current duplicated mapping:

```rust
OutputItem::Message(msg) => Some(InputItem::Message(msg.into())),
OutputItem::Reasoning(r) => Some(InputItem::Reasoning(r)),
OutputItem::FunctionCall(f) => Some(InputItem::FunctionCall(f)),
OutputItem::WebSearchCall(_) | OutputItem::Unknown => None,
```

Possible helper:

```rust
impl OutputItem {
    pub fn to_input_item(&self) -> Option<InputItem> {
        match self {
            OutputItem::Message(msg) => Some(InputItem::Message(msg.clone().into())),
            OutputItem::Reasoning(r) => Some(InputItem::Reasoning(r.clone())),
            OutputItem::FunctionCall(f) => Some(InputItem::FunctionCall(f.clone())),
            OutputItem::WebSearchCall(_) | OutputItem::Unknown => None,
        }
    }
}
```

Then storage and executor code can share it:

```rust
// storage/types/item.rs
InOutItem::Output(output) => output.to_input_item()
```

```rust
// engine.rs
for input_item in output_items.iter().filter_map(OutputItem::to_input_item) {
    append_input_item(input, input_item);
}
```

This keeps `engine.rs` focused on orchestration: collect model tool calls, ask the registry to dispatch gateway-owned calls, append tool outputs, and avoid carrying web-search-specific or repeated conversion logic.

Some other follow-up extraction seams that would make `engine.rs` less dense without creating too many files:

```text
executor/gateway.rs
- GatewayCallDispatch / GatewayCallExecution / GatewayCallResult / GatewayCallEventPlan
- gateway_dispatches()
- execute_gateway_dispatch()
- execute_gateway_calls()
- gateway_public_output()
- gateway_event_plans()
- emit_gateway_start_events()
- emit_gateway_completed_events()
- emit_sse_json()
- output_item_value()
- error_sse_chunk()
- append_input_item()
- append_output_items_to_input()
- append_tool_outputs()
- append_gateway_calls_to_new_input()
- function_calls()
- public_output_items()

executor/upstream.rs
- fetch_blocking_payload()
- fetch_stream_payload()

executor/persist.rs
- should_persist()
- persist_if_needed()
```

Keeping these gateway-related helpers together in one file gives us a cleaner first extraction point, while still leaving room to refactor and design the gateway event flow separately later.

Two smaller cleanup notes while touching this code:

```rust
let public_output = public_output_items(current_output.clone(), &registry, &gateway_results);
```

Can we avoid cloning the full `current_output` here? `public_output_items()` could take `&[OutputItem]` and return a new vector, or the code could be reordered so `current_output` is consumed only after the later bookkeeping that still needs to borrow it.

Also, `emit_sse_json()` returns `ExecutorResult<()>`, but it currently ignores `sender.send(...)` failures with `let _ = ...`. If a closed SSE receiver should stop execution, we should map the send failure into `ExecutorError`. If it is intentionally non-fatal, can we make that explicit with a short comment/helper so this does not look like an accidentally dropped error?

franciscojavierarceo:
@maralbahari incorporated the feedback 🙏

maralbahari (approved):
@franciscojavierarceo Thank you for addressing the comments. The PR looks good now. we shall refactor the gateway execution and event emits in an extensible design later once we have more built-in tools.

PR #87, opened 2026-07-07
fix: persist effective tool config for response continuations

### LINKED ISSUE #86: Response continuation checkpoints drop inherited tool config
## Summary

When a Responses API request uses `previous_response_id` and omits `tools`, the executor correctly rehydrates the prior response metadata and sends the inherited tools/tool choice upstream. However, the persisted checkpoint for that follow-up turn stores `ctx.original_request.tools` and `ctx.original_request.tool_choice`, not the effective values from the enriched request that was actually sent upstream.

As a result, a two-turn continuation works, but the next continuation from the second response can lose the inherited tool configuration.

## Reproduction

1. Send turn 1 with a function tool, e.g. `lookup_weather`, and `store: true`.
2. Send turn 2 with `previous_response_id` pointing at turn 1, but omit `tools`.
3. The request sent upstream for turn 2 includes the inherited tool configuration.
4. Persist turn 2. Its metadata records `effective_tools: null` because the raw client request omitted `tools`.
5. Send turn 3 with `previous_response_id` pointing at turn 2. The server no longer has the inherited tool configuration to rehydrate.

## Impact

This makes response checkpoints inconsistent with the model-visible state used to generate them. Multi-turn tool workflows can work for one continuation and then silently lose tool availability on the next continuation, forcing clients to resend tools on every turn despite using server-side `previous_response_id` state.

## Expected behavior

Persisted response metadata should reflect the effective request configuration used for the model call. If a turn inherited tools/tool choice during rehydration, the response checkpoint for that turn should preserve those effective values for later continuations.

## References

- `docs/adr/ADR-02_response_store.md` says stored responses must preserve enough state to rehydrate the next turn, including the effective tool configuration and tool-choice information.
- `docs/index.md` describes `previous_response_id` as server-managed conversation history so clients do not need to replay state.
- The current response continuation path already rehydrates from `stored.metadata.effective_tools` in `crates/agentic-core/src/executor/rehydrate.rs`.

harivilasp:
## Summary

Fixes #86.

Persist response checkpoint metadata from the enriched request that was actually sent upstream, rather than the raw client request. This preserves inherited `tools`, `tool_choice`, and instructions after `previous_response_id` rehydration so later continuations do not lose model-visible state.

This matches the response-store intent in `docs/adr/ADR-02_response_store.md`, which requires stored responses to preserve enough state to rehydrate the next turn, including effective tool configuration and tool-choice information.

## Test Plan

- `cargo +stable test -p agentic-core`
- `cargo +stable fmt --all --check`

Added an integration regression test covering a three-checkpoint scenario: first turn defines a function tool, second turn inherits it via `previous_response_id`, and the persisted second response retains the inherited tool metadata for subsequent continuations.

### DIFF
--- a/crates/agentic-core/src/executor/modes/conversation.rs
+++ b/crates/agentic-core/src/executor/modes/conversation.rs
@@ -92,9 +92,9 @@ impl ConversationHandler {
         let metadata = ResponseMetadata {
             model: ctx.enriched_request.model,
             previous_response_id: ctx.original_request.previous_response_id,
-            effective_tools: ctx.original_request.tools,
-            effective_tool_choice: ctx.original_request.tool_choice,
-            effective_instructions: ctx.original_request.instructions,
+            effective_tools: ctx.enriched_request.tools,
+            effective_tool_choice: ctx.enriched_request.tool_choice,
+            effective_instructions: ctx.enriched_request.instructions,
         };
 
         let mut new_items = Vec::with_capacity(ctx.new_input_items.len() + output_items.len());
--- a/crates/agentic-core/src/executor/modes/response.rs
+++ b/crates/agentic-core/src/executor/modes/response.rs
@@ -72,9 +72,9 @@ impl ResponseHandler {
         let metadata = ResponseMetadata {
             model: ctx.enriched_request.model,
             previous_response_id: ctx.original_request.previous_response_id,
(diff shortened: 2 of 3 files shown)


maralbahari (approved):
@harivilasp thank you for the fix.

Issue #32, opened 2026-05-27
Enhance micro-benchmark harness to track proxy overhead under realistic conditions


ashwing:
### Motivation

As a Rust-based gateway proxying requests to vLLM's Responses API, agentic-api adds a network hop and processing overhead. The existing Criterion benchmark (`crates/agentic-server/benches/proxy_bench.rs`) establishes a baseline comparing direct-to-LLM vs proxied latency, but doesn't capture behavior under realistic conditions: concurrent load, varying payload sizes, or streaming TTFB. As the proxy evolves (e.g., #29's crate restructure), we need to detect regressions and understand where overhead comes from.

### Current State

The existing benchmark:
- Uses Criterion with `async_tokio` runtime
- Spawns a mock LLM server (fixed JSON or SSE chunks)
- Spawns the gateway pointing at the mock
- Measures four scenarios: `non_stream/direct`, `non_stream/proxied`, `stream/direct`, `stream/proxied`

This gives a good sense of raw proxy overhead in ideal conditions (serial requests, tiny payloads).

### Gaps

1. **Concurrency** — The benchmark runs one request at a time. Real deployments see 10-100+ concurrent requests. We need to measure how p50/p95/p99 latencies degrade under load, and whether the gateway introduces queuing or backpressure issues.

2. **Body size scaling** — Current payloads are ~50 bytes. Real Responses API payloads range from 1KB (simple tool calls) to 100KB+ (long prompts, large tool schemas). We need to measure whether overhead scales linearly, or if there are hidden copy/allocation bottlenecks.

3. **Streaming TTFB** — For SSE endpoints, what matters most is time-to-first-chunk, not total transfer time. The current benchmark measures total elapsed time, which masks TTFB degradation.

4. **Regression detection in CI** — The benchmarks exist but aren't tracked. Without baseline comparison (e.g., `criterion --save-baseline`), we won't catch regressions introduced by refactors or dependency updates.

### Proposed Approach

I acknowledge these micro-benchmarks are tricky — mock backends don't capture network variance, async benchmarks have high noise, and streaming TTFB measurement requires custom instrumentation. Phased approach:

**Phase 1: Concurrency and body size variants**
- Extend the existing benchmark with concurrency groups (1, 10, 50 concurrent requests)
- Add body size variants: 100B, 1KB, 10KB, 100KB
- This should reveal whether the gateway's forwarding introduces bottlenecks under load

**Phase 2: Streaming TTFB**
[...]

PR #88, opened 2026-07-07
feat: benchmark proxy payload scaling

harivilasp:
## Summary

Refs #32.

Extends the existing `agentic-server` proxy Criterion benchmark with Phase 1 coverage for proxy-overhead payload-size scaling:

- keeps the existing direct vs proxied non-streaming and streaming baselines
- sends explicit `store: false` request bodies so proxied cases exercise the proxy path
- adds non-streaming payload-size benchmarks for 1 KiB, 10 KiB, and 100 KiB prompts
- adds an integration test proving a below-limit 100 KiB `store: false` request reaches the mock vLLM on the proxy path with the expected request shape

The benchmark still uses the local mock LLM and local gateway, so it remains deterministic and does not require a live vLLM server.

Per review feedback, concurrency benchmarking is intentionally left out of this proxy-only benchmark and can be handled in `gateway_bench.rs` / database-backed gateway benchmarks.

## Benchmark Results

Run locally on this branch with:

```bash
cargo +stable bench -p agentic-server --bench benches -- 'non_stream/body_size'
```

| Prompt size | Direct median | Proxied median | Approx. overhead |
| --- | ---: | ---: | ---: |
| 1 KiB | 39.900 µs | 50.737 µs | +10.837 µs |
| 10 KiB | 44.493 µs | 53.605 µs | +9.112 µs |
| 100 KiB | 76.618 µs | 82.224 µs | +5.606 µs |

These numbers use the local mock upstream and should be treated as reference data for this benchmark harness, not production latency claims.

## Test Plan

- `cargo +stable fmt --all --check`
- `rtk test cargo +stable test -p agentic-server`
- `cargo +stable clippy -p agentic-server --all-targets -- -D warnings`
- `cargo +stable test -p agentic-server --benches --no-run`
- `cargo +stable bench -p agentic-server --bench benches -- 'non_stream/body_size'`

### DIFF
--- a/crates/agentic-server/benches/proxy_bench.rs
+++ b/crates/agentic-server/benches/proxy_bench.rs
@@ -1,4 +1,5 @@
 use std::convert::Infallible;
+use std::sync::Arc;
 
 use axum::body::Body;
 use axum::extract::Request;
@@ -6,7 +7,7 @@ use axum::response::{IntoResponse, Response};
 use axum::routing::{get, post};
 use axum::{Router, serve};
 use bytes::Bytes;
-use criterion::{Criterion, criterion_group};
+use criterion::{BenchmarkId, Criterion, criterion_group};
 use futures::stream;
 use http::StatusCode;
 use tokio::net::TcpListener;
@@ -18,7 +19,9 @@ use agentic_core::executor::{ConversationHandler, ExecutionContext, ResponseHand
 use agentic_core::proxy::ProxyState;
 use agentic_core::storage::{ConversationStore, ResponseStore};
 use agentic_server::app::{AppState, ServerConfig, build_router};
-use std::sync::Arc;
+
+const CONTENT_TYPE_JSON: &str = "application/json";
+const PAYLOAD_SIZES: [usize; 3] = [1024, 10 * 1024, 100 * 1024];
 
 fn bench_config(llm_url: &str) -> Config {
     Config {
@@ -58,7 +61,7 @@ async fn responses_handler(req: Request) -> Response {
     }
 
     let out = r#"{"id":"resp_bench","object":"response","status":"completed"}"#;
-    (StatusCode::OK, [("content-type", "application/json")], out).into_response()
+    (StatusCode::OK, [("content-type", CONTENT_TYPE_JSON)], out).into_response()
 }
 
 async fn spawn_llm() -> String {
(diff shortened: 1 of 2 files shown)


franciscojavierarceo:
@harivilasp can you resolve conflicts?

maralbahari:
@harivilasp Thanks for the PR. 
 I think it would be good to have concurrency benchmark be on `gateway_bench.rs` since all `proxy_bench.rs` does is to forward request to the vLLM upstream only. whereas the gateway benchmark includes conversation/responses rehydration path.  We have been prioritizing sqlite support for agentic-api gateway so to truly test concurrency on gateway would require testing with postgres.

maralbahari:
@harivilasp could you please add some benchmarking result in PR description for `direct` and `proxied` differences so that we can use for reference later on.

harivilasp:
Sure checking

harivilasp:
Thanks, that makes sense. This PR currently focuses on proxy payload-size overhead only and does not include a concurrency benchmark. I agree that a proper concurrency benchmark should live in gateway_bench.rs. I can keep this PR scoped to payload-size benchmarking and follow up separately with a gateway concurrency benchmark.

crates/agentic-server/benches/proxy_bench.rs
```diff
@@ -18,7 +19,9 @@ use agentic_core::executor::{ConversationHandler, ExecutionContext, ResponseHand
 use agentic_core::proxy::ProxyState;
 use agentic_core::storage::{ConversationStore, ResponseStore};
 use agentic_server::app::{AppState, ServerConfig, build_router};
-use std::sync::Arc;
+
+const CONTENT_TYPE_JSON: &str = "application/json";
+const PAYLOAD_SIZES: [usize; 3] = [1024, 10 * 1024, 100 * 1024];
```
ashwing: 1–100 KiB is a fine starting point. Just flagging that the allocation stuff #32 was worried about (Gap 2) really only shows at MiB scale — leseb mentioned file_data goes up to 32 MiB, image_url 20 MiB. Not for this PR, but when someone adds a 1 MiB+ tier: heads up the mock caps `to_bytes` at 10 MiB (~line 40), so anything at/above that plus the envelope will just fail to parse until the cap goes up. And repeated `x` is totally fine here, no need for real-looking payloads — the gateway just forwards the bytes, it never looks at them.

ashwing (commented):
Nice to finally have a number on whether forwarding cost scales with body size. Good timing too since the forward path is churning right now (#82 just merged, #85 is reworking gateway.rs).

One thing though: as it stands this runs once and prints numbers, and nobody's going to eyeball them against anything later. The whole reason #32 wanted this was regression detection, and that only kicks in once there's a saved baseline on main and PR branches get compared against it (`criterion --save-baseline` basically). That's the Phase 3 bit in the issue. I'd push to get that wiring in not long after this lands, since a regression from the gateway.rs rework is exactly the kind of thing it'd catch.

Few smaller things inline.

crates/agentic-server/benches/proxy_bench.rs
```diff
@@ -102,91 +105,110 @@ async fn spawn_gateway(config: Config) -> String {
...
+
+    for size in PAYLOAD_SIZES {
+        group.bench_with_input(BenchmarkId::new("direct", size), &size, |b, &size| {
+            let url = responses_url(llm_url);
+            let body = request_body(size, false);
+            b.to_async(rt)
+                .iter(|| post_response(client.clone(), url.clone(), body.clone()));
+        });
+
+        group.bench_with_input(BenchmarkId::new("proxied", size), &size, |b, &size| {
```
ashwing: Small thing: `BenchmarkId::new("proxied", size)` uses `size`, which is the repeat length, not the actual serialized body size (the json wrapper adds a chunk on top). Might be worth measuring `body.len()` and labelling with that, or just renaming it to prompt length, otherwise someone reads these later as literal payload bytes.

crates/agentic-server/benches/proxy_bench.rs
```diff
@@ -58,7 +61,7 @@ async fn responses_handler(req: Request) -> Response {
     }
 
     let out = r#"{"id":"resp_bench","object":"response","status":"completed"}"#;
-    (StatusCode::OK, [("content-type", "application/json")], out).into_response()
+    (StatusCode::OK, [("content-type", CONTENT_TYPE_JSON)], out).into_response()
```
ashwing, line 65: Could you note in the description that the mock always returns the same ~60-byte response no matter the request size? So this is really measuring the request/upload side only, not the response. Which is the right thing for the prompt-size question — just don't want anyone down the line taking these as full round-trip numbers.

crates/agentic-server/benches/proxy_bench.rs
harivilasp: Addressed in df1f463. I kept this PR at the current 1-100 KiB prompt-size tiers and left the MiB-scale `to_bytes` cap as a future concern rather than expanding scope here.

crates/agentic-server/benches/proxy_bench.rs
harivilasp: Addressed in df1f463. I renamed the size sweep to `prompt_bytes`/`PROMPT_SIZES` so the Criterion labels describe the repeated prompt length rather than implying exact serialized JSON body bytes.

crates/agentic-server/benches/proxy_bench.rs
harivilasp: Addressed in df1f463. I added a note by the mock response that it stays fixed-size so the prompt-size benchmark isolates request/upload overhead rather than response-size scaling.

franciscojavierarceo (approved):

PR #89, opened 2026-07-07
feat: add remote MCP gateway support with `read_resource` as first built-in MCP tool

maralbahari:
## Summary

This PR focuses on remote MCP over HTTP and implements read_mcp_resource as the initial built-in MCP gateway tool for HTTP MCP servers. A follow-up PR will add full stdio MCP support, including trusted command and args configuration.

Requests can now pass an MCP tool declaration such as:

```json
{
  "type": "mcp",
  "name": "read_mcp_resource",
  "server_label": "repo",
  "server_url": "http://127.0.0.1:8000/mcp"
}
```

The gateway normalizes the MCP tool into a model-facing function tool, executes the resulting `read_mcp_resource` call through an RMCP client, and returns a public `mcp_tool_call` output item.

## What Changed

- Added `ResponsesTool::Mcp` support.
- Added MCP tool params for remote HTTP MCP servers and stdio MCP server configuration.
- Added an RMCP client and request-scoped MCP client pool.
- Added the `read_mcp_resource` MCP gateway handler.
- Added MCP output item and streaming event mapping.
- Added MCP cassette recording support.
- Added MCP cassette tests in `crates/agentic-core/tests/mcp_tool_test.rs`.

## Codex `read_mcp_resource` Mapping

Codex normally exposes MCP resources to the model as a local function tool named `read_mcp_resource` with arguments like:

```json
{
  "server": "server_label",
  "uri": "repo://crates/agentic-server-core/tests/cassettes/web_search/gpt_oss_web_search_nonstreaming.yaml"
}
```

The gateway now recognizes Codex's `read_mcp_resource` function declaration when it includes MCP metadata. The metadata maps the model-provided `server` label to the MCP server URL, then the gateway registers the function as a gateway-owned MCP entry instead of treating it as a client-owned function.

Expected metadata shape:

```json
{
  "metadata": {
    "mcp_servers": {
      "server_label": {
        "server_url": "http://127.0.0.1:8000/mcp"
      }
    }
  }
}
```

The gateway handles the mapping when this metadata is present.

## Running Codex Through Agentic API

Configure Codex with both the agentic-api Responses provider and an MCP server label. For example, in `$CODEX_HOME/config.toml`:

```toml
model_provider = "agentic-api"

[model_providers.agentic-api]
name = "agentic-api"
base_url = "http://localhost:9000/v1"
wire_api = "responses"
requires_openai_auth = false
supports_websockets = true

[mcp_servers.your_server_label]
url = "http://127.0.0.1:8000/mcp"
default_tools_approval_mode = "approve"
```

Start an MCP server separately. For example, the RMCP streamable HTTP example server can be used for local testing:

```bash
cargo run -p mcp-server-examples --example servers_counter_streamhttp
```

Any MCP server can be used as long as the Codex `mcp_servers.<your_server_label>.url` value points to it and the model calls `read_mcp_resource` with the same `server_label`.

Start the agentic-api gateway on `http://localhost:9000/v1`, then run Codex with that config:

```bash
CODEX_HOME=/codexhome/.codex \
codex --disable image_generation \
  -c model_provider=agentic-api \
  -m model-name
```

Inside Codex, ask it to use `read_mcp_resource` with server `your_server_label` and a resource URI exposed by the MCP server.

## MCP Server Used For Recording

The cassettes were recorded with the RMCP streamable HTTP example server from `modelcontextprotocol/rust-sdk`.

The example server was run separately from this repo:

```bash
cargo run -p mcp-server-examples --example servers_counter_streamhttp
```

That server listens at:

```text
http://127.0.0.1:8000/mcp
```

For recording, the example server exposed a stable MCP resource URI that maps to a YAML cassette file in this repository:

```text
repo://crates/agentic-server-core/tests/cassettes/web_search/gpt_oss_web_search_nonstreaming.yaml
```

The model was asked to call `read_mcp_resource` for that URI and then summarize the YAML cassette. This verified that the gateway could connect to the MCP server, read a real resource, feed the result back to the model, and complete the response.

## How The Cassettes Were Recorded

Run the RMCP example server first, then from the agentic-api repository root:

```bash
MCP_SERVER_LABEL=server_label \
MCP_SERVER_URL=http://127.0.0.1:8000/mcp \
MCP_RESOURCE_URI=repo://crates/agentic-server-core/tests/cassettes/web_search/gpt_oss_web_search_nonstreaming.yaml \
bash crates/agentic-server-core/tests/cassettes/record_mcp_cassettes.sh
```

The successful MCP cassettes are:

```text
crates/agentic-server-core/tests/cassettes/mcp/mcp-read-resource-Qwen-Qwen3-30B-A3B-FP8-nonstreaming.yaml
crates/agentic-server-core/tests/cassettes/mcp/mcp-read-resource-Qwen-Qwen3-30B-A3B-FP8-streaming.yaml
```

The non-streaming cassette contains a completed `mcp_tool_call` for `read_mcp_resource` and a final assistant message summarizing the web search cassette.

The cassette script also records unhappy paths:

```text
crates/agentic-server-core/tests/cassettes/mcp/mcp-read-resource-unreachable-server-Qwen-Qwen3-30B-A3B-FP8-streaming.yaml
crates/agentic-server-core/tests/cassettes/mcp/mcp-read-resource-connection-refused-Qwen-Qwen3-30B-A3B-FP8-streaming.yaml
crates/agentic-server-core/tests/cassettes/mcp/mcp-read-resource-missing-resource-Qwen-Qwen3-30B-A3B-FP8-streaming.yaml
```

These cover the gateway's failure behavior when the MCP server URL is rejected or unavailable, when the URL is loopback but no MCP server is listening, and when the MCP server returns `resource_not_found` for the requested URI.

## Validation

Validated with:

```bash
cargo test -p agentic-server-core -- --test-threads=4
cargo clippy -p agentic-server-core --all-targets -- -D warnings
cargo test -p agentic-server-core --test mcp_tool_test
```


crates/agentic-core/src/tool/mcp/client.rs
```diff
@@ -0,0 +1,274 @@
...
+                custom_headers.insert(
+                    HeaderName::try_from(name).map_err(McpError::InvalidHeaderName)?,
+                    HeaderValue::try_from(value).map_err(McpError::InvalidHeaderValue)?,
+                );
+            }
+            config = config.custom_headers(custom_headers);
+        }
+        let transport = StreamableHttpClientTransport::from_config(config);
+        let service = AgenticMcpClientHandler
+            .serve(transport)
```
franciscojavierarceo: This probably wants the same timeout treatment as the later `list_tools` / `resources/read` calls. Right now a request-scoped MCP server can accept the HTTP connection and then never finish the initialize handshake, which leaves this request stuck before we even call the model. Could we wrap both this `serve(...)` and the stdio one below in `tokio::time::timeout(..., ...)` and return `McpError::Timeout { operation: McpOperation::Connect }`?

crates/agentic-server-core/tests/mcp_tool_test.rs
```diff
@@ -0,0 +1,163 @@
...
+    assert!(text.contains("web_search_preview"));
+    assert!(text.contains("Potato - Wikipedia"));
+    assert!(
+        output.iter().any(|item| matches!(item, OutputItem::Message(_))),
+        "cassette output should include a final assistant message"
+    );
+}
+
+#[test]
+fn read_mcp_resource_cassette_nonstreaming() {
```
ashwing, line 118: Coverage here is two happy-path `read_mcp_resource` cassettes (this + the streaming one). The `tools/call` discovered-tool execution path, `connect_stdio`, the unknown-server / connection-error path, and the 30s timeout all have no integration coverage — and those failure modes are most of the point for MCP (dead server, hung tool, bad args). The pool unit tests cover SSRF + config parsing well; it's the executor behavior on the sad paths that's unexercised. Worth cassettes/tests for at least `tools/call` + one error path before this is load-bearing.

ashwing (commented):
Read through client/pool/handler/registry + the tests. Comments inline — the main two are test coverage and the silent registry collision; the rest are confirm/observe. Also flagged the #83 merge overlap so we sequence it deliberately.

crates/agentic-core/src/tool/mcp/registry.rs
```diff
@@ -0,0 +1,78 @@
...
+                    tool_type: ToolType::Mcp,
+                    config,
+                    server_label: Some(server_label.clone()),
+                    handler: Some(handler),
+                },
+            ));
+        }
+    }
+
+    (specs, ToolRegistry::with_entries(entries))
```
ashwing: `with_entries` collects straight into a `HashMap`, so this is silent last-write-wins on key collision. Two servers exposing the same `{server_label}__{tool_name}`, or a discovered tool whose exposed name collides with the `read_mcp_resource` built-in, will silently overwrite one entry with no signal. `ToolRegistry::build_with_handlers` warns on duplicate names — this path should at least match that (warn, or reject).

crates/agentic-server-core/src/tool/mcp/pool.rs
```diff
@@ -0,0 +1,259 @@
...
+    connection_errors: HashMap<String, String>,
+}
+
+impl McpClientPool {
+    pub async fn from_params(params: &[McpToolParam]) -> Self {
+        let servers: HashMap<String, McpServerEntry> = params.iter().filter_map(server_entry_from_param).collect();
+        Self::from_config(servers).await
+    }
+
+    pub async fn from_config(servers: HashMap<String, McpServerEntry>) -> Self {
```
ashwing, line 47: The `server_url` allowlist (loopback/`AGENTIC_MCP_ALLOWED_HOSTS`, credentials rejected, metadata-endpoint test) runs in `server_entry_from_param` on the request path — not here in `from_config`. That's right if `from_config` is operator-only. Worth confirming no client-reachable path can reach `from_config`, since that's the boundary the whole SSRF control rests on.

crates/agentic-server-core/src/tool/mcp/handler.rs
```diff
@@ -0,0 +1,257 @@
...
+        server,
+        call.name.clone(),
+        arguments,
+        GatewayCallStatus::InProgress,
+        None,
+        None,
+    ))
+}
+
+pub enum McpHandlerKind {
```
ashwing, line 79: `ReadResourceSpec`/`ToolCallSpec` implement `GatewayExecutor::execute` only to return `Config("...spec-only handler cannot execute")` — executor variants that structurally can't execute, i.e. an invalid state the type lets you construct. If spec generation (`normalize`) were separable from execution, the enum wouldn't carry two arms that always error at runtime. Non-blocking; worth revisiting when the MCP surface settles.

crates/agentic-core/src/executor/engine.rs
```diff
@@ -63,6 +64,30 @@ fn error_sse_chunk(message: &str) -> String {
     format!("data: {event_json}\n\n")
 }
 
+async fn build_gateway_registry(tools: &[ResponsesTool], exec_ctx: &ExecutionContext) -> ToolRegistry {
```
ashwing: Heads-up: `build_gateway_registry` + the `run_until_gateway_tools_complete` changes here overlap #83, which refactors the same function into a `LoopDecision` match + per-call timeout and edits the same `gateway_public_output`/`gateway_event_plans` arms. Complementary (MCP output shaping vs loop control) but textually conflicting, so whoever lands second rebases. No strong preference on order — flagging so we sequence it rather than collide. If #83 lands first the MCP `execute` path should slot into `registry.dispatch()` with little change; happy to rebase #83 the other way if you'd rather land this first.

maralbahari: @ashwing thank you ashwin for the heads. I think on this path there are many changes specially by https://github.com/vllm-project/agentic-api/pull/84. I think we should first merge https://github.com/vllm-project/agentic-api/pull/84 and followed by https://github.com/vllm-project/agentic-api/pull/83 since you are more familiar with the design doc we already discussed before to implement both execution on gateway which is `web_search` and with codex namespace tool being executed on codex acting as client side would be complete cases of both scenarios. then #89 cause for MCP it would be simple to add.  let me know your thoughts

crates/agentic-core/src/tool/mcp/client.rs
maralbahari: added the connection timeout.

crates/agentic-core/src/tool/mcp/registry.rs
maralbahari: I have modified this file. `with_entries` is not required anymore

crates/agentic-server-core/src/tool/mcp/pool.rs
maralbahari: `from_config`  is currently only reached via from_params, which validates every entry through server_entry_from_param before constructing the `McpServerEntry`  it passes in. Any other caller that wants to use `from_config` directly would first have to construct `McpServerEntry` values themselves  and at that point, responsibility for SSRF-safety shifts to whatever code builds those entries.

crates/agentic-server-core/src/tool/mcp/handler.rs
maralbahari: since we are using the `ToolHandler` interface for normalize, validate and execution (implemented via GatewayExecutor that implements ToolHandler) . with use of handlers we are normalizing and building the entries to registry and we use them to execute. in case of MCP since we are required to have pool/client connection during normalization we do not need to build that connection the spec kind of Mcp handlers are meant to assist and avoid that connection so that only during registry build we create those connections. I added documentation on top this class to clarify. I think we can revisit the design in future  perhaps to cater special cases like MCP. but as off now we are using tool handlers as means of communication between normalize and build.

crates/agentic-server-core/tests/mcp_tool_test.rs
maralbahari: @ashwin added unhappy paths scenarios.

maralbahari:
@franciscojavierarceo @ashwing ready to review again. now codex `read_resources_mcp` is mapped to `agentic-api` gateway as gateway mcp execution as well.

crates/agentic-server-core/src/tool/mcp/handler.rs
```diff
@@ -0,0 +1,424 @@
...
+
+    #[must_use]
+    pub fn tool_call(client: Arc<McpClient>) -> Self {
+        Self::with_kind(McpHandlerKind::ToolCall {
+            spec: McpSpec::Tool,
+            client: Some(client),
+        })
+    }
+
+    pub async fn discovered_tool_handlers(&self, factory: &McpHandlerFactory) -> Vec<McpDiscoveredHandler> {
```
ashwing, line 202: Nothing calls this on the branch — it's the discovery pass (`tools/list` → the `{server_label}__{tool_name}` handlers), and `list_tools` / the discovered `tools/call` only run through here. Looks like it wants to run inside `build_with_handlers` in the `Mcp` arm and drop its results into `entries` next to `insert_mcp_entry` — build's already async and each handler already carries the exposed name + executor, so it'd slot in without touching dispatch. Right now discovered `tools/call` just can't be reached. Wire it there + a test, or leave it for the follow-up?

crates/agentic-server-core/src/tool/mcp/client.rs
```diff
@@ -0,0 +1,279 @@
...
+    tool_timeout: Duration,
+}
+
+impl McpClient {
+    /// Connects to an MCP server over streamable HTTP.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`McpError::Connect`] if the MCP initialization handshake fails.
+    pub async fn connect(server_url: &str, headers: Option<HashMap<String, String>>) -> Result<Self, McpError> {
```
ashwing, line 99: `validate_request_server_url` only checks the hostname string — it never resolves it — and then `connect` hands the host to the transport, which resolves and connects on its own. So the IP we actually connect to is never checked against the allowlist. Loopback-only default is fine, but the moment someone puts a hostname in `AGENTIC_MCP_ALLOWED_HOSTS` you've got a rebinding window: the name passes validation, then resolves to whatever at connect time. Either pin the resolved IP between validate and connect, or just document that allowlisted hosts have to be trusted.

ashwing (commented):
Took another pass after the rebase — collision warning, the LoopDecision loop, and the Codex `read_mcp_resource` bridge all look good now. `read_mcp_resource` works end to end.

One scope thing before this goes in, and a security note inline.

The `tools/call` discovery path (`discovered_tool_handlers` → `list_tools` → `tools/call`) is all there but nothing calls it, so discovered-tool calling doesn't actually work end to end yet. If this PR is just `read_mcp_resource` for now (which matches the design doc's first slice), I'd move the discovery bits to the follow-up instead of landing them unwired. If you want them here, they just need to get called + a test. What's the plan?

(Side note, not a problem — stdio looks deliberately config-only, since `server_entry_from_param` only ever builds Http entries. So a request can't make the gateway spawn a process. Just calling it out so it's not lumped in with the above.)

crates/agentic-server-core/src/tool/mcp/handler.rs
maralbahari: @ashwing yes this is intentional for now cause there will be another follow up PR for this path support. the focus of this  PR is interfacing resource/tool spec from `rmcp` package with first implementation of `read_resources` which is type of resource spec from `rmcp` package

crates/agentic-server-core/src/tool/mcp/client.rs
maralbahari: I’ve documented this trust boundary in `docs/design/mcp-gateway-integration.md` under “HTTP server allowlist security,” with an additional comment next to `AGENTIC_MCP_ALLOWED_HOSTS` in `pool.rs`.

ashwing (approved):
Went through the responses — everything from the last two passes is addressed:

- Unhappy-path coverage is there now (unreachable / connection-refused / missing-resource cassettes + the failed-`read_mcp_resource` assertions).
- Registry collision is a `tracing::warn!` on duplicate name now, matching the function-handler behavior.
- DNS-rebinding on allowlisted hosts is documented as a trust boundary (design doc + the note by `AGENTIC_MCP_ALLOWED_HOSTS`) — reasonable given the operator opt-in, agreed it doesn't need IP pinning in this PR.

The discovered `tools/call` / stdio path being unwired is fine as a follow-up — read-resource-first is a sensible first slice, and you've got the interface in place for it.

`read_mcp_resource` works end to end and CI's green. LGTM.

crates/agentic-server-core/src/tool/mcp/pool.rs
```diff
@@ -0,0 +1,284 @@
...
+        _ => return Err("URL scheme must be http or https".to_owned()),
+    }
+
+    if !url.username().is_empty() || url.password().is_some() {
+        return Err("URL must not include credentials".to_owned());
+    }
+
+    let host = url.host_str().ok_or_else(|| "URL must include a host".to_owned())?;
+    if is_allowed_request_host(host) {
+        return Ok(value.to_owned());
```
franciscojavierarceo, line 182: Could we disable redirects for this transport, or validate each redirect target against the same allowlist? This only checks the initial hostname, while the default reqwest client follows redirects, so an allowed MCP endpoint could redirect the gateway to an internal address that would have been rejected here. The allowlist should cover the address we actually connect to.

franciscojavierarceo (approved):
The MCP implementation and tests look good. I left one follow-up comment for redirect handling, which is being addressed separately.

PR #90, opened 2026-07-08
docs: add project roadmap

franciscojavierarceo:
## Summary

- Add a root `ROADMAP.md` describing vLLM Agentic API's role as the Rust, vLLM core-optimized layer for stateful agentic APIs.
- Capture the planned sequence: Responses API hydration, Codex support through Responses, server-side tool execution, Messages, Interactions, and broader stateful agentic APIs.
- Clarify the ownership boundary between vLLM core (`vllm-project/vllm`) and vLLM Agentic API (`vllm-project/agentic-api`).

## Test Plan

- `git diff --check -- ROADMAP.md`


ROADMAP.md
```diff
@@ -0,0 +1,155 @@
...
+Codex depends on stateful continuation, tool-call preservation, and a broad set
+of agentic tool and response shapes. `agentic-api` should accept Codex-compatible
+Responses traffic, preserve the fields Codex needs, and execute or return tool
+calls according to the ownership model for each tool.
+
+Core work:
+
+- Preserve Codex-used tool declarations and response item shapes.
+- Support client-owned tools, gateway-owned tools, and provider-owned tools.
+- Preserve unknown tool and item shapes without executing them by default.
```
ashwing: `provider-owned` is listed here but not spelled out like client/gateway-owned are below. Does provider-owned mean vLLM core runs the tool during inference (agentic-api never sees the call), or the gateway passes it through to some provider? 

Related: how does ownership get decided in the first place — does the client declare it per-tool in the request, or does the gateway resolve it by registry/type? Both are contract-level, probably worth a sentence each.

ROADMAP.md
```diff
@@ -0,0 +1,155 @@
...
+- Support model aliasing and routing needed for local vLLM core-backed Codex
```
ashwing: Is "don't execute unknown shapes by default" the blanket rule — anything not recognized as gateway-owned gets returned to the client, never run? Reads like the default posture that bounds what the server will ever execute, so might be worth stating as such rather than as one bullet.

ROADMAP.md
```diff
@@ -0,0 +1,155 @@
...
+  usage.
```
ashwing: Model aliasing/routing reads like a different concern from tool ownership — it's name-mapping, not execution. Is that agentic-api's job or the gateway's (Praxis)? Affects whether the alias map lives here or at the edge.

ROADMAP.md
```diff
@@ -0,0 +1,155 @@
...
+Interactions should provide the higher-level agentic workflow surface once the
+Responses and Messages foundations are solid. This work should reuse the same
+state, orchestration, and tool execution model rather than creating a parallel
+stack.
+
+Core work:
+
+- Define the Interactions API contract.
+- Reuse the shared execution loop for model calls and tool calls.
+- Support durable interaction state.
```
ashwing, line 118: Loop reuse comes up for both Messages and Interactions. Is the shared loop meant to be API-agnostic — operating on items + tool calls rather than Responses-specific shapes? That's what decides whether they genuinely share it or fork a parallel stack.

ashwing (commented):
A few scope/boundary questions inline — nothing blocking.

maralbahari (approved):
@franciscojavierarceo  clear and on point for me no further comment.

PR #91, opened 2026-07-09
fix: tighten codex namespace invariants

franciscojavierarceo:
## Summary

- reject Codex namespace declarations whose generated flat names collide with a declared top-level function or another namespace member before building the upstream request
- make namespace member resolution, registry construction, and `RequestPayload::to_upstream_request` fallible so direct callers cannot bypass collision validation
- validate named `tool_choice` values with the existing `NonEmptyToolName` type so empty function choices are rejected at the serde boundary
- refresh the Codex integration design doc to describe the current typed executor namespace flatten/restore path and collision behavior

## Test Plan

- `cargo fmt -- --check`
- `cargo test -p agentic-core --test stateful_responses_integration test_codex_namespace_collision_with_top_level_function_is_rejected --quiet`
- `cargo test -p agentic-core tool::codex::tests --quiet`
- `cargo test -p agentic-core --test tool_normalization_test --quiet`
- `cargo test -p agentic-core codex --quiet`
- `cargo clippy -p agentic-core --tests -- -D warnings`
- `git diff --check`
- `cargo test -p agentic-core --quiet`

### DIFF
--- a/crates/agentic-core/src/types/request_response.rs
+++ b/crates/agentic-core/src/types/request_response.rs
@@ -5,7 +5,7 @@ use super::io::{
     FunctionTool, InputItem, InputMessage, InputMessageContent, OutputItem, ResponseUsage, ResponsesInput, ToolChoice,
 };
 use super::tools::ResponsesTool;
-use crate::tool::CodexNamespaceHandler;
+use crate::tool::{CodexNamespaceHandler, ToolError};
 use crate::utils::common::serialize_to_string;
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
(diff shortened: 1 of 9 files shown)


crates/agentic-core/src/types/request_response.rs
```diff
@@ -73,6 +73,10 @@ fn is_absent_or_default_tool_choice(choice: &Option<ToolChoice>) -> bool {
 impl RequestPayload {
     /// Construct an `UpstreamRequest` suitable for forwarding to vLLM.
     ///
+    /// Executor callers validate namespace collisions before serializing the
```
ashwing: Wired paths all validate first (engine.rs:110), so we're good today. One nagging thought though: `to_upstream_request` is `pub` and calls `resolve` directly — the "validate first" rule only holds via a doc-comment + a `debug_assert` that vanishes in release. A future caller coming in this way silently gets last-write-wins again. Might want `resolve_namespace_members` to just return a `Result` so it's type-enforced. Non-blocking.

ashwing (approved):
This closes the #84 collision case nicely — reject-before-build covers both the top-level and sibling-member clashes, and `tool_choice` now goes through `NonEmptyToolName`. Approving. One non-blocking thought inline.

PR #92, opened 2026-07-09
chore: rename core crate for release

franciscojavierarceo:
## Summary
- rename the publishable core package from `agentic-core` to `agentic-server-core` while preserving the Rust library target name `agentic_core`
- keep the workspace release version at `0.1.0` and point the server dependency at `agentic-server-core = 0.1.0`
- mark `agentic-praxis` as unpublished and update docs references to the new core crate name
- exclude local DB files, tests, and benches from the published core tarball

## Test Plan
- `cargo fmt -- --check`
- `cargo test`
- `cargo publish --dry-run -p agentic-server-core --allow-dirty`
- `cargo package -p agentic-server-core --allow-dirty --list`

Note: `agentic-server` packaging/publish dry-run cannot complete until `agentic-server-core v0.1.0` exists in the crates.io index; Cargo fails dependency resolution with `no matching package named agentic-server-core found`, which is expected before publishing the core crate first.


Cargo.toml
```diff
@@ -16,7 +16,7 @@ all = { level = "deny", priority = -1 }
 pedantic = { level = "warn", priority = -1 }
 
 [workspace.dependencies]
-agentic-core = { path = "crates/agentic-core" }
+agentic-core = { package = "agentic-server-core", path = "crates/agentic-server-core", version = "0.1.0" }
```
ashwing, line 19: The description says 0.0.1 (workspace version + this pin), but it's 0.1.0 here and the workspace `version` at line 6 is unchanged too — the diff doesn't bump anything. Which one's intended for the release? Worth pinning down since crates.io versions are immutable once published.

Minor while you're here: the packaged tarball still ships `tests/` and `benches/` sources but excludes their cassette fixtures, so `cargo test` from the published crate can't find them. Harmless for consumers, but excluding `tests/**` + `benches/**` too would slim it if you want.

ashwing (commented):
One question on the version before this goes out for a release, inline.

Cargo.toml
franciscojavierarceo: whoops this was a miss, fixed.

ashwing (approved):

maralbahari (approved):
LGTM

maralbahari:
@franciscojavierarceo @ashwing let's merge the rename after https://github.com/vllm-project/agentic-api/pull/83?
I added a comment for this. once addressed we can include this refactor with the release

PR #95, opened 2026-07-11
fix: tune SQLite concurrency and add mixed read/write coverage

jiahuei:
This PR improves SQLite's concurrency behavior so the default local storage path is more robust under realistic read/write workloads.

Even if production deployments eventually standardize on PostgreSQL, this work keeps the storage layer disciplined around short transactions, concurrent access, and deterministic sequence allocation, which are useful invariants for production databases too. SQLite should be as reliable as practical for development, testing, and lightweight deployments while preserving a clear path to PostgreSQL.

## Summary

- Enable SQLite WAL mode for writable file-backed databases and configure connection pragmas for concurrency:
  - `busy_timeout`
  - `foreign_keys`
  - `synchronous=NORMAL`
  - `journal_size_limit`
  - `temp_store`
  - `mmap_size`
- Keep SQLite as the default storage backend and use a small multi-connection pool for file-backed databases.
- Keep in-memory SQLite single-connection to avoid splitting schema/data across private in-memory connections.
- Shorten write transaction windows by moving serialization before `pool.begin()`.
- Compute conversation item sequence ranges inside the insert statement, avoiding a separate pre-write `SELECT MAX(seq)`.
- Add env-only SQLite tuning via:
  - `SQLITE_MAX_CONNECTIONS`, default `4` for file-backed SQLite; in-memory SQLite always uses `1`
  - `SQLITE_JOURNAL_SIZE_LIMIT_BYTES`, default `6144000`
  - `SQLITE_TEMP_STORE`, default `memory`
  - `SQLITE_MMAP_SIZE_BYTES`, default `268435456`
- Ignore SQLite WAL sidecar files in `.gitignore`.

## Test Plan

- Added storage regression coverage with `test_sqlite_multi_pool_mixed_read_write_concurrency`, which opens independent SQLite pools to the same file, runs concurrent readers and writers, asserts no lock errors, and verifies final item count plus gap-free sequence numbers.
- Added same-pool mixed read/write regression coverage with `test_sqlite_same_pool_mixed_read_write_concurrency`, which verifies file-backed SQLite pools can use multiple connections safely.
- Added WAL startup regression coverage with `test_sqlite_wal_preflight_retries_locked_database`, which verifies WAL initialization retries after a transient SQLite locked-database error.
- Added SQLite pool sizing tests for file-backed pool size configuration and the in-memory single-connection override.
- Added SQLite pragma tests for WAL, busy timeout, foreign keys, synchronous mode, journal size limit, temp store, mmap size, and explicit custom tuning values.
- Added server config tests that verify SQLite tuning is env-only, not exposed as CLI flags, and that env parser defaults/invalid values behave correctly.
- Added benchmark coverage with `concurrent_conversation_mixed_read_write`, measuring one writer plus N concurrent readers across independent SQLite pools.

Validated with:

```bash
cargo fmt -- --check
cargo test --locked -p agentic-server-core sqlite_max_connections -- --nocapture
cargo test --locked -p agentic-server-core sqlite_memory_pool_stays_single_connection -- --nocapture
cargo test --locked -p agentic-server-core test_sqlite_multi_pool_mixed_read_write_concurrency -- --nocapture
cargo test --locked -p agentic-server-core test_sqlite_same_pool_mixed_read_write_concurrency -- --nocapture
cargo test --locked -p agentic-server-core test_sqlite_wal_preflight_retries_locked_database -- --nocapture
cargo test --locked -p agentic-server-core storage::pool::tests::test_sqlite_connection_pragmas -- --nocapture
cargo test --locked -p agentic-server sqlite_tuning -- --nocapture
cargo test --locked -p agentic-server-core --bench benches --no-run
cargo clippy --locked --all-targets -- -D warnings
cargo test --locked
```

### DIFF
--- a/crates/agentic-server-core/src/storage/pool.rs
+++ b/crates/agentic-server-core/src/storage/pool.rs
@@ -2,7 +2,14 @@
 
 use std::sync::Arc;
 
+use sqlx::AnyConnection;
 use sqlx::any::AnyPoolOptions;
+
+use crate::config::SqliteConfig;
+
+const SQLITE_MEMORY_MAX_CONNECTIONS: u32 = 1;
+const DEFAULT_MAX_CONNECTIONS: u32 = 10;
+const SQLITE_BUSY_TIMEOUT_MS: u64 = 5_000;
 
 /// Generic database pool type supporting `SQLite`, `PostgreSQL`, and `MySQL`.
 pub type DbPool = sqlx::Pool<sqlx::Any>;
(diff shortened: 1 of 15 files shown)


crates/agentic-server-core/src/storage/pool.rs
```diff
@@ -54,20 +127,37 @@ fn prepare_db_url(url: Option<&str>) -> String {
...
+    } else {
+        DEFAULT_MAX_CONNECTIONS
+    };
+    let mut options = AnyPoolOptions::new().max_connections(max_connections);
+    if url.starts_with("sqlite") {
+        options = options.after_connect(move |conn, _meta| Box::pin(configure_sqlite_connection(conn, sqlite_config)));
+    }
+    let pool = options.connect(&url).await?;
+    if sqlite_should_enable_wal(&url) {
+        sqlx::query("PRAGMA journal_mode = WAL").execute(&pool).await?;
```
franciscojavierarceo: [P1] could we make WAL initialization retryable/serialized? `PRAGMA journal_mode = WAL` needs an exclusive lock, and SQLite does not honor `busy_timeout` while changing journal mode. If another gateway process or an in-flight write transaction is using this DB during startup, this error bubbles up and the server fails to start. I reproduced `database is locked` with `BEGIN EXCLUSIVE` held. Please move this into a one-time retryable preflight (or require WAL to be preconfigured) and add a regression test for startup with an active transaction.

jiahuei: Changed:

- Added bounded WAL preflight retry: 5 total attempts, 1s delay, only for SQLite `SQLITE_BUSY` / `SQLITE_LOCKED` primary error codes.
- Replaced the one-shot `PRAGMA journal_mode = WAL` with the retry helper.
- Added regression coverage for a locked DB during WAL preflight.

We still expect SQLite deployments here to be single-process and restarted stop-before-start, and production is far more likely to use another DB like Postgres, but this makes the SQLite startup path more robust under transient contention.

Verified:

```
cargo fmt -- --check
cargo check --locked -p agentic-server-core
cargo test --locked -p agentic-server-core storage::pool::tests -- --nocapture
```

franciscojavierarceo (commented):
minor follow-up: the PR description still uses `-p agentic-core`, but the Cargo package name here is `agentic-server-core`, so those commands fail before running tests. Could you update the validation commands and add the new `test_sqlite_wal_preflight_retries_locked_database` coverage to the listed test plan?

jiahuei:
> minor follow-up: the PR description still uses `-p agentic-core`, but the Cargo package name here is `agentic-server-core`, so those commands fail before running tests. Could you update the validation commands and add the new `test_sqlite_wal_preflight_retries_locked_database` coverage to the listed test plan?

Sure, its updated

franciscojavierarceo (approved):

PR #97, opened 2026-07-13
ci: add manual crates release workflow

franciscojavierarceo:
## Summary
- add a manual `workflow_dispatch` workflow for crates.io releases
- default the button to dry-run mode and require an explicit toggle for real publishing
- gate the workflow to `main` and require the actor to have `admin` or `maintain` repo permission
- publish in the proven order: `agentic-server-core` first, then `agentic-server` after the crates.io index sees core
- follow the vLLM release-ref pattern by creating `releases/vX.Y.Z`, tagging `vX.Y.Z`, and generating GitHub Release notes from the published commit
- scope `CARGO_REGISTRY_TOKEN` only to the actual publish steps

## Test Plan
- `SKIP=no-commit-to-branch uvx pre-commit run --files .github/workflows/release-crates.yml`
- `git diff --check -- .github/workflows/release-crates.yml`

### DIFF
--- a/.github/workflows/release-crates.yml
+++ b/.github/workflows/release-crates.yml
@@ -1,0 +1,149 @@
+name: Release crates
+
+run-name: Release crates (${{ inputs.dry_run && 'dry run' || 'publish' }})
+
+on:
+  workflow_dispatch:
+    inputs:
+      dry_run:
+        description: "Run packaging and publish checks without uploading crates"
+        required: true
+        type: boolean
+        default: true
+
+concurrency:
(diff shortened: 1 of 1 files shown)


tjtanaa (approved):
LGTM for now, as our PR merge interval is pretty spread out. 

We will have to consider allowing maintainer to specify a specific commit to avoid racing condition of PR merge and version creation. Another thing, as we progress, we might want to exercise what vLLM does where we will freeze a version branch and add necessary patches to it before we release an official version.

PR #99, opened 2026-07-14
feat: add Anthropic Messages gateway proxy

franciscojavierarceo:
## Summary

- add transparent Anthropic Messages proxy routes for `/v1/messages` and `/v1/messages/count_tokens`
- preserve raw request bodies, query parameters, Anthropic/Claude Code headers, upstream statuses, JSON responses, SSE streams, and error bodies
- keep the existing Responses proxy API compatible while adding a path-aware internal proxy helper
- add recording-upstream integration coverage using Claude Code-style client-owned `WebSearch`/`WebFetch` tools and arbitrary request fields
- cover the protocol-contract requirements tracked in issue #114: verbatim block-first `system` forwarding, open-list `anthropic-beta` forwarding, and unchanged `anthropic-version`

## Test Plan

- `cargo fmt -- --check`
- `cargo test -p agentic-server --test messages_test`
- `cargo test -p agentic-server`
- `git diff --check`

PR #98 provides the design and live-spike evidence for the follow-up Phase 2 server-side tool loop: gateway tools must be injected only into upstream vLLM requests, executed inside Agentic API, and hidden from Claude Code. This PR intentionally remains the transparent proxy floor; no gateway-owned web-search execution is included.

### DIFF
--- a/crates/agentic-server/src/handler/http/messages.rs
+++ b/crates/agentic-server/src/handler/http/messages.rs
@@ -4,12 +4,12 @@ use bytes::Bytes;
 
 use agentic_core::proxy::{ProxyAuth, ProxyRequest, proxy_request_with_path};
 
-use super::super::common::{convert_response, read_bytes};
+use super::super::common::{convert_response, read_bytes_with_auth};
 use crate::app::AppState;
 
 async fn proxy_messages(state: &AppState, request: Request, path: &'static str) -> Response {
     let (parts, body) = request.into_parts();
-    let body: Bytes = match read_bytes(body).await {
+    let body: Bytes = match read_bytes_with_auth(body, ProxyAuth::Anthropic).await {
         Ok(body) => body,
         Err(response) => return response,
     };
--- a/crates/agentic-server/src/handler/http/messages.rs
+++ b/crates/agentic-server/src/handler/http/messages.rs
@@ -1,0 +1,38 @@
+use axum::extract::{Request, State};
+use axum::response::Response;
+use bytes::Bytes;
+
+use agentic_core::proxy::{ProxyAuth, ProxyRequest, proxy_request_with_path};
+
+use super::super::common::{convert_response, read_bytes};
+use crate::app::AppState;
+
+async fn proxy_messages(state: &AppState, request: Request, path: &'static str) -> Response {
+    let (parts, body) = request.into_parts();
+    let body: Bytes = match read_bytes(body).await {
+        Ok(body) => body,
+        Err(response) => return response,
+    };
+
+    convert_response(
+        proxy_request_with_path(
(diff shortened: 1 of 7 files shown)


crates/agentic-server/src/handler/http/messages.rs
```diff
@@ -0,0 +1,38 @@
...
+
+    convert_response(
+        proxy_request_with_path(
+            ProxyRequest {
+                headers: parts.headers,
+                body,
+                query: parts.uri.query().map(str::to_owned),
+            },
+            path,
+            ProxyAuth::Anthropic,
```
ashwing, line 25: On this Anthropic path, upstream errors pass straight through (good — `messages_preserves_upstream_error_status_and_body` covers it). But our own failures here — timeout → 504, unreachable upstream → 502, the 413 from `read_bytes` above — all come back through `error_response` in proxy.rs, which is OpenAI-shaped: `{"error":{"type":"api_error","param":null,"code":...}}`. That's the wrong shape for a Claude Code client — it can't read a 502/504 from us, and `param`/`code` don't belong in an Anthropic response.

Since this PR is what puts an Anthropic endpoint in front of `error_response`, I think it should return an Anthropic-shaped error too — `{"type":"error","error":{"type":"api_error"|"overloaded_error","message":...}}`. Threading the `ProxyAuth` you're passing here into `error_response` (or a small `anthropic_error_response`) would do it. Reasoning's in the #98 error-forwarding section.

crates/agentic-server/tests/messages_test.rs
```diff
@@ -0,0 +1,197 @@
...
+        .await
+        .unwrap();
+
+    assert_eq!(response.status(), StatusCode::OK);
+    assert_eq!(response.text().await.unwrap(), r#"{"input_tokens":3}"#);
+    assert_eq!(requests.lock().await[0].uri, "/v1/messages/count_tokens");
+}
+
+#[tokio::test]
+async fn messages_preserves_upstream_error_status_and_body() {
```
ashwing, line 176: This nails upstream-error passthrough. The mirror case for our own errors isn't here — point the gateway at a dead upstream and assert the body shape. That's the test that would catch the OpenAI envelope from the messages.rs comment, so worth adding alongside that fix.

ashwing (commented):
The proxy path itself looks right. Only thing I'd change before merge is the error shape on the Anthropic route — left it inline.

ashwing (approved):
The `error_response_for_auth` split covers it — Anthropic path returns the `{"type":"error","error":{...}}` shape now, across the timeout/transport/body-read/413 sites, and `read_bytes` keeps the OpenAI default for the Responses path. The dead-upstream test pins the 502 body shape too. CI's green. LGTM.

PR #101, opened 2026-07-14
feat: support codex custom tool type

haoshan98:
Codex declares freeform tools such as `apply_patch` with `type: "custom"`. Unlike function tools, these declarations can
include an opaque grammar, and their calls carry raw text in `input` rather than JSON in `arguments`. The typed Responses
path previously could not preserve that protocol through upstream inference, streaming, storage, and continuation.

- Add typed request, response, and input models for custom tool declarations, `custom_tool_call` items, and `custom_tool_call_output` results while preserving unknown format fields.
- Forward custom tool declarations to the upstream Responses endpoint without converting them to function tools.
- Accept and forward named custom `tool_choice` objects such as
  `{"type":"custom","name":"apply_patch"}` alongside existing function choices.
- Normalize and accumulate custom-tool streaming input events, return custom calls to the client for execution, and retain the call/output pair during stateful continuation.
- Keep custom calls client-owned in the shared tool loop while gateway-owned MCP and web-search calls continue to execute through their registered handlers.
- Persist terminal WebSocket responses before emitting completion so Codex can immediately continue a client-owned tool call without losing stored state when it closes the socket.
- Add custom-context coverage for the existing local WebSocket `generate: false` prewarm path, confirming that prompt
  and custom-tool declarations are preserved without upstream inference.
- Preserve and log structured upstream streaming failure details to make rejected tool requests diagnosable.
- Add an isolated catalog/cache seed helper that writes both Codex catalog files so freeform `apply_patch` remains
  available after the ordinary model cache expires, with canonical-path checks that keep writes inside `/tmp`.
- Extend the cassette recorder with custom-call continuation support and add direct-vLLM HTTP, gateway HTTP/SSE,
  gateway WebSocket, OpenAI HTTPS/SSE, and OpenAI WebSocket recordings that preserve raw input and
  `custom_tool_call_output`.
- Document direct gateway, `codex exec`, and interactive Codex verification workflows.


crates/agentic-server-core/src/types/io/output.rs
```diff
@@ -138,6 +155,31 @@ impl TryFrom<&EventPayload> for FunctionToolCall {
...
+            return Err(ExecutorError::ParseError("expected OutputItemAdded payload".into()));
+        };
+        let id = if item_id.is_empty() {
+            uuid7_str("ctc_")
+        } else {
+            item_id.clone()
+        };
+        Ok(Self {
+            id,
+            status: Some("in_progress".to_owned()),
```
maralbahari: you can use `MessageStatus` enum.

crates/agentic-server-core/src/types/request_response.rs
```diff
@@ -41,12 +39,12 @@ pub struct UpstreamRequest<'a> {
...
-    /// Normalised tools forwarded to vLLM — always `Vec<FunctionTool>` regardless of
-    /// what tool types the client declared. Codex namespace tools are flattened
-    /// before this is built.
+    /// Tools forwarded to vLLM. Namespace members are flattened to ordinary
+    /// function declarations; native custom declarations retain their freeform
+    /// wire shape.
     /// Skipped when empty so vLLM does not receive an empty array.
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub tools: Option<Vec<FunctionTool>>,
+    pub tools: Option<Vec<Value>>,
```
maralbahari: this type change from `FunctionTool` to `Value` seems like messing up most of the current tool contract. revisit this I think even the free form should be normalized to FunctionTool type

crates/agentic-server-core/src/executor/accumulator.rs
```diff
@@ -66,6 +68,15 @@ impl InFlight {
                 item.status = MessageStatus::Completed;
                 output.push(OutputItem::Message(item));
             }
+            Self::CustomToolCall { mut item, input } => {
+                if item.input.is_empty() {
+                    item.input = input;
+                }
+                if item.status.as_deref().is_none_or(|status| status == "in_progress") {
+                    item.status = Some("completed".to_owned());
```
maralbahari: instead of using the hardcoded string for status. you can use the `MessageStatus` enum. see the above inflight items how they use it.

crates/agentic-server-core/src/types/io/output.rs
haoshan98: Ok

crates/agentic-server-core/src/executor/accumulator.rs
haoshan98: Ok

crates/agentic-server-core/src/types/request_response.rs
haoshan98: Replace `Vec<Value>` with a typed heterogeneous `UpstreamTool` enum instead, to contain both function and native-custom variants

crates/agentic-server-core/src/tool/normalize.rs
```diff
@@ -53,6 +57,10 @@ impl ResponsesTool {
                 |param| CodexNamespaceHandler.normalize(&param),
                 vec![],
             ),
+            Self::Custom(p) => {
+                tracing::debug!(name = %p.name, "custom tool retained for native upstream forwarding");
```
maralbahari, line 61: @haoshan98 can just add the normalization logic for custom params here if custom params suppose to be converted to function tool as well. then, should not need separate matching logic in ResponsesPayload.to_upstream_tools() to check custom type and normalize. can simply call ResponsesTool.to_function_tools, meaning putting ResponsesPayload.to_upstream_tools()  logic untouched. however if custom param would not be converted into function tool type then can ignore this check entirely?

haoshan98: A `custom` tool uses raw-text input rather than JSON function arguments, so it cannot be converted into a function tool. The current upstream path handles `Custom` before calling `to_function_tools()`, so this arm is not executed there. However, `to_function_tools(&self)` accepts the full `ResponsesTool` enum, and Rust requires its match to cover every variant.

maralbahari (approved):
@haoshan98 Thanks LGTM

PR #117, opened 2026-07-14
fix: disable redirects for MCP HTTP clients

franciscojavierarceo:
## Summary

Disable HTTP redirects in the RMCP client used for request-provided MCP servers. The existing hostname allowlist validates only the initial URL; following redirects could otherwise reach an unallowlisted internal address.

Adds a regression test proving the client returns the redirect response instead of following it.

## Test Plan

- `cargo fmt -- --check`
- `cargo test -p agentic-server-core http_client_does_not_follow_redirects -- --nocapture`
- `cargo test -p agentic-server-core --test mcp_tool_test -- --nocapture`
- `cargo clippy -p agentic-server-core --all-targets -- -D warnings`
- `git diff --check`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -48,6 +48,7 @@ dependencies = [
  "http",
  "indexmap",
  "reqwest 0.12.28",
+ "reqwest 0.13.4",
  "rmcp",
  "serde",
  "serde_json",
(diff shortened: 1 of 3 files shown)


ashwing (approved):
LGTM.

- Redirects are disabled at `connect()`, which is the only HTTP path — closes the #89 redirect gap where it could leak.
- `connect_stdio` is stdio, so no redirect surface there.
- Test pins it: asserts the 307 returns rather than being followed.
- `reqwest-mcp` 0.13 next to workspace 0.12 looks odd, but it has to match rmcp's transport major, and 0.13 + rustls are already in the tree via rmcp.

Issue #115, opened 2026-07-14
[Claude Code - Stage 2] Server-side gateway tool loop for Messages


ashwing:
Part of #113 (Claude Code integration). **This is the core value-add.**

When the model emits a **gateway-owned** `tool_use` (`web_search`, MCP), the gateway must execute it server-side and continue the loop — never handing it back to Claude Code. The spike proved this is the *only* mechanism that works: Claude Code errors on any tool it didn't declare, so gateway tools must stay invisible to the client (see #98 → Verification).

Scope:
- [ ] Map Anthropic `tool_use`/`tool_result` content blocks to the shared tool framework's gateway-owned vs. client-owned classification.
- [ ] Run the hide-the-call loop (`run_until_gateway_tools_complete`, landed in #83) on the Messages path: inject the gateway tool into the upstream call only, execute it, append the `tool_result`, loop, surface only the final assistant message.
- [ ] Emit correct Anthropic SSE for surfaced calls and suppress hidden gateway calls (reuses the existing `emit_gateway_*_events` logic on `main`; keep block `index` contiguous).
- [ ] Handle parallel `tool_use`, the full `stop_reason` set (incl. `pause_turn`), and streaming `input_json_delta` args.
- [ ] **Record the acceptance cassette** — the upstream Messages turns (turn 0: model emits the gateway `tool_use`; turn 1: final text after the `tool_result`) so the loop is covered hermetically in CI, not only live. This records the *upstream* side, so it reuses the #94 recorder pattern and doesn't depend on the loop existing — it can start immediately. May need a small recorder change for the Messages response shape.
- [ ] Acceptance test: a real gateway-owned `web_search` turn completes end-to-end through `/v1/messages` with the tool hidden from the client, replayed from that cassette.

**Dependency:** the shared loop is currently typed to `RequestPayload`/`ResponsePayload`; Stage 2's clean form needs it factored to an API-agnostic shape. That factoring is owned by the pending Layering ADR (tool-framework.md Future Work), so this should sequence after / alongside that decision, not reshape the core unilaterally. (The cassette recording sub-task above has no such dependency and can begin right away.)

Happy to take this one.

ashwing:
Closing — the core value-add landed in #131 (server-side gateway tool loop on `/v1/messages`) with the acceptance cassettes in #123.

Against the original scope:
- ✅ `tool_use`/`tool_result` gateway-vs-client classification — `tool_seam.rs`
[...]

PR #123, opened 2026-07-15
test: add Anthropic Messages recording mode + gateway-tool cassettes

ashwing:
## Summary

Part of #115 (Claude Code — Stage 2). Adds a `messages` mode to the cassette recorder for the Anthropic Messages API (`/v1/messages`), and records the upstream gateway-tool traffic that the Claude Code server-side tool loop will replay in tests.

The Stage 2 server-side loop itself is gated on making the executor loop API-agnostic (it's currently typed to Responses `RequestPayload`/`ResponsePayload`) — that's a core refactor for a separate change. Recording the acceptance cassette has **no such dependency** and is the unblocked prerequisite #115 calls out, so this lands the recorder + fixtures ahead of the loop.

What's here:
- **`messages` recorder mode** (`record_cassette.py`): sends Anthropic `/v1/messages` with tools, stateless (resends full `messages` history each turn). With `--tool-outputs`, a turn following a `tool_use` feeds back matching `tool_result` blocks keyed by tool name — mirroring how a gateway tool loop feeds results back. Streaming reconstructs the final message from the Anthropic SSE (`content_block_delta` / `input_json_delta` / `message_delta`).
- **Recorded cassettes** (`messages/`): a two-turn `web_search` gateway-tool session — model emits the gateway `tool_use`, `tool_result` is fed back, model produces the final answer. Both streaming and non-streaming.
- **Reproducer** `record_messages_cassettes.sh` + a README modes-table row.

## Test Plan

- Recorded against **vLLM 0.25.1** (the current release, which serves `/v1/messages` natively), `Qwen/Qwen3-30B-A3B-FP8`, hermes tool parser + qwen3 reasoning parser.
- `python -m py_compile record_cassette.py` clean; `--help` shows the new mode; `cargo build --workspace` unaffected (no Rust changed).
- Recorded cassettes verified: 2 turns each, `POST /v1/messages`, turn 1 emits `tool_use(web_search)` with `stop_reason: tool_use`, turn 2 request carries the full Anthropic history (`user` → `assistant` with `tool_use` → `user` with `tool_result`), turn 2 produces the final `end_turn` answer.
- **No Rust consumer yet:** the Messages cassette loader lands with the Stage 2 loop these fixtures are for. Recording them now (per #115's carved-out sub-task) de-risks that work and keeps the fixtures reproducible.
- Note: streaming reasoning arrives as `<think>`-tagged `text` deltas while non-streaming arrives as `thinking` blocks — this is an accurate capture of what vLLM emits on each path, recorded intentionally so the future parser handles both.

### DIFF
--- a/crates/agentic-server-core/tests/cassettes/README.md
+++ b/crates/agentic-server-core/tests/cassettes/README.md
@@ -30,7 +30,8 @@ The recorder scripts (`record_reasoning_cassettes.sh`, `record_tool_call_cassett
 
 | Mode | Description |
 |------|-------------|
-| `responses` | Chains turns via `previous_response_id`. Only mode supported with `--vllm`. Common mode for gateway-backed built-in tool cassettes. |
+| `responses` | Chains turns via `previous_response_id`. Supported with `--vllm`. Common mode for gateway-backed built-in tool cassettes. |
+| `messages` | Anthropic Messages API (`/v1/messages`). Stateless: resends the full `messages` history each turn. With `--tool-outputs`, a turn following a `tool_use` feeds back matching `tool_result` blocks (keyed by tool name) instead of prompting. Supported with `--vllm`. |
 | `conv` | Creates a conversation object, passes `conversation` id each turn. |
 | `isolation` | Two independent conversations (A and B) recorded into one cassette. |
 | `mixed` | Turn 1 uses `conversation` id, turns 2+ switch to `previous_response_id`. |
--- a/crates/agentic-server-core/tests/cassettes/messages/messages-web-search-Qwen-Qwen3-30B-A3B-FP8-nonstreaming.yaml
+++ b/crates/agentic-server-core/tests/cassettes/messages/messages-web-search-Qwen-Qwen3-30B-A3B-FP8-nonstreaming.yaml
@@ -1,0 +1,169 @@
+turns:
+- filename: t1
+  request:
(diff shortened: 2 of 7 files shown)


maralbahari (approved):
@ashwing thanks. just to check my understanding from about the flow. So this PR is only adding cassettes file recorded from vllm upstream `v1/messages` entrypoint then in a follow up PR would add in the request/response payload mapping on gateway for messages api?
then in another PR handling the SSE chunk mapping?

ashwing:
Yes, exactly. This PR is just the recorder `messages` mode + the recorded upstream `/v1/messages` cassettes — no gateway behavior. The follow-up (Stage 2, #115) adds the actual server-side path: Messages request/response payload mapping into the shared tool loop, plus the gateway-owned tool execution that replays against these cassettes.

SSE/streaming chunk mapping I'd keep as its own piece rather than fold into #115 — the streaming cassette here already captures the Anthropic `content_block_delta` / `input_json_delta` frames, so it can be tested independently. So roughly: (1) this — fixtures; (2) #115 — request/response mapping + server-side tool loop; (3) streaming event mapping. The exact split of (2)/(3) depends on how the loop's API-agnostic factoring lands, which is still open.

ashwing:
@maralbahari @franciscojavierarceo can we merge this to unblock the loop changes?

PR #128, opened 2026-07-16
fix: preserve parallel tool call setting

### LINKED ISSUE #127: Preserve parallel_tool_calls in Responses gateway requests
## Summary

`/v1/responses` requests can include `parallel_tool_calls` to control whether the model may emit multiple tool calls in parallel. The gateway currently drops this field on the executor/tool-normalization path because it is not represented in `RequestPayload` or forwarded by `to_upstream_request`.

## Impact

Clients that send `"parallel_tool_calls": false` can still have the setting omitted from the upstream vLLM request after gateway normalization. That makes the gateway behavior differ from the client request and from transparent proxy expectations.

## Expected behavior

When a client sets `parallel_tool_calls`, the gateway should preserve the value and forward it to vLLM along with the normalized Responses request.

## Reproduction shape

Send a `/v1/responses` request through a path that uses gateway normalization, for example with a gateway-owned tool such as `web_search_preview`:

```json
{
  "model": "test",
  "input": [{"type": "message", "role": "user", "content": "hi"}],
  "tools": [{"type": "web_search_preview"}],
  "parallel_tool_calls": false,
  "store": false,
  "stream": false
}
```

The upstream request should contain `"parallel_tool_calls": false`.

harivilasp:
## Summary
- parse `parallel_tool_calls` on `/v1/responses` requests
- forward the field through `RequestPayload::to_upstream_request` after gateway tool normalization
- add unit and route-level regression coverage proving `parallel_tool_calls: false` reaches the mock vLLM request

Closes #127.

## Why
The gateway already preserves several Responses API generation/control fields (`include`, `temperature`, `top_p`, `max_output_tokens`, `truncation`, `metadata`) while normalizing tools for vLLM. `parallel_tool_calls` was missing from the typed request model, so clients that explicitly disabled parallel tool calls had that setting silently dropped on the executor path.

## Validation
- `cargo test -p agentic-server --test responses_test`
- `cargo test --workspace`
- `cargo clippy --workspace --all-targets -- -D warnings`
- `cargo fmt -- --check`

Focused tests were written first and failed before the implementation because `parallel_tool_calls` serialized as absent/null instead of `false`.

### DIFF
--- a/crates/agentic-server-core/src/types/request_response.rs
+++ b/crates/agentic-server-core/src/types/request_response.rs
@@ -88,6 +88,18 @@ impl RequestPayload {
     /// flat name collides with a top-level function tool or another namespace
     /// member.
     pub fn to_upstream_request(&self, stream: bool) -> Result<UpstreamRequest<'_>, ToolError> {
+        let has_gateway_owned_tool = self.declares_gateway_owned_tool();
+        if has_gateway_owned_tool && self.parallel_tool_calls == Some(true) {
+            return Err(ToolError::Config(
+                "parallel_tool_calls must be false when using built-in tools".into(),
+            ));
+        }
+        let parallel_tool_calls = if has_gateway_owned_tool {
+            Some(false)
+        } else {
+            self.parallel_tool_calls
+        };
+
         let renamed_tools = self
             .tools
             .as_deref()
@@ -112,7 +124,21 @@ impl RequestPayload {
             max_output_tokens: self.max_output_tokens,
             truncation: self.truncation.as_deref(),
             metadata: self.metadata.as_ref(),
-            parallel_tool_calls: self.parallel_tool_calls,
+            parallel_tool_calls,
+        })
+    }
+
+    fn declares_gateway_owned_tool(&self) -> bool {
+        self.tools.as_deref().is_some_and(|tools| {
+            tools.iter().any(|tool| {
+                matches!(
+                    tool,
+                    ResponsesTool::Mcp(_)
+                        | ResponsesTool::WebSearch(_)
+                        | ResponsesTool::FileSearch(_)
+                        | ResponsesTool::CodeInterpreter(_)
+                )
+            })
         })
     }
 }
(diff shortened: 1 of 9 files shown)


jiahuei:
We need to handle cases where built-in tools are allowed/specified by the user, in which case `parallel_tool_calls` needs to be validated such that only `false` is allowed. Otherwise it is impossible to handle cases where the model decides to call both built-in and custom tools in parallel.

[This is the behaviour of OpenAI](https://developers.openai.com/api/docs/guides/function-calling#parallel-function-calling)

harivilasp:
Addressed in 9844bc3. Built-in/gateway-owned tools now force upstream `parallel_tool_calls: false` when the client omits it, and explicitly reject `parallel_tool_calls: true` for built-in tools. Client function tools can still preserve `parallel_tool_calls: true`.

Verified with:
- `cargo fmt --check`
- `cargo test -p agentic-server-core types::request_response`
- `cargo test -p agentic-server-core --test web_search_tool_test`

maralbahari:
@harivilasp could you please resolves merge conflict, thanks.

harivilasp:
> @harivilasp could you please resolves merge conflict, thanks.
sure on it

harivilasp:
resolved merge confilcts

crates/agentic-server-core/src/types/request_response.rs
```diff
@@ -234,6 +263,87 @@ mod tests {
...
+
+        let upstream = payload
+            .to_upstream_request(false)
+            .expect("function tools allow parallel calls");
+        let value = serde_json::to_value(upstream).unwrap();
+        assert_eq!(value["parallel_tool_calls"], true);
+    }
+
+    #[test]
+    fn to_upstream_request_sets_serial_tool_calls_for_builtin_tools() {
```
jiahuei, line 387: The tests cover built-in-only and function-only requests separately, but there is no explicit mixed-tools regression test. Although the implementation seems to handle the mixed case correctly, I’d suggest adding that test for clarity.

crates/agentic-server-core/src/types/request_response.rs
```diff
@@ -89,6 +91,18 @@ impl RequestPayload {
     /// flat name collides with a top-level function tool or another namespace
     /// member.
     pub fn to_upstream_request(&self, stream: bool) -> Result<UpstreamRequest<'_>, ToolError> {
+        let has_gateway_owned_tool = self.declares_gateway_owned_tool();
```
jiahuei: nit: Can we rename `gateway_owned_tool` to `built_in_tool`, following `TERMINOLOGY.md`?

crates/agentic-server-core/src/types/request_response.rs
harivilasp: Added a mixed built-in plus function-tool regression test in cd11ad7. It verifies mixed requests default to serial `parallel_tool_calls: false` while preserving both upstream tools.

crates/agentic-server-core/src/types/request_response.rs
harivilasp: Renamed the local helper/variable from gateway-owned terminology to `built_in_tool` in cd11ad7 to match TERMINOLOGY.md.

crates/agentic-server-core/src/types/request_response.rs
```diff
@@ -113,9 +127,24 @@ impl RequestPayload {
...
+    fn declares_built_in_tool(&self) -> bool {
+        self.tools.as_deref().is_some_and(|tools| {
+            tools.iter().any(|tool| {
+                matches!(
+                    tool,
+                    ResponsesTool::Mcp(_)
+                        | ResponsesTool::WebSearch(_)
+                        | ResponsesTool::FileSearch(_)
+                        | ResponsesTool::CodeInterpreter(_)
+                )
```
maralbahari: declares_built_in_tool's `matches!(Mcp | WebSearch | FileSearch | CodeInterpreter)`  is a hand-maintained copy of the logic in `ToolType::is_gateway_owned (!matches!(Function | CodexNamespace))`. When the next built-in variant lands, someone will update one and forget the other.  another regression problem is codex `read_mcp_resource` from client owned is mapped to gateway owned.  so would need a check like
`matches!(tool, ResponsesTool::Function(p) if maybe_mcp_function(p).is_some()))` to put everything together then a fix like below would resolve the issue:

```rust
impl ResponsesTool {
    /// The `ToolType` the registry would assign this declaration; `None` for `Unknown`.
    /// Mirrors the classification in `ToolRegistry::build_with_handlers`.
    fn tool_type(&self) -> Option<ToolType> {
        match self {
            // read_mcp_resource functions with valid MCP metadata register as
            // MCP entries (registry.rs), so classify them the same way here.
            Self::Function(p) if maybe_mcp_function(p).is_some_and(|params| !params.is_empty()) => {
                Some(ToolType::Mcp)
            }
            Self::Function(_) => Some(ToolType::Function),
            Self::Namespace(_) => Some(ToolType::CodexNamespace),
            Self::Mcp(_) => Some(ToolType::Mcp),
            Self::WebSearch(_) => Some(ToolType::WebSearch),
            Self::FileSearch(_) => Some(ToolType::FileSearch),
            Self::CodeInterpreter(_) => Some(ToolType::CodeInterpreter),
            Self::Unknown => None,
        }
    }

    pub fn is_gateway_owned(&self) -> bool {
        self.tool_type().is_some_and(ToolType::is_gateway_owned)
    }
}
```
Then `declares_built_in_tool` collapses to `tools.iter().any(ResponsesTool::is_gateway_owned)`. One source of truth, no signature churn, and `to_function_tools` keeps its single job.

maralbahari (changes requested):

crates/agentic-server-core/src/types/request_response.rs
jiahuei: agreed, on this note, perhaps the test can be made more comprehensive by randomly selecting a supported built-in tool in the request

maralbahari: @harivilasp the test `test_gateway_normalization_preserves_parallel_tool_calls` covers only the happy path. as @jiahuei suggested, maybe could have a more comprehensive test? random selection of supported tools and both parallel bool type?

harivilasp: Addressed in ea65d78. The mixed-tools test now iterates every supported built-in declaration alongside a client function and verifies both values: `parallel_tool_calls: false` is accepted and normalized to serial, while `true` is rejected. This is deterministic rather than randomly sampled, so each supported tool is covered on every run.

maralbahari: @harivilasp Thanks, sorry the PR https://github.com/vllm-project/agentic-api/pull/101 is merged  and there is conflict on `tool/normalize.rs` could you please resolve the conflict?

harivilasp: sure resolving it

harivilasp: should be resolved now

maralbahari (approved):
@harivilasp  LGTM, Thanks

Issue #113, opened 2026-07-14
[Claude Code - Tracking] Integration via Anthropic Messages API


ashwing:
## Overview

Support **Claude Code** as a client of `agentic-api`, the way Codex is supported today. Claude Code speaks the Anthropic **Messages API** (`/v1/messages`), not Responses — so this is a new surface, but a **stateless** one: Claude Code sends the full conversation every turn, so the gateway holds no state (no storage, no rehydration, no persistence). Design doc and live-spike results: #98. Executes [ROADMAP §4 Messages API](https://github.com/vllm-project/agentic-api/blob/main/ROADMAP.md).

## What the gateway must implement

1. Transparent `/v1/messages` + `/v1/messages/count_tokens` proxy to vLLM.
2. The Anthropic gateway protocol contract (forward `anthropic-*` headers + error bodies unchanged, preserve the `system` attribution block, accept either auth header).
3. A **server-side** tool loop that executes gateway-owned tools (`web_search`, MCP) without exposing them to the client, mapping Anthropic `tool_use`/`tool_result` to the shared tool framework.

No storage layer. No `previous_response_id` analog. No stored-response object.

## Spike findings (already settled, see #98)

Verified live against the real Claude Code CLI (2.1.208) + vLLM 0.25.1:
- **Gateway-owned tools only work server-side.** Claude Code executes every `tool_use` against its own local registry and errors `No such tool available` on anything it didn't declare — it never waits for the gateway to supply a `tool_result`. Advertising a gateway tool to the client cannot work; the gateway must run the loop itself and surface only the final message.
- **vLLM 0.25.1 accepts every advanced field Claude Code sends** (`thinking` adaptive/enabled, `output_config`, `context_management`, `metadata`) at HTTP 200, alone and combined — no field-level `400` hazard. Only real constraint is context-window sizing.

## Stages

- [x] **Stage 0 — spike / de-risk** (done; findings above and in #98)
- [x] **Stage 1 — transparent proxy + protocol contract** → #114 (proxy landed in #99)
- [ ] **Stage 2 — server-side gateway tool loop** (the core value-add) → #115
- [ ] **Stage 3 — parity hardening** (cassettes + streaming/tool edge cases) → #116

PR #131, opened 2026-07-16
feat: run gateway tool loop natively on /v1/messages

ashwing:
Part of #113; implements #115 (Stage 2).

## What

Run the **gateway tool loop natively on `/v1/messages`**: when a request declares a gateway-owned tool, the gateway executes it server-side, hides it, and returns only the final assistant message. No gateway tool → unchanged transparent proxy.

This also makes **Claude Code's web search run on the gateway** — the Codex experience, for Claude Code.

## How

The loop is native to Anthropic Messages — it talks vLLM `/v1/messages` upstream and forwards the client's request untouched, so every Anthropic field (`tool_choice`, `stop_sequences`, thinking, …) is preserved. It reuses only the protocol-neutral tool layer (`ToolRegistry::dispatch`) via a small seam; it never touches `RequestPayload`/`ResponsePayload`.

- `types/messages/tool_seam` — `tool_use` ↔ `FunctionToolCall`, `ToolOutput` → `tool_result`, gateway/client classification.
- `executor::messages_loop` — non-streaming loop (hide-the-call, feed `tool_result` back, round cap).
- `executor::messages_stream` — streaming loop + accumulator presenting one logical message across rounds (single lifecycle, contiguous block indices, gateway `tool_use` suppressed).
- `handler` — routes gateway-tool requests to the loop; Anthropic error envelope on failure.

### Claude Code web search

Claude Code declares web search as a client tool named `WebSearch` (not a typed server tool like Codex's `web_search`), so it is client-owned by default. Operators opt in with `MESSAGES_GATEWAY_TOOL_ALIASES="WebSearch=web_search"` (empty by default) to have the gateway execute it — the ownership doctrine's "client function is client-owned *unless configured* as gateway-owned" ([`codex-integration.md`](docs/design/codex-integration.md)). Dispatch canonicalizes the name to the `web_search` executor and adapts the arguments (`allowed_domains`→`include_domains`, `blocked_domains`→`exclude_domains`). See the README's Claude Code section and `docs/design/messages-gateway-tool-classification.md`.

## Tests

- `cargo fmt --check` / `clippy --workspace --all-targets -D warnings` / `cargo test --workspace` — clean.
- Unit: tool seam (classification, alias parsing, name canonicalization, arg adapter), streaming accumulator.
- Acceptance: replay live-recorded `/v1/messages` cassettes — single-round, parallel, sequential multi-round, no-tool-call, streaming.
- Edges: tool failure/timeout → error `tool_result`, upstream error → Anthropic envelope, round cap, malformed args, mixed client+gateway tools, proxy fallthrough.
- **Live (G6e, Qwen3 + real You.com):** curl scenarios; real Codex CLI (typed `web_search`) and real Claude Code CLI (aliased `WebSearch`) both drove the loop server-side with the call hidden — routed `messages_loop`, live search backend hit, single streaming lifecycle, no raw-Responses leakage.

## Known limitations (intentional)

- **`pause_turn` surfaced, not continued** — only `stop_reason: tool_use` drives another round; other stops return to the client.
- **MCP over Messages out of scope** — its wire shape isn't expressible in the Anthropic tool declaration yet.
- **`WebSearch` domain filters** — the executor treats include/exclude domains as mutually exclusive, so a call setting both `allowed_domains` and `blocked_domains` returns a graceful error result.

## Follow-up: consolidation with the Responses loop

This adds a second gateway-tool loop alongside the Responses one; the duplication is deliberate (different wire shapes; the typed core shouldn't be reshaped ahead of the Layering ADR). Tool execution is already shared. Planned follow-ups (post-ADR): a `GatewayTurn` trait + one generic loop, a shared stream-normalizer contract, hoisted shared limits, and structural typed-tool + per-request opt-in classification on the Messages side (extends the same `GatewayToolMap` seam).


crates/agentic-server-core/src/executor/messages_stream.rs
```diff
@@ -0,0 +1,445 @@
...
+
+/// State machine that turns per-round Anthropic SSE into one client-visible
+/// message. Fed line-by-line via [`Self::push`].
+struct MessagesStreamAccumulator {
+    message_started: bool,
+    /// Next client-visible block index (contiguous across rounds).
+    next_index: u32,
+    /// Map upstream (per-round) block index → client index, for the blocks we
+    /// forward this round. Cleared each round.
+    index_map: std::collections::HashMap<u64, u32>,
```
maralbahari: you can use top of the file importing for the HashMap

crates/agentic-server-core/src/executor/messages_stream.rs
```diff
@@ -0,0 +1,445 @@
...
+    /// Translate one upstream SSE line into zero or more client SSE lines.
+    fn push(&mut self, line: &str) -> Vec<String> {
+        let Some(data) = line.strip_prefix("data: ") else {
+            return Vec::new();
+        };
+        let data = data.trim();
+        if data == "[DONE]" {
+            return Vec::new();
+        }
+        let Ok(mut event) = serde_json::from_str::<Value>(data) else {
```
maralbahari, line 251: could use common utility function from `utils::common:`?

crates/agentic-server-core/src/executor/messages_loop.rs
```diff
@@ -0,0 +1,171 @@
...
+        }
+    });
+    join_all(futures).await
+}
+
+/// Append the assistant turn (its gateway `tool_use` blocks) and a following
+/// user turn (the matching `tool_result` blocks) to the request `messages` so
+/// the next upstream round sees the resolved calls. These stay internal — the
+/// client never sees them (hide-the-call).
+fn append_round_to_history(request: &mut Value, resolved: &[ResolvedCall]) {
```
maralbahari: this is also repeated in `messages_stream.rs`
and both implementations reconstruct the assistant turn using only gateway tool_use blocks. Any preceding text, thinking, redacted-thinking, or signature blocks are discarded. This loses conversation state and can make extended-thinking follow-up requests invalid. Pass the original/reconstructed complete assistant content into `append_round_to_history`, preserving block order, while suppressing gateway blocks only from the client-facing response.

crates/agentic-server-core/src/executor/messages_stream.rs
```diff
@@ -0,0 +1,445 @@
...
+    let event = json!({"type": "error", "error": {"type": "api_error", "message": message}});
+    let json = String::from_utf8(serialize_to_vec_or_default(&event)).unwrap_or_default();
+    format!("event: error\ndata: {json}\n\n")
+}
+
+/// Execute reconstructed gateway calls (concurrent, per-call timeout). Errors
+/// become error `tool_result`s (E5).
+async fn execute_gateway_calls(calls: &[StreamedCall], registry: &ToolRegistry) -> Vec<ResolvedStreamCall> {
+    let futures = calls.iter().map(|c| async move {
+        let input: Value = serde_json::from_str(&c.input_json).unwrap_or_else(|_| json!({}));
```
maralbahari: Malformed or incomplete partial_json silently becomes `{}`. This can execute a real gateway tool with arguments the model never supplied. Parsing failure should produce an `is_error: true` tool result—or terminate with a protocol error—without dispatching the tool. Missing IDs/names should likewise be rejected rather than defaulted.

crates/agentic-server-core/src/executor/messages_loop.rs
```diff
@@ -0,0 +1,171 @@
...
+                    gateway_calls.push(block.clone());
+                } else {
+                    has_client_tool_use = true;
+                }
+            }
+        }
+
+        // Terminal: the model didn't ask for a gateway tool (or asked for a
+        // client tool we must hand back, or stopped for another reason).
+        if gateway_calls.is_empty() || has_client_tool_use || stop_reason != Some("tool_use") {
```
maralbahari: When one assistant turn contains both kinds of tool call, the function returns the original message unchanged. That exposes the gateway-owned call, contradicting the module’s “hidden gateway calls” contract and differing from streaming behavior, which suppresses it. Return a filtered client-visible message or define a consistent mixed-call policy across both paths.

crates/agentic-server-core/src/executor/messages_stream.rs
```diff
@@ -0,0 +1,445 @@
...
+            suppressed_indices: std::collections::HashSet::new(),
+            calls: Vec::new(),
+            call_by_index: std::collections::HashMap::new(),
+            ended_on_tool_use: false,
+            has_client_tool_use: false,
+            final_message_delta: None,
+        }
+    }
+
+    fn begin_round(&mut self) {
```
maralbahari, line 198: `begin_round()` does not reset `final_message_delta`. If a subsequent upstream response ends cleanly without a `message_delta`, `finish()` can emit the preceding round’s stop_reason: tool_use, followed by a fabricated message_stop. Reset round-terminal state and require a valid terminal sequence before treating EOF as success.

maralbahari (commented):
@ashwing thanks for the PR. I added some inline comments

crates/agentic-server-core/src/executor/messages_stream.rs
ashwing: Done, moved to the top-of-file imports.

crates/agentic-server-core/src/executor/messages_stream.rs
ashwing: Switched `sse`/`error_sse` over to `serialize_to_string` from `utils::common`.

crates/agentic-server-core/src/executor/messages_loop.rs
ashwing: Good catch — both loops were rebuilding the assistant turn from just the gateway `tool_use`. Fixed both: non-streaming feeds the model's full `content` forward, and streaming now buffers every block per round (thinking/text/signature + tool_use) and reconstructs the complete turn in order, suppressing the gateway call from the client only. Added repro tests for both paths that assert thinking + signature survive into the next round.

crates/agentic-server-core/src/executor/messages_stream.rs
ashwing: Fixed. Invalid/incomplete `partial_json` (and non-object input on the non-streaming side) now yields an `is_error: true` tool_result instead of dispatching with `{}` — the tool never runs on args the model didn't send. Same handling in both loops.

crates/agentic-server-core/src/executor/messages_loop.rs
ashwing: Went with the consistent mixed-call policy: hide the gateway call, surface the client `tool_use`, stop the loop — matching streaming. Non-streaming now strips gateway `tool_use` from the client-facing content instead of returning the message verbatim.

crates/agentic-server-core/src/executor/messages_stream.rs
ashwing: Fixed — `begin_round()` now resets `final_message_delta`, so a clean-EOF round can't re-emit the previous round's `stop_reason: tool_use`. Repro test covers the tool_use-round-then-clean-EOF sequence.

crates/agentic-server-core/src/executor/messages_loop.rs
```diff
@@ -0,0 +1,188 @@
...
+///
+/// `request` is the client's parsed request body as JSON — forwarded upstream
+/// with `stream:false` forced and its `messages` extended each round.
+///
+/// # Errors
+/// Returns [`ExecutorError`] on upstream failure or unparseable upstream JSON.
+/// Gateway-tool execution failures do **not** error — they become error
+/// `tool_result`s fed back to the model.
+pub async fn run_messages_loop(
+    mut request: Value,
```
maralbahari, line 56: why instead of raw json value not using `MessagesRequest`  that you have in `types/messages.request.rs` ?

maralbahari: @ashwing if the raw lossless json is really required maybe consider passing a per-request context containing both the typed `MessagesRequest` for safe field access and the raw JSON Value for lossless upstream forwarding? or you can follow the pattern we use for RequestPayload for Responses API where it implements `to_upstream` function.  so wouldnt need to maintain a context where it holds lossless for upstream and typed version for gateway.

ashwing: Fair question. The difference from `RequestPayload::to_upstream_request` is that that path *transforms* — it flattens Codex namespace tools, normalizes every tool type to `Vec<FunctionTool>`, resolves `tool_choice`. Typed→typed is the actual work there. This loop is a pass-through: it forwards the client's request to vLLM `/v1/messages` verbatim and only reads `tools` + rewrites `messages`/`stream`, so there's no transform step a typed request would carry.

One concrete catch with going fully typed here: `ContentBlock` has `#[serde(other)] Unknown`, so deserializing→reserializing through `MessagesRequest` would silently drop any block type we don't model (redacted_thinking, future block types, provider extensions). Raw `Value` forwards those untouched, which matters for a proxy.

That said — you're right that a per-request context (typed `MessagesRequest` for safe field access + raw `Value` for lossless forwarding) is the cleaner long-term shape, and it lines up with the loop-consolidation follow-up already noted in the PR description. I'd rather not reshape the request plumbing inside this PR (it'd touch the routing path and widen the diff), so I'll file it as a follow-up and link it here. Sound good?

maralbahari: @ashwing thanks ashwin, you can create Issue for this for the follow up. approving this PR.

maralbahari (approved):

PR #135, opened 2026-07-20
fix: hashing long namespace flatten name

haoshan98:
- Bound gateway-generated Codex namespace member names to the upstream 64-character function-name limit while leaving shorter names unchanged.
- Shorten long names with a readable prefix and deterministic 16-hex fingerprint of the complete name. The request-scoped namespace map continues to rewrite namespaced `tool_choice` values and restore function calls to their public `{ namespace, name }` representation.
- Reject ambiguous generated names when they collide with a function-call registry key claimed by a function tool, an `mcp` built-in tool, a normalized built-in tool, or another namespace member. Validation runs before upstream inference or MCP handler construction.
- Add boundary, determinism, Unicode, `tool_choice`, restoration, WebSocket round-trip, and MCP collision regression coverage, and document the bounded model-visible format and collision behavior.

### DIFF
--- a/crates/agentic-server-core/src/tool/codex.rs
+++ b/crates/agentic-server-core/src/tool/codex.rs
@@ -13,10 +13,30 @@ use super::registry::{ToolEntry, ToolType};
 // unlikely to collide with user functions, and can be restored to
 // `{ namespace, name }` on the way back to the client.
 pub const MODEL_VISIBLE_NAMESPACE_MEMBER_PREFIX: &str = "agentic_ns__";
+pub const MAX_MODEL_VISIBLE_TOOL_NAME_LEN: usize = 64;
+
+const HASHED_NAMESPACE_MEMBER_SUFFIX_LEN: usize = 18;
+
+fn stable_name_hash(value: &str) -> u64 {
+    const FNV_OFFSET_BASIS: u64 = 0xcbf2_9ce4_8422_2325;
+    const FNV_PRIME: u64 = 0x0000_0100_0000_01b3;
+
+    value.bytes().fold(FNV_OFFSET_BASIS, |hash, byte| {
+        (hash ^ u64::from(byte)).wrapping_mul(FNV_PRIME)
+    })
+}
 
 #[must_use]
 pub fn model_visible_namespace_member_name(namespace: &str, member: &str) -> String {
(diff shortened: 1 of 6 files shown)


franciscojavierarceo (commented):
the generated shortened name needs to check against MCP registry keys. 

`typed_top_level_tool_names` excludes `ResponsesTool::Mcp`, but `ToolRegistry` later inserts MCP entries under `p.name`. 

If a later MCP declaration has a `p.name` matching the shortened name, it overwrites the client-owned namespace entry and the model’s call is dispatched as gateway-owned MCP work. 

Before this change the long name was rejected upstream; shortening makes this misrouting path reachable. 

Please reject collisions across all model-visible registry keys and add a regression test with a long namespace member followed by a matching MCP declaration.

maralbahari (approved):
@haoshan98 Thanks for the fix can you fix the merge conflict please. LGTM.

maralbahari (changes requested):
@haoshan98 will merge after addressing @franciscojavierarceo comment and merge conflict

crates/agentic-server-core/src/tool/codex.rs
```diff
@@ -59,6 +79,27 @@ struct NamespaceCallMapping {
     upstream_name: String,
 }
 
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+enum TopLevelRegistryToolKind {
```
maralbahari: this Enum sounds the same to the `ResponsesTool`. the `ResponsesTool` has `tool_type` function see if you can reuse that?

franciscojavierarceo: i thought we wanted more separation between Messages and Responses not less?

maralbahari: @franciscojavierarceo yes there will be more separation my request is simply that we do have `ToolType` enum and it is used in `ResponsesToo` for mapping between tool and `tool_type` also used in `tool_registry`. so the `ToolType` enum as  one single ground truth of what is supported on gateway rather than having to maintain multiple enums and might forget to update them. the `ResponseTool` is only used for responses api for now. on messages api as off now based on https://github.com/vllm-project/agentic-api/pull/131 PR merged is maintaining separate contract and a tool build registry that is plugin web search only as off now.

maralbahari (approved):
LGTM

PR #136, opened 2026-07-21
fix: normalize multi-round gateway streaming

harivilasp:
## Summary

- Normalize multi-round gateway streaming into one client-visible event sequence.
- Preserve lossless upstream event frames while emitting gateway-executed built-in tool lifecycle events.
- Supersedes #132, which GitHub closed automatically after its source branch was renamed.

## Test Plan

- `cargo fmt -- --check`
- `cargo test`
- `pre-commit run --all-files`

### DIFF
--- a/crates/agentic-server-core/src/events/normalize.rs
+++ b/crates/agentic-server-core/src/events/normalize.rs
@@ -16,7 +16,10 @@ pub fn normalize_sse_line(line: &str) -> Option<EventFrame> {
     }
 
     let json: Value = deserialize_from_str_opt(data_str)?;
-    let event_type = SSEEventType::from(json.get("type")?.as_str()?);
+    let event_type = json
+        .get("type")
+        .and_then(Value::as_str)
+        .map_or(SSEEventType::Other, SSEEventType::from);
 
     let payload = extract_payload(event_type, &json);
     let wire: WireEvent = deserialize_from_value_opt(json)?;
--- a/crates/agentic-server-core/src/events/normalize.rs
+++ b/crates/agentic-server-core/src/events/normalize.rs
@@ -15,54 +15,17 @@ pub fn normalize_sse_line(line: &str) -> Option<EventFrame> {
         return None;
     }
 
-    let wire: WireEvent = deserialize_from_str_opt(data_str)?;
-    let json = wire.to_value();
-
-    let event_type = classify_event_type(&wire.event_type);
+    let json: Value = deserialize_from_str_opt(data_str)?;
+    let event_type = SSEEventType::from(json.get("type")?.as_str()?);
 
     let payload = extract_payload(event_type, &json);
+    let wire: WireEvent = deserialize_from_value_opt(json)?;
 
     Some(EventFrame {
         event_type,
         payload,
-        sequence_number: wire.sequence_number,
         wire,
     })
-}
-
-/// Map a wire-format event type string to our enum.
-fn classify_event_type(type_str: &str) -> SSEEventType {
-    match type_str {
-        "response.created" => SSEEventType::ResponseCreated,
-        "response.in_progress" => SSEEventType::ResponseInProgress,
-        "response.completed" | "response.done" => SSEEventType::ResponseCompleted,
-        "response.failed" => SSEEventType::ResponseFailed,
-        "response.incomplete" => SSEEventType::ResponseIncomplete,
-        "response.output_item.added" => SSEEventType::OutputItemAdded,
-        "response.output_item.done" => SSEEventType::OutputItemDone,
-        "response.output_text.delta" => SSEEventType::OutputTextDelta,
-        "response.output_text.done" => SSEEventType::OutputTextDone,
-        "response.content_part.added" => SSEEventType::ContentPartAdded,
-        "response.content_part.done" => SSEEventType::ContentPartDone,
-        "response.function_call_arguments.delta" => SSEEventType::FunctionCallArgumentsDelta,
-        "response.function_call_arguments.done" => SSEEventType::FunctionCallArgumentsDone,
-        "response.custom_tool_call_input.delta" => SSEEventType::CustomToolCallInputDelta,
-        "response.custom_tool_call_input.done" => SSEEventType::CustomToolCallInputDone,
-        "response.reasoning_text.delta" => SSEEventType::ReasoningTextDelta,
-        "response.reasoning_text.done" => SSEEventType::ReasoningTextDone,
-        "response.reasoning_part.added" => SSEEventType::ReasoningPartAdded,
-        "response.reasoning_part.done" => SSEEventType::ReasoningPartDone,
-        "response.reasoning_summary_text.delta" => SSEEventType::ReasoningSummaryTextDelta,
-        "response.reasoning_summary_text.done" => SSEEventType::ReasoningSummaryTextDone,
-        "response.file_search_call.searching" => SSEEventType::FileSearchCallSearching,
-        "response.file_search_call.completed" => SSEEventType::FileSearchCallCompleted,
-        "response.web_search_call.in_progress" => SSEEventType::WebSearchCallInProgress,
-        "response.web_search_call.searching" => SSEEventType::WebSearchCallSearching,
-        "response.web_search_call.completed" => SSEEventType::WebSearchCallCompleted,
-        "response.mcp_tool_call.in_progress" => SSEEventType::McpToolCallInProgress,
-        "response.mcp_tool_call.completed" => SSEEventType::McpToolCallCompleted,
-        _ => SSEEventType::Other,
-    }
 }
 
 /// Extract a typed payload from the JSON body based on the classified event type.
--- a/crates/agentic-server-core/src/events/normalize.rs
+++ b/crates/agentic-server-core/src/events/normalize.rs
@@ -1,6 +1,6 @@
 use serde_json::Value;
 
-use super::types::{EventFrame, EventPayload, SSEEventType, SSEItemType};
+use super::types::{EventFrame, EventPayload, SSEEventType, SSEItemType, WireEvent};
 use crate::utils::common::{deserialize_from_str_opt, deserialize_from_value_opt};
 
 /// Normalize a raw SSE data line into a typed [`EventFrame`].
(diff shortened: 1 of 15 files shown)


crates/agentic-server-core/src/executor/engine.rs
```diff
@@ -249,23 +259,26 @@ fn run_stream(ctx: RequestContext, exec_ctx: Arc<ExecutionContext>, auth: Option
...
+                        Ok((Ok((payload, ctx)), mut stream_accumulator)) => {
                             // Codex may close its WebSocket as soon as it receives
                             // `response.completed`. Persist before exposing that
                             // event so a custom call/output continuation cannot be
                             // cancelled by the client disconnect.
-                            let terminal_event = payload.as_terminal_response_chunk();
                             let ch = exec_ctx.conv_handler.clone();
                             let rh = exec_ctx.resp_handler.clone();
-                            if let Err(e) = persist_if_needed(payload, ctx, ch, rh).await {
+                            if let Err(e) = persist_if_needed(payload.clone(), ctx, ch, rh).await {
```
maralbahari: why there is a need for `payload.clone()` ? the older version didnt require it.

crates/agentic-server-core/src/executor/upstream.rs
```diff
@@ -193,7 +182,7 @@ fn defer_or_flush_function_event(
             pending_unnamed_function_events
                 .entry(item_id.clone())
                 .or_default()
-                .push(line.to_owned());
+                .push(frame.clone());
```
maralbahari: the entire `frame.clone()` seems too expensive. see how can be it be avoided.

crates/agentic-server-core/src/executor/gateway_accumulator.rs
```diff
@@ -0,0 +1,212 @@
...
+        }
+        frame.wire.sequence_number = Some(self.take_sequence_number());
+        rebase_output_index(&mut frame.wire, output_offset);
+        true
+    }
+
+    pub(crate) fn terminal_response_chunk(&mut self, payload: &ResponsePayload) -> ExecutorResult<Option<String>> {
+        let mut frame = terminal_response_frame(payload)?;
+        if !self.process_event(&mut frame, 0) {
+            return Ok(None);
```
maralbahari: A payload with final status `in_progress` maps to `SSEEventType::ResponseInProgress` (line 95), which the lifecycle dedup has usually already consumed from the relayed stream, so this returns `Ok(None)` and the client gets `[DONE]` with no terminal `response` snapshot. The terminal event is not the same thing as the early lifecycle announcement; the previous revision always emitted it, and no test covers this case.

crates/agentic-server-core/src/events/normalize.rs
```diff
@@ -16,58 +16,18 @@ pub fn normalize_sse_line(line: &str) -> Option<EventFrame> {
...
 
     let json: Value = deserialize_from_str_opt(data_str)?;
-
-    let event_type = json
-        .get("type")
-        .and_then(Value::as_str)
-        .map_or(SSEEventType::Other, classify_event_type);
-
-    let sequence_number = json.get("sequence_number").and_then(Value::as_u64);
+    let event_type = SSEEventType::from(json.get("type")?.as_str()?);
```
ashwing: The switch to `SSEEventType::from(json.get("type")?.as_str()?)` drops any `data:` frame that parses as JSON but has no string `type` — previously `map_or(SSEEventType::Other, ...)` kept it and forwarded it as `Other`. Given the "preserve lossless upstream frames" goal, is dropping intended here? Responses SSE always carries `type` in practice, so likely benign, but a typeless keepalive/data frame from a provider would now vanish silently instead of passing through. Either keeping the `Other` fallback for the type-missing case (distinct from malformed JSON, which should still drop) or a one-line comment noting the deliberate drop would make the intent clear — plus an `assert!(normalize_sse_line("data: {\"foo\":1}").is_none())` (or `.is_some()`) to lock it in; the normalizer tests currently cover `[DONE]`, malformed JSON, and unknown `type`, but not valid-JSON-without-`type`.

ashwing (approved):
Reworked cleanly — the accumulator is owned now (no `Arc<Mutex>` guard held across `.await`), error frames route through `error_chunk` so they stay numbered, the synthetic/deferred events share one `EventFrame`/`process_event` path, and the multi-round test covers single-lifecycle + monotonic sequence + cross-round `output_index`. One question on `normalize_sse_line` below; non-blocking.

crates/agentic-server-core/src/executor/engine.rs
```diff
@@ -249,23 +259,26 @@ fn run_stream(ctx: RequestContext, exec_ctx: Arc<ExecutionContext>, auth: Option
...
                                 warn!("persist failed: {e}");
                             }
 
-                            yield terminal_event;
+                            match stream_accumulator.terminal_response_chunk(&payload) {
+                                Ok(Some(chunk)) => yield chunk,
+                                Ok(None) => {}
```
maralbahari: Ok(None)  seems to be regression in behavior compared to the older version. Silently dropping the terminal chunk means the stream ends with bare `[DONE]` and no final `response` state. If `Ok(None)` is really possible here, it needs a log and a test; otherwise make the terminal chunk unsuppressible and delete this arm.

maralbahari (commented):
@harivilasp thanks for the changes. I added a few concerns inline comments.

harivilasp:
Thank you for the detailed review. All five points are addressed in ae0f5b1, with focused regressions plus the full local test and lint suite.

crates/agentic-server-core/src/executor/engine.rs
harivilasp: Removed in ae0f5b1. The terminal SSE chunk is now prepared before persistence, then the payload moves into persist_if_needed. This keeps persistence before client-visible emission without cloning the payload.

crates/agentic-server-core/src/executor/upstream.rs
harivilasp: Removed in ae0f5b1. ResponseAccumulator now processes the frame before stream handling, and deferred EventFrames are moved into the pending queue and moved out again on flush. No deep frame clone remains.

crates/agentic-server-core/src/executor/gateway_accumulator.rs
harivilasp: Fixed in ae0f5b1. Terminal frames now bypass lifecycle dedup but still use the shared sequence stamper, so a final in_progress response is always emitted with the next sequence number. Added emits_in_progress_terminal_event_after_lifecycle_event.

crates/agentic-server-core/src/events/normalize.rs
harivilasp: Fixed in ae0f5b1. A missing or non-string type maps to Other; WireEvent preserves the absent wire field during serialization, so the frame relays losslessly. Added test_typeless_json_event_is_preserved.

crates/agentic-server-core/src/executor/engine.rs
harivilasp: Fixed in ae0f5b1 alongside the terminal lifecycle fix. terminal_response_chunk now always returns a serialized terminal frame, so the bare DONE path and the Ok(None) arm are gone.

crates/agentic-server-core/src/executor/accumulator.rs
```diff
@@ -224,30 +224,29 @@ impl ResponseAccumulator {
 
     pub(crate) fn process_sse_line(&mut self, line: &str) {
         if let Some(frame) = normalize_sse_line(line) {
```
maralbahari: @harivilasp could you please add back the `capture_terminal_details_if_needed` here. instead of `process_event`
follow the same behavior as the original code.

maralbahari:
> Thank you for the detailed review. All five points are addressed in [ae0f5b1](https://github.com/vllm-project/agentic-api/commit/ae0f5b12ad4e293ff11fcf4cd8be9dc711841775), with focused regressions plus the full local test and lint suite.

@harivilasp Thank you so much for the hard work. there has been many back and forth to address design and regression since this is a hot path in the agentic-loop need to be extra careful not introducing overhead and regression in behaviors.

crates/agentic-server-core/src/executor/engine.rs
```diff
@@ -249,23 +259,26 @@ fn run_stream(ctx: RequestContext, exec_ctx: Arc<ExecutionContext>, auth: Option
                             yield error_sse_chunk(&format!("stream task failed: {e}"));
```
franciscojavierarceo: Could we keep this error on the same sequence allocator as the rest of the stream? If the spawned task panics after sending one or more events, this branch emits an error without a sequence_number, so the public sequence is no longer contiguous. A regression test that panics after the first emitted event would cover the gap.

crates/agentic-server-core/src/executor/engine.rs
```diff
@@ -135,8 +135,15 @@ async fn run_until_gateway_tools_complete(
             }
         }
         let has_client_owned = has_client_owned_calls(&current_output, &registry);
-        let gateway_results =
-            execute_and_emit_output_calls(&current_output, &registry, combined_output.len(), stream_events).await?;
+        let gateway_results = execute_and_emit_output_calls(
```
franciscojavierarceo: What happens when one upstream response contains a gateway-owned call at output index 0 and a client-owned function call at index 1? fetch_stream_payload emits the client call immediately, but the gateway lifecycle is not synthesized until this call runs, so clients see index 1 before index 0. Could we buffer the later public events, or emit the gateway lifecycle when we identify the call, and add a mixed-call streaming test?

harivilasp:
Thank you for the careful feedback. It is good for my learning as well as I work through review

Local verification passed: cargo fmt --all -- --check, cargo check, cargo clippy --all-targets -- -D warnings, cargo test (464 passed, 1 ignored), all pre-commit hooks, and git diff --check.

I am open to any further feedback.

crates/agentic-server-core/src/executor/accumulator.rs
harivilasp: Terminal failure details are captured again at the raw SSE boundary, matching the original behavior while keeping a single normalization pass.

crates/agentic-server-core/src/executor/engine.rs
harivilasp: Spawn-task panic errors now use the same contiguous sequence allocator as the rest of the stream, with a panic-after-first-event regression test.

crates/agentic-server-core/src/executor/engine.rs
harivilasp: Mixed gateway-owned and client-owned calls are emitted in output-index order. All gateway calls in the round still execute concurrently, unnamed client events remain ordered, and deferred provider-specific frames remain lossless.

franciscojavierarceo (approved):
Thanks for the fixes. The panic path now keeps the stream sequence contiguous, and the mixed gateway/client ordering test covers the other issue. Looks good to me.

PR #139, opened 2026-07-21
fix: align MCP call responses and streaming events with OpenAI [2/2]

maralbahari:
## Summary

This PR follows the MCP request, discovery, and execution changes from the first MCP shipment. It aligns the gateway's public MCP tool-call responses with the OpenAI Responses contract, using recorded OpenAI behavior as the ground truth.

This PR:

- Replaces the gateway-specific `mcp_tool_call` output item with OpenAI's `mcp_call` item.
- Exposes the original MCP identity through:
  - `server_label`
  - `name`
  - string-encoded `arguments`
- Adds the OpenAI-compatible `approval_request_id`, `output`, and `error` fields.
- Represents MCP execution failures as structured `mcp_tool_execution_error` values.
- Emits the OpenAI MCP call streaming lifecycle:
  - `response.mcp_call.in_progress`
  - `response.mcp_call_arguments.delta`
  - `response.mcp_call_arguments.done`
  - `response.mcp_call.completed`
  - `response.mcp_call.failed`
- Ensures internal function calls used for upstream inference do not leak into the public response.
- Updates response accumulation to recognize and preserve `mcp_call` output items.
- Adds a recorder that captures matching MCP scenarios against OpenAI and the gateway.
- Adds paired OpenAI and gateway cassettes for:
  - tool discovery without invocation
  - successful streaming tool calls
  - successful blocking tool calls
  - missing required arguments
  - invalid argument types
- Compares gateway output items, error shapes, and streaming event transitions directly against the recorded OpenAI behavior.

The separate OpenAI `mcp_list_tools` discovery lifecycle is not emitted yet and remains follow-up work.

This PR is stacked on the first MCP shipment and should initially target `fix-mcp-tool-call-1`. After the first PR merges, it can be rebased and retargeted to `main`.

## Test Plan

- `cargo check --workspace`
- `cargo test --workspace`
- `cargo test -p agentic-server-core --test mcp_tool_test`
- `cargo fmt -- --check`


franciscojavierarceo (commented):
One additional documentation issue could not be attached inline because the file is outside this PR's diff:

[P2] `docs/design/codex-integration.md` lines 140 and 151 still document the removed `mcp_tool_call` wire shape. Please update both tables to describe `mcp_call`, `server_label`, `name`, `output`, structured errors, and the new streaming lifecycle.

The PR head otherwise passed the full `agentic-server-core` test suite, targeted Clippy with warnings denied, formatting checks, and the current GitHub checks.

crates/agentic-server-core/src/types/io/output.rs
```diff
@@ -256,36 +256,74 @@ impl WebSearchCall {
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct McpCall {
     pub id: String,
-    pub server: String,
-    pub tool: String,
-    pub arguments: Value,
+    pub server_label: String,
+    pub name: String,
+    pub arguments: String,
     pub status: GatewayCallStatus,
```
franciscojavierarceo: [P1] Preserve all documented MCP call statuses

`McpCall.status` reuses `GatewayCallStatus`, which only accepts `in_progress`, `completed`, and `failed`. The OpenAI Responses schema also permits `calling` and `incomplete`, and the status may be omitted. Because `ResponseAccumulator::from_json` later uses `filter_map` when deserializing output items, any of those valid shapes causes the entire `mcp_call` item to disappear silently. Please introduce a dedicated optional `McpCallStatus` covering the documented values and add blocking/streaming tests for `calling`, `incomplete`, and omitted status.

Schema: https://developers.openai.com/api/reference/resources/responses/methods/create

crates/agentic-server-core/src/tool/mcp/handler.rs
```diff
@@ -61,38 +61,31 @@ pub(crate) fn output_item(
...
 }
 
 #[must_use]
 pub(crate) fn started_output_item(call: &FunctionToolCall, tool_ref: &McpToolRef) -> OutputItem {
-    let arguments =
-        deserialize_from_str_opt::<Value>(&call.arguments).unwrap_or_else(|| Value::Object(serde_json::Map::new()));
-
-    OutputItem::McpToolCall(McpToolCall::new(
+    OutputItem::McpCall(McpCall::new(
         call_output_id(call),
```
franciscojavierarceo, line 85: [P2] Reuse one public MCP item ID across the streaming lifecycle

`started_output_item` and `output_item` call `call_output_id` independently. When neither source ID has the expected `fc_`/`call_` prefix, each invocation generates a different UUID. The added/in-progress/arguments events then reference one `item_id`, while completed/failed and output-item-done reference another. `FunctionToolCall` does not enforce either prefix, so this is reachable with other provider ID shapes. Please allocate the public ID once in the event/call state or derive it deterministically, and add a prefixless-ID lifecycle test.

crates/agentic-server-core/tests/cassettes/record_mcp_cassettes.sh
```diff
@@ -0,0 +1,252 @@
...
+SCRIPTS_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "$SCRIPTS_DIR/../../../.." && pwd)"
+BASE_DIR="$SCRIPTS_DIR/mcp"
+TOOLS_FILE="$BASE_DIR/tools.json"
+GATEWAY_URL="${GATEWAY_URL:-http://localhost:9000}"
+MODEL="${MODEL:-Qwen/Qwen3.5-35B-A3B-FP8}"
+MODEL_SLUG="$(echo "$MODEL" | tr '/: ' '---')"
+MCP_RECORD_SET="${MCP_RECORD_SET:-all}"
+COUNTER_SERVER_LABEL="${COUNTER_SERVER_LABEL:-counter}"
+COUNTER_MCP_SERVER_URL="${COUNTER_MCP_SERVER_URL:-https://discharge-exhibit-attempt-sponsored.trycloudflare.com/mcp}"
```
franciscojavierarceo: [P2] Require an explicit trusted MCP endpoint for recording

This defaults the recorder to a public TryCloudflare quick-tunnel hostname that currently does not resolve. `OPENAI_MCP_SERVER_URL` inherits the same nonempty value, so the later required-value check cannot detect missing configuration; the generated tool also uses `require_approval: "never"`. Please remove the public URL default, require an operator-controlled endpoint (or provision a maintained fixture), and validate it before recording or rewriting cassettes.

crates/agentic-server-core/src/executor/accumulator.rs
```diff
@@ -309,7 +309,7 @@ impl ResponseAccumulator {
                 },
             ) => self.complete_custom_tool_call(item_id, item),
             (SSEEventType::OutputItemDone, EventPayload::OutputItemDone { item, .. }) => {
-                if let Some(output_item @ (OutputItem::WebSearchCall(_) | OutputItem::McpToolCall(_))) =
+                if let Some(output_item @ (OutputItem::WebSearchCall(_) | OutputItem::McpCall(_))) =
```
ashwing: `[bug]` `mcp_call` lands in `output` out of order relative to reasoning/message.

`WebSearchCall | McpCall` return `None` at `OutputItemAdded` (line 296), so they're never parked in `in_flight` — then at `OutputItemDone` they `self.output.push(...)` eagerly, while Reasoning/Message/FunctionCall stay in `in_flight` and only flush at finalize. So when reasoning (`output_index` 0) precedes an `mcp_call` (`output_index` 1) — exactly what the gateway cassette emits (`reasoning@0, mcp_call@1, mcp_call@2, reasoning@3, message@4`) — the mcp_call flushes first and reasoning is appended after it. Persisted `output` ends up `[mcp, mcp, reasoning, reasoning, message]` instead of the true interleaving.

`mcp_tool_test.rs:176` doesn't catch it because it reconstructs via `from_json` on the already-ordered terminal payload, never through `from_sse_lines`. This repro fails on the branch:

```rust
#[test]
fn repro_reasoning_before_mcp_call_preserves_order() {
    let lines = vec![
        r#"data: {"type":"response.created","response":{"id":"resp_abc"}}"#.to_string(),
        r#"data: {"type":"response.output_item.added","output_index":0,"item":{"id":"rs_1","type":"reasoning","summary":[]}}"#.to_string(),
        r#"data: {"type":"response.reasoning_text.done","text":"thinking...","item_id":"rs_1"}"#.to_string(),
        r#"data: {"type":"response.output_item.added","output_index":1,"item":{"type":"mcp_call","id":"mcp_1","server_label":"counter","name":"increment","arguments":"","status":"in_progress","approval_request_id":null,"output":null,"error":null}}"#.to_string(),
        r#"data: {"type":"response.mcp_call.completed","item_id":"mcp_1","output_index":1}"#.to_string(),
        r#"data: {"type":"response.output_item.done","output_index":1,"item":{"type":"mcp_call","id":"mcp_1","server_label":"counter","name":"increment","arguments":"{}","status":"completed","approval_request_id":null,"output":"1","error":null}}"#.to_string(),
        r#"data: {"type":"response.reasoning_text.done","text":"thinking...","item_id":"rs_1"}"#.to_string(),
        r#"data: {"type":"response.done","response":{"id":"resp_abc","status":"completed","usage":{"input_tokens":5,"output_tokens":2,"total_tokens":7}}}"#.to_string(),
    ];
    let acc = ResponseAccumulator::from_sse_lines(lines, None);
    assert!(matches!(acc.output[0], OutputItem::Reasoning(_))); // fails: got McpCall
    assert!(matches!(acc.output[1], OutputItem::McpCall(_)));
}
```

Fix: park a placeholder in `in_flight` at `OutputItemAdded` and finalize it in place at `OutputItemDone`, the way `test_accumulator_message_then_reasoning_preserves_order` already relies on. Pre-existing for `WebSearchCall`, but #139 adds `mcp_call` to the same arm.

crates/agentic-server-core/src/types/io/output.rs
```diff
@@ -256,36 +256,74 @@ impl WebSearchCall {
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
-pub struct McpToolCall {
+#[serde(untagged)]
```
ashwing, line 294: `[bug]` unknown `mcp_call.error` shape silently drops the whole item.

Same silent-drop class as your `status` note, but on `error`. `McpCallError` is `#[serde(untagged)]` over just `Text` + `ToolExecution({type, content[]})`. OpenAI's `mcp_call.error` also emits `mcp_protocol_error`/`mcp_http_error` = `{type, code, message}` (no `content` array). No untagged variant matches -> the whole `OutputItem` fails to deserialize -> `accumulator.rs`'s `deserialize_from_value_opt` returns `None` and the entire `mcp_call` disappears. `#[serde(other)] Unknown` doesn't save it — that only catches unknown `type` tags, not a field that fails within a matched variant. So a real tool *failure* is exactly the case that vanishes. Repro fails on the branch:

```rust
#[test]
fn repro_unknown_mcp_error_shape_is_not_dropped() {
    let lines = vec![
        r#"data: {"type":"response.created","response":{"id":"resp_abc"}}"#.to_string(),
        r#"data: {"type":"response.output_item.added","output_index":0,"item":{"type":"mcp_call","id":"mcp_1","server_label":"counter","name":"increment","arguments":"","status":"in_progress","approval_request_id":null,"output":null,"error":null}}"#.to_string(),
        r#"data: {"type":"response.output_item.done","output_index":0,"item":{"type":"mcp_call","id":"mcp_1","server_label":"counter","name":"increment","arguments":"{}","status":"failed","approval_request_id":null,"output":null,"error":{"type":"mcp_protocol_error","code":-32000,"message":"boom"}}}"#.to_string(),
        r#"data: {"type":"response.done","response":{"id":"resp_abc","status":"completed","usage":{"input_tokens":5,"output_tokens":2,"total_tokens":7}}}"#.to_string(),
    ];
    let acc = ResponseAccumulator::from_sse_lines(lines, None);
    assert_eq!(acc.output.len(), 1); // fails: got 0
}
```

Fix: add a catch-all (`Other(Value)`) to `McpCallError`, or make `error: Option<Value>` at the wire boundary. Same remediation family as the `McpCallStatus` ask — worth bundling.

ashwing (commented):
Two accumulator-side drops that the cassette tests miss — both are only reachable through the streaming `from_sse_lines` path, not the `from_json` reconstruction the tests use. I wrote repro tests for both against this branch; they fail. Inline.

The `mcp_call` output-item + error shapes match the recorded OpenAI behavior 1:1, and there's no await-holding-lock on the call path.

Two non-blocking follow-ups:
- The failed path (`status=Failed` -> `mcp_call.failed`) has no synthetic-emit unit test asserting `sequence_number` stays contiguous — only the completed branch does. Given the earlier error-path sequence gaps in #132/#136, worth one.
- The `mcp_list_tools` deferral means the emitted stream is a strict subset of OpenAI's (no list-tools output item before `mcp_call`). Known/accepted — just worth a line in the mcp-gateway doc so clients don't assume it's present.

crates/agentic-server-core/src/types/io/output.rs
maralbahari: Added `McpCallError::Unknown(Value)` as the final untagged fallback, so unrecognized OpenAI error shapes are preserved verbatim without dropping the enclosing `mcp_call`; the SSE regression now covers this case.

crates/agentic-server-core/src/executor/accumulator.rs
maralbahari: great catch. Fixed by adding typed McpCall in-flight handling: it is created on OutputItemAdded, updated via ApplyDone on OutputItemDone, and finalized in insertion order. I kept the pre-existing WebSearch behavior, could you please help to open this bug issue for web search and help to fix as well?

crates/agentic-server-core/tests/cassettes/record_mcp_cassettes.sh
maralbahari: Removed the stale public endpoint default; recording now requires and validates an explicitly trusted MCP URL before writing, with successful recordings atomically replacing existing cassettes.

crates/agentic-server-core/src/types/io/output.rs
maralbahari: Added an optional dedicated `McpCallStatus` covering all documented values, with blocking and streaming regression tests confirming `calling`, `incomplete`, and omitted statuses are preserved.Removed the stale public endpoint default; recording now requires and validates an explicitly trusted MCP URL before writing, with successful recordings atomically replacing existing cassettes.

crates/agentic-server-core/src/tool/mcp/handler.rs
maralbahari: Changed the prefixless fallback to derive a deterministic public MCP ID from the source `id` and `call_id`, so independently constructed started and completed items now reuse one ID across the entire streaming lifecycle; added a prefixless-ID lifecycle test.

maralbahari:
@franciscojavierarceo @ashwing addressed the issues. ready to check again thanks

franciscojavierarceo (commented):
Holding approval on the reproduced done-only MCP ordering bug. The current head otherwise passes the full `agentic-server-core` suite (269 unit tests plus integration/cassette tests), Clippy with warnings denied, formatting, shell syntax, and all GitHub checks.

crates/agentic-server-core/src/executor/accumulator.rs
```diff
@@ -392,6 +393,29 @@ impl ResponseAccumulator {
...
+        if *item_type == SSEItemType::McpCall {
+            if let Some(InFlight::McpCall { item }) = self.in_flight.get_mut(item_id) {
+                item.apply_done(payload, &mut String::new());
+                return;
+            }
+        }
+        if let Some(output_item @ (OutputItem::WebSearchCall(_) | OutputItem::McpCall(_))) =
+            deserialize_from_value_opt::<OutputItem>(raw_item.clone())
+        {
+            self.output.push(output_item);
```
franciscojavierarceo: [P1] Preserve output order when `output_item.added` is omitted

This fallback pushes a done-only MCP item immediately, while earlier reasoning/message items remain in `in_flight` until `response.done`. That reverses valid output order when, for example, reasoning is added at index 0 and an MCP `output_item.done` arrives at index 1 without a preceding added event. The branch explicitly supports done-only MCP events in `test_streaming_preserves_all_documented_mcp_call_statuses`, so the two supported paths need to compose.

I reproduced this on `83e1fd0`: the accumulator returned `[McpCall, Reasoning]` instead of `[Reasoning, McpCall]`. Please retain fallback items by `output_index` and finalize all slots in index order, with a combined regression test.

crates/agentic-server-core/src/tool/mcp/handler.rs
```diff
@@ -393,7 +386,8 @@ fn call_output_id(call: &FunctionToolCall) -> String {
     if let Some(suffix) = call.call_id.strip_prefix("call_").filter(|suffix| !suffix.is_empty()) {
         return format!("mcp_{suffix}");
     }
-    crate::utils::uuid7_str("mcp_")
+    let source_identity = format!("{}\0{}", call.id, call.call_id);
+    format!("mcp_{:016x}", stable_name_hash(&source_identity))
```
franciscojavierarceo, line 390: [P2] Keep ID-less parallel MCP calls distinct

When both provider identifiers are absent, `FunctionToolCall` deserializes them as empty strings and every call hashes the same `"\0"` source identity. Parallel ID-less calls therefore receive the same public MCP item ID. Please allocate the public ID once per call and carry it through the event plan/result, or otherwise give the absent-identity case unique stored state. Add a regression with two calls whose `id` and `call_id` are both omitted.

crates/agentic-server-core/tests/cassettes/mcp/tools.json
```diff
@@ -0,0 +1,9 @@
+[
+  {
+    "type": "mcp",
+    "server_label": "counter",
+    "server_url": "https://gabriel-displays-manufacturers-structures.trycloudflare.com/mcp",
```
franciscojavierarceo: [P2] Keep the tracked reusable fixture on a safe endpoint

The recorder script now correctly requires an explicit operator-trusted endpoint, but the committed reusable fixture still points to a public ephemeral TryCloudflare tunnel with `require_approval: "never"`. This moves the stale/untrusted endpoint problem rather than removing it. Please keep this tracked file on a loopback or non-routable example value and generate the operator-supplied recording config in a temporary file; historical cassette request bodies can retain the recorded URL.

crates/agentic-server-core/src/tool/mcp/handler.rs
franciscojavierarceo: Fixed in vllm-project/agentic-api@44d3bbc: deserialized function calls with a missing or empty provider item ID now receive a unique request-local ID, while started/completed MCP items still reuse the same public ID. Added a parallel ID-less lifecycle regression. Please cherry-pick `44d3bbc`; GitHub rejected a direct push because maintainer edits are disabled.

crates/agentic-server-core/tests/cassettes/mcp/tools.json
franciscojavierarceo: Fixed in vllm-project/agentic-api@44d3bbc: the tracked reusable fixture now uses loopback, and recorder-generated tool configuration lives in a temporary file removed on exit. The negative recorder guard preserves the tracked fixture. Please cherry-pick `44d3bbc`; the source fork does not allow maintainer edits.

crates/agentic-server-core/src/executor/accumulator.rs
franciscojavierarceo: Fixed in vllm-project/agentic-api@44d3bbc: streaming items now retain their `output_index` and are finalized in upstream order. Added a regression covering reasoning at index 0 followed by a done-only MCP call at index 1. The PR source fork disables maintainer edits, so please cherry-pick `44d3bbc` onto this branch.

crates/agentic-server-core/tests/cassettes/mcp/tools.json
maralbahari: @franciscojavierarceo thanksss. I just pushed the cherry pick

franciscojavierarceo (approved):

PR #141, opened 2026-07-22
fix: complete WebSocket shutdown handshake

franciscojavierarceo:
## Summary

- Complete the WebSocket close handshake after draining an active response during shutdown.
- Reject post-shutdown WebSocket requests before they can start inference.
- Make the shutdown regression deterministic with a ping/pong receipt barrier and strict close acknowledgment.

This fixes the Linux CI race observed at caf6bf1 where unread client frames could make socket teardown surface ECONNRESET after the active response completed.

## Test Plan

- `cargo fmt -- --check`
- `cargo test`
- 100 repeated runs of `test_websocket_shutdown_drains_active_response_before_closing`
- gstack `/review` with specialist, adversarial, and structured re-review gates
- Claude read-only worktree review

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,9 @@ build/
 # Rust
 target/
 
+# Local worktrees
+.worktrees/
+
 # IDE
 .idea/
 .vscode/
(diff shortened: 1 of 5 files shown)


maralbahari (approved):
LGTM

Issue #102, opened 2026-07-14
Deploy agentic-api as a remotely persistent, authenticated service


franciscojavierarceo:
## Summary

Make agentic-api easy to deploy as a production container on Kubernetes or a platform-as-a-service.

The deployment should support durable remote conversations and responses across restarts and replicas, with PostgreSQL as the production database and an external OpenAI-compatible/vLLM inference backend.

## Proposed architecture

```
Client
  -> OIDC-authenticated application boundary
  -> agentic-api container
       -> managed PostgreSQL
       -> OpenAI-compatible inference service
```

The gateway remains a single Rust HTTP/WebSocket service. The container should run the gateway only; integrated `serve` mode can remain useful for local development, while hosted deployments point `LLM_API_BASE` at a separately managed inference service.

## Authentication and authorization

Prioritize compatibility with Codex and Claude Code before designing application-issued API-key management.

- Preserve standard `Authorization: Bearer ...` handling.
- Preserve Anthropic-compatible `x-api-key` handling where required.
- Pass client credentials to the configured upstream without inventing a new token format.
- Defer API-key generation, storage, and management until the client integration path is clearer.
- Treat OIDC as a possible future browser/admin or external-auth-proxy concern, not a prerequisite for CLI compatibility.
- Enforce the authenticated identity/tenant on every conversation, response, and item read/write once the identity boundary is defined.

## Work breakdown

- [x] [#105 Containerize agentic-api for production deployment](https://github.com/vllm-project/agentic-api/issues/105)
- [x] [#103 Make PostgreSQL persistence production-ready](https://github.com/vllm-project/agentic-api/issues/103)
- [x] [#104 Add OIDC authentication with GitHub login](https://github.com/vllm-project/agentic-api/issues/104)
- [ ] [#107 Enforce user and tenant ownership for persisted state](https://github.com/vllm-project/agentic-api/issues/107)
- [ ] [#108 Add Kubernetes deployment and operational manifests](https://github.com/vllm-project/agentic-api/issues/108)
- [ ] [#109 Add Render deployment configuration and documentation](https://github.com/vllm-project/agentic-api/issues/109)
- [ ] [#110 Add deployment, authentication, and persistence end-to-end verification](https://github.com/vllm-project/agentic-api/issues/110)

## Current repository anchors

- `DATABASE_URL` and PostgreSQL support already exist in the Rust server/core.
- `/health` and `/ready` endpoints already exist.
[...]

PR #142, opened 2026-07-23
fix: enable production PostgreSQL persistence

### LINKED ISSUE #103: Make PostgreSQL persistence production-ready
Parent: #102

Harden the existing SQLx/PostgreSQL storage path for hosted deployments.

Scope:
- Verify all migrations work on PostgreSQL, including repeatable startup behavior.
- Define migration ownership/startup policy for one or more replicas.
- Add connection-pool, timeout, and TLS configuration guidance.
- Ensure conversation/response/item writes are durable across restart and scale-out.
- Document managed PostgreSQL setup and backup expectations.

Acceptance: a clean PostgreSQL instance can be initialized and used by the gateway with stateful continuation.

See the parent issue: https://github.com/vllm-project/agentic-api/issues/102

franciscojavierarceo:
## Summary

- make the existing persistence layer work with PostgreSQL through portable SQL, configurable pool/timeouts, per-conversation row locking, and a database-unique conversation sequence index
- pin every PostgreSQL pool connection to the validated persistence schema, preserve existing data during the INTEGER-to-BIGINT upgrade, and fail closed on split or incompatible schemas
- verify real writes once at startup, keep readiness probes read-only and privilege-aware, and return consistent machine-readable streaming persistence errors
- add PostgreSQL migration/concurrency/readiness coverage and document production rollout, TLS, pooling, schema, and duplicate-sequence guidance

This is the PostgreSQL persistence slice of #102.

Closes #103.
Refs #102.

## Test Plan

- cargo fmt --all -- --check
- cargo clippy --all-targets -- -D warnings
- cargo test
- uvx pre-commit run --all-files
- TEST_POSTGRES_URL=... cargo test -p agentic-server-core --lib -- --ignored --nocapture
- TEST_POSTGRES_URL=... cargo test -p agentic-server-core --test postgres_storage_integration -- --ignored --nocapture
- read-only Claude review: no actionable findings
- gstack pre-landing specialist, red-team, and adversarial reviews: no remaining findings
- uv run --with-requirements docs/requirements.txt mkdocs build --strict (reaches the pre-existing broken docs/design/claude-code-integration.md link to ROADMAP.md)


crates/agentic-server-core/src/storage/models/response.rs
```diff
@@ -44,7 +44,7 @@ pub async fn create_in_tx(
     sqlx::query_as::<_, Response>(
         "INSERT INTO responses \
          (id, conversation_id, previous_response_id, history_item_ids, metadata, created_at) \
-         VALUES (?, ?, ?, ?, ?, ?) RETURNING *",
+         VALUES ($1, $2, $3, $4, $5, $6) RETURNING *",
```
franciscojavierarceo, line 47: this works for both postgres and sqlite, postgres rejects "?"

crates/agentic-server-core/src/executor/persist.rs
```diff
@@ -23,7 +24,12 @@ pub(crate) async fn persist_if_needed(
     resp_handler: ResponseHandler,
 ) -> ExecutorResult<()> {
     if should_persist(&ctx) {
-        persist_response(payload, ctx, conv_handler, resp_handler).await
+        persist_response(payload, ctx, conv_handler, resp_handler)
+            .await
+            .map_err(|source| {
+                error!(error = ?source, "failed to persist response");
+                crate::executor::error::ExecutorError::Persistence(Box::new(source))
+            })
```
maralbahari, line 32: nit: this line is too long, can import on top of the file `use crate::executor::error{ExecutorError}`

crates/agentic-server-core/src/storage/pool.rs
```diff
@@ -86,6 +86,29 @@ fn sqlite_max_connections(url: &str, config: SqliteConfig) -> u32 {
     }
 }
 
+fn is_postgres_url(url: &str) -> bool {
+    url.starts_with("postgres://") || url.starts_with("postgresql://")
```
maralbahari: Parse the database URL once instead of detecting backends with `starts_with`.
SQLx parses the URL through `url::Url`, which normalizes URI schemes, while these checks are case-sensitive. A valid `POSTGRESQL://... ` connection succeeds but bypasses every `POSTGRES_*`  setting and `after_connect`, leaving statement and lock timeouts unset. SQLite has the same classification problem. The redactor also misses case variants. Introduce a validated DatabaseBackend parsed from `Url::scheme()` and reuse it for configuration, pool construction, diagnostics, and redaction.

crates/agentic-server-core/src/storage/schema.rs
```diff
@@ -3,13 +3,160 @@
...
+        if let Err(error) = configure_postgres_migration_timeout(&mut connection, postgres_migration_timeout).await {
+            let _ = connection.close().await;
+            return Err(error);
+        }
+    }
+
+    let migration_result = sqlx::migrate!("./migrations")
+        .run(&mut *connection)
+        .await
+        .map_err(|error| sqlx::Error::Configuration(error.to_string().into()));
```
maralbahari: Converting MigrateError to `error.to_string()` and wrapping it as `sqlx::Error::Configuration` discards the source chain and means the new `sqlx::Error::Migrate` classification can never handle actual migrator failures. Preserve it with `sqlx::Error::Migrate(Box::new(error))`, or use a project error variant with `#[source]`

maralbahari (commented):
Thanks for the work. I left some inline comments

crates/agentic-server-core/src/storage/models/conversation.rs
```diff
@@ -58,12 +58,39 @@ pub async fn get_or_create(pool: &DbPool, id: &str) -> DbResult<Conversation> {
...
 /// Returns `DbResult::Err` if the database query fails.
 pub async fn get(pool: &DbPool, id: &str) -> DbResult<Option<Conversation>> {
-    sqlx::query_as::<_, Conversation>("SELECT * FROM conversations WHERE id = ?")
+    sqlx::query_as::<_, Conversation>("SELECT * FROM conversations WHERE id = $1")
         .bind(id)
         .fetch_optional(pool)
         .await
 }
 
+/// Locks an existing conversation for the lifetime of the transaction.
```
jiahuei, line 68: The current lock gives a correct append order in storage, but it does not give serial conversation semantics for inference. OpenAI enforces serial turn semantics across the whole operation, not merely append order: a concurrent request receives HTTP 400 and must retry:

```text
openai.BadRequestError: Error code: 400 - {'error': {'message': 'Another process is currently operating on this conversation. Please retry in a few seconds.', 'type': 'invalid_request_error', 'param': 'conversation', 'code': 'conversation_locked'}}
```

In our current implementation, both requests rehydrate the conversation before inference, while the `SELECT ... FOR UPDATE` lock is only acquired during persistence after inference completes. Two concurrent requests can therefore both generate from the same history `H`:

```text
Request A: rehydrate H -> generate A'
Request B: rehydrate H -> generate B'

A persists first: H, A-input, A-output
B acquires the lock next and appends: B-input, B-output
```

The resulting `seq` order is correct and deterministic, but B's output was generated without A's input or output in context. In other words, `seq` records serialized append order, not the complete history observed by the model.

If a conversation is intended to have strictly causal, one-turn-at-a-time semantics, this needs an inference-level concurrency policy: a per-conversation distributed lease or queue spanning rehydrate through persistence. The lease should use a fencing token so an expired holder cannot later persist stale output. Holding this database transaction lock while calling the model would enforce the rule but would unnecessarily hold a database connection and lock for the full inference duration.

While distributed lease could wait till another PR, this PR can instead ensure correctness by adding a UNIQUE constraint and removing this row lock.

crates/agentic-server-core/src/storage/models/response.rs
franciscojavierarceo: Confirmed. SQLx `Any` uses PostgreSQL-style numbered placeholders for both PostgreSQL and SQLite, while PostgreSQL rejects `?`. I’m retaining `$1`; the cross-backend storage tests pass on `2c1ba13`.

crates/agentic-server-core/src/executor/persist.rs
franciscojavierarceo: Fixed in `2c1ba13`: `ExecutorError` is now imported directly and the call site uses the short name.

crates/agentic-server-core/src/storage/pool.rs
franciscojavierarceo: Fixed in `2c1ba13`. A shared validated `DatabaseBackend` now parses and normalizes the URL scheme once, and pool configuration, backend diagnostics, timeout setup, and URL redaction all use it. Regression tests cover uppercase PostgreSQL/SQLite schemes and invalid URLs.

crates/agentic-server-core/src/storage/schema.rs
franciscojavierarceo: Fixed in `2c1ba13`. Migration failures now convert through `sqlx::Error::from`, preserving the typed `Migrate` variant and source chain; the error-classification test covers this path.

crates/agentic-server-core/src/storage/models/conversation.rs
franciscojavierarceo: Agreed that strict causal turn semantics require coordination spanning rehydrate → inference → persistence. I don’t think a UNIQUE constraint plus removal of the row lock makes this PR correct, though: if B computes `MAX(seq)` after A commits, its rows remain unique even though B inferred from stale history, and without the lock concurrent writers can also race on the same next sequence. The persistence lock therefore still protects append integrity, while a separate inference-level lease/queue with fencing is needed to reject or serialize stale turns. The broader concurrency policy should be agreed and implemented as its own focused change rather than implied by a storage-only constraint.

crates/agentic-server-core/src/executor/persist.rs
franciscojavierarceo: Rebase note: the implemented fix is now in [`334072d`](https://github.com/vllm-project/agentic-api/commit/334072d). This supersedes the pre-rebase SHA referenced above.

crates/agentic-server-core/src/storage/pool.rs
franciscojavierarceo: Rebase note: the shared validated `DatabaseBackend` fix and its regression tests are now in [`334072d`](https://github.com/vllm-project/agentic-api/commit/334072d). This supersedes the pre-rebase SHA referenced above.

crates/agentic-server-core/src/storage/schema.rs
franciscojavierarceo: Rebase note: the migration source-preservation fix and test are now in [`334072d`](https://github.com/vllm-project/agentic-api/commit/334072d). This supersedes the pre-rebase SHA referenced above.

crates/agentic-server-core/src/storage/schema.rs
```diff
@@ -3,13 +3,293 @@
...
+         FROM required \
+         WHERE EXISTS ( \
+             SELECT 1 \
+             FROM pg_constraint constraint_metadata \
+             JOIN pg_class table_relation ON table_relation.oid = constraint_metadata.conrelid \
+             JOIN pg_namespace table_namespace ON table_namespace.oid = table_relation.relnamespace \
+             WHERE table_namespace.nspname = current_schema() \
+             AND table_relation.relname = required.table_name \
+             AND constraint_metadata.contype::text = required.constraint_type \
+             AND pg_get_constraintdef(constraint_metadata.oid) = required.definition \
```
ashwing, line 135: `[concern]` `current_schema()`-pinned constraint validation fails to start under a non-`public` search_path.

`postgres_required_constraint_count` matches each constraint by exact string `pg_get_constraintdef(oid) = definition` **and** pins `table_namespace.nspname = current_schema()`, and it runs unconditionally on the auto-migrate path (`widen_postgres_integer_columns` -> `validate_required_postgres_schema`) — i.e. right after sqlx applies our own migrations into `public`.

I reproduced this on `postgres:17-alpine`: an app role with a role-named schema and the **default managed-Postgres** `search_path = "$user", public` makes `current_schema()` return the role's schema, not `public`. The validation then counts **0 of 6** required constraints and returns `Err`, so every replica fails to start even though the schema is correct:

```
as role 'agentic' (search_path="$user",public): current_schema() = agentic
constraint count found under current_schema(): 0    # expected 6
```

(I also checked the PG-version angle — 16 and 17 emit byte-identical `pg_get_constraintdef` output, so formatting drift isn't the issue; the schema-resolution is.)

It fails loud, not silent, so no corruption — but "app deployed with a per-role schema" is a common managed-PG default, not an edge case. Two options: resolve the tables' actual namespace instead of `current_schema()`, or scope the strict constraint-string check to the supervisor/`AGENTIC_API_SCHEMA_READY` path and trust sqlx's migration tracking on auto-migrate.

crates/agentic-server-core/src/storage/models/conversation.rs
```diff
@@ -58,12 +59,39 @@ pub async fn get_or_create(pool: &DbPool, id: &str) -> DbResult<Conversation> {
...
+///
+/// `PostgreSQL` takes a row lock without writing the row. `SQLite` uses a no-op
```
ashwing, line 70: `[nit]` the doc comment says both backends "serialize per-conversation sequence allocation." Postgres does (`SELECT id ... FOR UPDATE`). But the SQLite no-op `UPDATE ... WHERE id = $1` takes SQLite's *database-level* write lock, so it serializes *all* conversation persists globally, not per-conversation. Correct for safety — the granularity claim is just off, and it hides a SQLite write-throughput ceiling worth naming.

ashwing (commented):
The earlier fixes hold up on `2c1ba13` and CI genuinely exercises the Postgres path via the `--ignored` integration run. One schema-validation portability issue I reproduced against a stock Postgres 17, plus an accuracy nit inline — none blocking.

Two more:
- `[nit]` `storage/response.rs` — the row-lock hardening is conversation-scoped only. `previous_response_id` continuations read `get(prev_id).history_item_ids` outside any lock, so two concurrent continuations of the same response read the same history and fork. That's legitimate under OpenAI response-chaining (forks are allowed) and each row is independent — not a bug — but worth stating so nobody assumes response chains are serialized the way conversations now are.
- The two `docs/superpowers/...pr-142-review-followups*.md` files look like PR-process planning docs rather than user-facing deployment docs, and they're not in the mkdocs nav. There's precedent (a websocket-shutdown pair already on main), so if this is a deliberate planning archive that's fine — just checking they're meant to ship, and if so maybe exclude `docs/superpowers/` from the mkdocs build so it doesn't rot in the published docs.

crates/agentic-server-core/src/storage/models/conversation.rs
jiahuei: Hmm even with the lock, if B computes MAX(seq) after A commits, the same issue will happen. At least without the lock, if B computes MAX before A commits, it will fail to persist.

franciscojavierarceo: i split the optimistic version check into #162 so this PR can stay focused on storage integrity. the follow-up captures the conversation’s last sequence with rehydrated history, compares it under the existing row lock before persistence, and returns the OpenAI-compatible `conversation_locked` 400 if the conversation changed. it covers HTTP, SSE, and WebSocket (including `generate: false`) plus SQLite and two-pool PostgreSQL races. the broader pre-inference distributed lease/queue for suppressing external tool side effects remains separate policy work.

jiahuei (approved):

PR #144, opened 2026-07-23
feat: add Responses compaction support

franciscojavierarceo:
## Summary

- add OpenAI-compatible `POST /v1/responses/compact`, reusable compaction items, previous-response rehydration, persistence, and automatic threshold-based context compaction
- preserve canonical user messages, replace prior compaction checkpoints, translate local plaintext `encrypted_content` into assistant context for upstream inference, and accumulate compaction usage across gateway tool loops
- add reproducible OpenAI cassette recordings and offline replay coverage for standalone, repeated, round-trip, and automatic compaction paths using the existing `record_cassette.py` workflow
- preserve structured Responses input, including image content parts, in shared cassette replay helpers
- harden validation, error reporting, storage-disabled behavior, and compatibility with the current request schema

## Test Plan

- `cargo test`
- `cargo fmt -- --check`
- `uv tool run pre-commit run --all-files`
- smoke-tested `record_cassette.py` structured-input loading against `tests/cassettes/compaction/inputs/basic.json`
- verified the four OpenAI compaction cassette replay scenarios within the full Rust test suite


crates/agentic-server-core/src/executor/engine.rs
```diff
@@ -488,12 +497,18 @@ impl ExecuteRequest {
             tools = self.payload.tools.as_ref().map_or(0, Vec::len),
             "executor received responses request"
         );
-        let ctx = rehydrate_conversation(self.payload, &self.exec_ctx).await?;
+        let mut ctx = rehydrate_conversation(self.payload, &self.exec_ctx).await?;
+        let compaction_usage = maybe_compact_context(&mut ctx, &self.exec_ctx, self.client_auth.as_deref()).await?;
```
maralbahari: Could you clarify the intended scope of automatic compaction? Should `compact_threshold` be evaluated only once before the initial inference, as it is now, or before every inference round in `run_until_gateway_tools_complete`? The latter would also re-evaluate the threshold after gateway tool call outputs are appended to the input, since those outputs may substantially increase the context size.

crates/agentic-server-core/src/executor/compaction.rs
```diff
@@ -0,0 +1,580 @@
...
+                    .and_then(|details| details.reason.clone())
+            })
+            .unwrap_or_else(|| "upstream returned no failure details".to_owned());
+        return Err(ExecutorError::CompactionFailed {
+            status: response.status.clone(),
+            details,
+        });
+    }
+    response_output_text(&response.output)
+        .ok_or_else(|| ExecutorError::InvalidRequest("compaction model returned no summary text".to_owned()))
```
maralbahari: A completed upstream response without summary text is an upstream compaction failure, not an invalid client request. Returning InvalidRequest produces HTTP 400 even though the request was valid. As there is `CompactionFailed` added into `ExecutorError` could return that here? so this is consistently reported as HTTP 502 alongside incomplete or failed summarization responses.

maralbahari (commented):
@franciscojavierarceo Thanks for the work. I have added some inline comment one is to clarify the auto compaction logic and another comment is for incorrect error type mapping.

crates/agentic-server-core/src/types/io/input.rs
```diff
@@ -81,10 +128,67 @@ pub enum InputItem {
...
     Reasoning(ReasoningOutput),
+    #[serde(rename = "compaction")]
+    Compaction(CompactionItem),
     #[serde(other)]
     Unknown,
 }
 
+#[derive(Deserialize)]
+#[serde(tag = "type")]
+enum TaggedInputItem {
```
maralbahari: `TaggedInputItem` duplicates every InputItem variant and then requires another exhaustive conversion back into InputItem. Could we simplify the custom deserializer by inspecting the "type" discriminator directly and deserializing into the corresponding payload type, while treating a missing "type" as the shorthand Message form? That would preserve Responses compatibility without maintaining a mirror enum and From<TaggedInputItem> implementation.

crates/agentic-server-core/src/types/io/input.rs
```diff
@@ -53,6 +61,45 @@ pub struct FunctionToolResultMessage {
     pub output: String,
 }
 
+/// A model-generated function call replayed as Responses input.
+///
+/// Input replay is intentionally more permissive than [`FunctionToolCall`]
+/// output: clients may omit `id` and `status` when passing prior items to a
+/// later request or to the compact endpoint.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct InputFunctionToolCall {
```
maralbahari, line 71: Could one `FunctionToolCall` type represent both wire forms by making `id` and `status` optional, with `new_input` leaving them absent and `as_output` populating them? This would remove `InputFunctionToolCall` and the conversion code. Is there a requirement that `OutputItem::FunctionCall` enforce the presence of these fields at the type level? If so, keeping separate types makes sense; otherwise, a shared type may be simpler.

crates/agentic-server-core/src/types/io/input.rs
franciscojavierarceo: Implemented. The InputItem deserializer now inspects the type discriminator and deserializes directly into the corresponding payload type. A missing type remains the shorthand Message form, and unknown future types remain InputItem::Unknown. I also added a regression ensuring a malformed known function_call is rejected instead of falling through as a shorthand message.

crates/agentic-server-core/src/executor/compaction.rs
franciscojavierarceo: Fixed. A completed response without summary text now returns CompactionFailed with status completed, which maps to HTTP 502 and upstream_error. The new unit regression asserts the error variant, HTTP status, and error code.

crates/agentic-server-core/src/executor/engine.rs
franciscojavierarceo: Confirmed against the OpenAI compaction guide: automatic compaction can trigger during a Responses generation when context crosses the threshold, before inference continues. Because our gateway tool loop represents multiple internal inference rounds within one Responses request, maybe_compact_context now runs at the start of every round. The integration regression begins below the threshold, lets gateway tool output push the context above it, then verifies compaction before the next inference and accumulated usage.

crates/agentic-server-core/src/types/io/input.rs
franciscojavierarceo: There is an output-side invariant, so I kept the types separate. OutputItem::FunctionCall is used throughout accumulation, gateway execution, and event emission with concrete id and status fields. Replayed Responses input permits both fields to be absent, which InputFunctionToolCall represents directly. Sharing the type would make required output state optional and spread validation through those output paths; the explicit conversion keeps that permissive-input to complete-output boundary localized.

maralbahari (approved):

PR #146, opened 2026-07-24
feat: add Kubernetes deployment path

### LINKED ISSUE #108: Add Kubernetes deployment and operational manifests
Parent: #102

Provide a Kubernetes deployment path for the gateway.

Scope:
- Deployment, Service, ConfigMap/Secret examples, and migration strategy.
- Liveness/readiness probes using existing endpoints.
- WebSocket and streaming ingress guidance.
- Resource requests/limits, graceful termination, and replica considerations.
- External inference and managed PostgreSQL configuration.

Acceptance: the documented manifests deploy the gateway and keep it ready only when required dependencies are usable.

See the parent issue: https://github.com/vllm-project/agentic-api/issues/102

franciscojavierarceo:
## Summary

- add a Kustomize-compatible Kubernetes base with replicated gateway pods, probes, resource controls, disruption protection, and default-deny ingress
- make `/ready` check PostgreSQL and the inference service concurrently, fail fast on dependency failure, authenticate upstream health checks, reject redirects, coalesce overlapping probes, and log only readiness transitions
- validate rendered Kubernetes resources in CI with checksum-pinned `kubectl` and `kubeconform`
- document safe Secret creation, PostgreSQL TLS verification, image configuration, authenticated ingress boundaries, rollout behavior, and persistence verification
- advance #102 and close #108

## Test Plan

- `cargo test`
- `uv run --with-requirements docs/requirements.txt mkdocs build`
- `kubectl kustomize deploy/kubernetes | kubeconform -strict -kubernetes-version 1.36.0` (7 valid resources)
- read-only Claude review and gstack pre-landing review, including independent adversarial review

### DIFF
--- a/crates/agentic-server-core/src/readiness.rs
+++ b/crates/agentic-server-core/src/readiness.rs
@@ -22,6 +22,43 @@ fn timeout_error(url: &str, timeout_s: f64) -> Error {
     }
 }
 
+/// Result of a single bounded inference-service health probe.
+#[derive(Debug)]
+pub enum LlmReadiness {
+    Ready,
+    Rejected(reqwest::StatusCode),
+    Unreachable(reqwest::Error),
+    TimedOut,
+}
+
+/// Probe the inference service's `/health` endpoint once.
+///
+/// # Errors
+///
+/// Returns an error when the configured bearer credential cannot be represented
+/// as an HTTP header.
+pub async fn probe_llm_readiness(
+    client: &reqwest::Client,
+    llm_api_base: &str,
+    openai_api_key: Option<&str>,
+    timeout: Duration,
+) -> Result<LlmReadiness, Error> {
+    let base = llm_api_base.trim_end_matches('/');
+    let url = format!("{base}/health");
+    let mut request = client.get(url);
+    if let Some(key) = openai_api_key.map(str::trim).filter(|key| !key.is_empty()) {
+        let value = reqwest::header::HeaderValue::from_str(&format!("Bearer {key}"))?;
+        request = request.header(reqwest::header::AUTHORIZATION, value);
+    }
+
+    Ok(match tokio::time::timeout(timeout, request.send()).await {
+        Ok(Ok(response)) if response.status() == reqwest::StatusCode::OK => LlmReadiness::Ready,
+        Ok(Ok(response)) => LlmReadiness::Rejected(response.status()),
+        Ok(Err(error)) => LlmReadiness::Unreachable(error),
+        Err(_) => LlmReadiness::TimedOut,
+    })
+}
+
 /// Poll LLM `/health` until it responds 200 or the timeout is reached.
 ///
 /// # Errors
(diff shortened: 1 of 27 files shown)


crates/agentic-server-core/src/readiness.rs
```diff
@@ -22,7 +22,44 @@ fn timeout_error(url: &str, timeout_s: f64) -> Error {
     }
 }
 
-/// Poll LLM `/health` until it responds 200 or the timeout is reached.
+/// Result of a single bounded inference-service health probe.
+#[derive(Debug)]
+pub enum LlmReadiness {
```
maralbahari, line 31: since this enum is  an externally visible enum through the public `agentic_core::readiness` module. add 
`#[non_exhaustive]`; otherwise adding a future readiness state becomes a breaking API change.

crates/agentic-server-core/src/readiness.rs
```diff
@@ -22,7 +22,44 @@ fn timeout_error(url: &str, timeout_s: f64) -> Error {
...
+    let base = llm_api_base.trim_end_matches('/');
+    let url = format!("{base}/health");
+    let mut request = client.get(url);
+    if let Some(key) = openai_api_key.map(str::trim).filter(|key| !key.is_empty()) {
+        let value = reqwest::header::HeaderValue::from_str(&format!("Bearer {key}"))?;
+        request = request.header(reqwest::header::AUTHORIZATION, value);
+    }
+
+    Ok(match tokio::time::timeout(timeout, request.send()).await {
+        Ok(Ok(response)) if response.status().is_success() => LlmReadiness::Ready,
```
maralbahari, line 74: Redirects can produce false-positive readiness. `request.send()` uses clients with reqwest’s default redirect-following policy. Consequently, /health returning 302 to a login or generic UI ending in 200 is classified as `LlmReadiness::Ready`. Startup proceeds and Kubernetes adds the pod to endpoints even though the configured health endpoint never succeeded directly.
Use a dedicated readiness client with `reqwest::redirect::Policy::none()` and classify 3xx as rejected. Add a redirect regression test.

crates/agentic-server/src/handler/http/models.rs
```diff
@@ -105,32 +106,70 @@ pub async fn health() -> impl IntoResponse {
...
+        &state.llm_api_base,
+        state.openai_api_key.as_deref(),
+        std::time::Duration::from_secs(2),
+    )
+    .await
+    {
+        Ok(LlmReadiness::Ready) => true,
+        Ok(LlmReadiness::Rejected(status)) => {
+            warn!("LLM backend not ready: {status}");
+            false
```
maralbahari, line 122: logs every failed Kubernetes probe at warn. During an extended outage this repeats every five seconds per replica. Rate-limit or log state transitions, and use structured fields such as `warn!(http.status = %status, ...)` and `warn!(error = ?error, ...)`

crates/agentic-server/tests/kubernetes_manifests_test.rs
```diff
@@ -0,0 +1,240 @@
...
+
+use serde_yml::Value;
+
+fn manifest_path(name: &str) -> PathBuf {
+    PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+        .join("../../deploy/kubernetes")
+        .join(name)
+}
+
+fn parse_manifest(name: &str) -> Value {
```
maralbahari, line 12: only parses YAML and checks selected fields. It cannot detect Kubernetes schema errors or broken Kustomize rendering. CI should render the base and validate it with a Kubernetes-aware schema validator.

crates/agentic-server-core/src/readiness.rs
franciscojavierarceo: Addressed in a30d97f: `LlmReadiness` is now `#[non_exhaustive]`, and downstream matching includes a conservative fallback.

crates/agentic-server-core/src/readiness.rs
franciscojavierarceo: Addressed in a30d97f: startup and recurring readiness use a dedicated `Policy::none()` client. Regression tests cover `/health` redirecting to a 200 login page and assert the gateway remains unready.

crates/agentic-server/src/handler/http/models.rs
franciscojavierarceo: Addressed in a30d97f: per-probe failure details are structured `debug!` events, while a shared tracker emits `warn!`/`info!` only when readiness changes. It also coalesces overlapping public probes onto the last completed result so repeated callers cannot multiply database/upstream work or force healthy pods unready.

crates/agentic-server/tests/kubernetes_manifests_test.rs
franciscojavierarceo: Addressed in a30d97f: CI now installs checksum-pinned kubectl 1.36.2 and kubeconform 0.7.0, renders the Kustomize base with pipefail, and strict-validates Kubernetes 1.36 schemas. Local validation found all 7 resources valid.

maralbahari (approved):

PR #148, opened 2026-07-27
feat: support Codex client-executed tool search

haoshan98:
# Support Codex client-executed tool search for function and namespace discovery

## Summary

This PR fully supports the Codex client-executed tool-search flow covered by the six Codex cassettes. It does not claim complete public Responses API tool-search parity. Codex searches its catalog and executes the selected function; the gateway adapts the protocol, validates call/output correlation, and preserves the flow across transports and response continuation.

- Add typed models for `tool_search`, `tool_search_call`, and `tool_search_output` that preserve extension fields and opaque loaded definitions, including the exact `defer_loading` wire field on function declarations.
- Normalize a client `tool_search` declaration into an ordinary upstream function named `tool_search`, preserving its description and parameter schema and setting `strict: false`.
- Restore a well-formed upstream fallback `function_call(name="tool_search")` as a canonical `tool_search_call`, and preserve the subsequent client `tool_search_output`. Blocking responses, HTTP streaming, and Responses WebSocket mode expose the canonical public items and restore the client-facing `response.tools` declarations, removing `defer_loading` only from functions validated as loaded.
- Promote loaded top-level functions and namespace function members only from a completed `tool_search_call` and a matching completed `tool_search_output` with the same nonempty call ID. Ignore unmatched, incomplete, duplicate, conflicting, and ambiguous promotions; prefer existing top-level function names on collisions.
- Restore the original namespace on dynamically loaded namespace-member calls before returning them to Codex.
- Preserve tool-search items through accumulation, persistence, `store: false` execution, and response continuation.

The supported three-turn client flow is:

1. Codex sends a client-executed `tool_search` declaration and a deferred function inside a Codex namespace. The gateway normalizes the search tool for the upstream provider, then restores the provider's fallback function call as a public `tool_search_call`.
2. Codex returns the matching `tool_search_output`. The gateway validates the completed pair, promotes the selected namespace function for the next inference round, and restores the namespace on the resulting function call.
3. Codex returns the function call output and receives the final assistant message.

The fallback function is private to the gateway-to-upstream boundary. Well-formed fallback calls with nonempty call IDs do not leak through blocking, HTTP streaming, or WebSocket public responses. Malformed fallbacks, including calls with missing or empty call IDs and invalid JSON arguments where applicable, pass through and are outside this PR.

### Responses API compatibility

The implementation aligns with the [OpenAI tool-search guide][tool-search-guide] for this Codex client-executed flow, with these explicit boundaries:

- Hosted or bare `tool_search` remains a native pass-through declaration and depends on upstream provider support. The gateway does not perform hosted discovery.
- MCP `defer_loading` is not supported end to end: `McpToolParam` does not model the field and MCP discovery remains eager. The live MCP-backed fixture is exposed to Codex as a namespace function, so it validates the supported function/namespace path rather than native Responses `type: "mcp"` defer loading.
- Client-output promotion supports top-level functions and namespace function members. It does not support every possible public Responses API loaded-tool type.
- `additional_tools` input items are not modeled.
- A synthesized gateway `tool_search_call` preserves call ID correlation semantics but does not reproduce the `tsc_` item IDs observed from OpenAI.


crates/agentic-server-core/src/executor/accumulator.rs
```diff
@@ -101,11 +105,13 @@ impl ResponseAccumulator {
             response_id,
             conversation_id,
             output: Vec::new(),
+            output_indices: Vec::new(),
             usage: None,
             status: ResponseStatus::InProgress,
             incomplete_details: None,
             error: None,
             in_flight: IndexMap::new(),
+            in_flight_indices: HashMap::new(),
```
maralbahari: why this new `output_indices` and `in_flight_indices` introduced?
the in_flight items are IndexMap to take advantage of both hashmap and resolving the order of items arriving.  I think the original behavior in accumulator should be preserved.

crates/agentic-server-core/src/executor/accumulator.rs
```diff
@@ -267,53 +281,21 @@ impl ResponseAccumulator {
...
-                        item,
-                        text: String::with_capacity(256),
-                    }),
-                    SSEItemType::WebSearchCall | SSEItemType::McpToolCall => None,
-                };
-                if let Some(inflight) = entry {
-                    self.in_flight.insert(item_id.clone(), inflight);
-                }
+            (SSEEventType::OutputItemAdded, payload @ EventPayload::OutputItemAdded { .. }) => {
+                self.begin_output_item(payload);
```
maralbahari: could you please put back the original code instead of refactoring in `self.begin_output_item. Any refactoring should be done in separate PR. it makes it difficult to review the changes only required for PR feature support.

crates/agentic-server-core/src/types/request_response.rs
```diff
@@ -30,7 +33,18 @@ pub struct RequestPayload {
     pub metadata: Option<Value>,
     pub parallel_tool_calls: Option<bool>,
     #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub reasoning: Option<Value>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub prompt_cache_key: Option<String>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
```
maralbahari: are these new fields "reasoning, prompt_cache_key" and "extras" related to the feature support? if not it can be added in separate PR

crates/agentic-server-core/src/executor/accumulator.rs
haoshan98: Restored to original `ResponseAccumulator` structure.

crates/agentic-server-core/src/types/request_response.rs
haoshan98: To improve general Codex/Responses compatibility, will include in another PR.

Issue #107, opened 2026-07-14
Enforce user and tenant ownership for persisted state


franciscojavierarceo:
Parent: #102

Connect authenticated identity to the existing tenant-aware storage model.

Scope:
- Define user, tenant, and membership/role semantics.
- Ensure every conversation, response, and item read/write is constrained by the authenticated principal.
- Prevent ID-based cross-tenant access, including previous-response continuation and conversation creation.
- Add migration/index requirements as needed.
- Add isolation tests for OIDC users and API keys.

Acceptance: a principal cannot read, continue, mutate, or delete another principal's persisted conversation state.

See the parent issue: https://github.com/vllm-project/agentic-api/issues/102

## Enterprise Readiness

Enhancement outcome: authenticated users and API keys can safely use shared stateful deployments.

Additional acceptance criteria:

- [ ] Derive ownership from verified identity or an explicitly trusted deployment boundary; an arbitrary client tenant header cannot choose the owner.
- [ ] Enforce ownership consistently for HTTP and Responses WebSocket continuation, conversation/item operations, compaction, and any persisted tool approval state.
- [ ] Scope reusable tool connections and credentials by the applicable security identity; credentials and content must not cross tenants.
- [ ] Document how existing unowned rows and single-tenant deployments are handled during rollout; test both supported database backends and concurrent access.

Coordinate with #155 for Conversations operations and #145 for approval records. Existing parent #102 remains relevant.

Enterprise Readiness tracker: https://github.com/vllm-project/agentic-api/issues/316

Issue #104, opened 2026-07-14
Add OIDC authentication with GitHub login


franciscojavierarceo:
Parent: #102

Add human-user authentication using OIDC, with GitHub as a supported identity provider.

Scope:
- OIDC discovery/JWKS validation and issuer/audience checks.
- Login/callback/session or bearer-token model appropriate for the API/UI boundary.
- Map stable OIDC subject plus issuer to an application user.
- Support GitHub login through provider configuration rather than GitHub-specific authorization logic.
- Define bootstrap/admin configuration and safe failure behavior.

Acceptance: a valid OIDC identity can call protected API routes and an invalid/expired token is rejected.

See the parent issue: https://github.com/vllm-project/agentic-api/issues/102

PR #149, opened 2026-07-28
feat: add optional OIDC bearer authentication

franciscojavierarceo:
## Summary

- add opt-in OIDC bearer authentication for every defined `/v1/*` HTTP and WebSocket route while keeping `/health` and `/ready` public
- validate asymmetric JWTs through secure discovery and bounded, cached, single-flight JWKS refreshes; preserve protocol-specific OpenAI, Anthropic, and Responses WebSocket error contracts
- enforce `OIDC_AUDIENCE` as the complete audience trust set and validate every presented `azp`
- consume identity credentials at the gateway boundary, keep OpenAI upstream credentials operator-owned, and preserve explicit Anthropic upstream `x-api-key` pass-through
- retain token expiration for long-lived WebSocket sessions and emit the canonical Responses error event after expiry
- document server configuration plus Codex and Claude Code client setup
- document a tested GitHub OAuth -> Dex OIDC -> agentic-api deployment flow, including PKCE login and production guidance

This is the first focused slice of #104 under the #102 authentication umbrella. Tenant and persisted-state ownership remain intentionally tracked by #107.

Refs #102
Refs #104

## Test Plan

- `cargo test`
- `uv run --with-requirements docs/requirements.txt mkdocs build`
- live local GitHub OAuth login through Dex v2.45.1 using authorization code with PKCE
- unauthenticated and authenticated `/v1/models` checks, plus authenticated `/v1/responses/compact`
- upstream credential-separation check proving the OIDC ID token was not forwarded
- regression coverage for strict audience/`azp` validation, cancellation-safe and bounded JWKS refreshes, Anthropic request IDs, OpenAI/Anthropic upstream credential policy, and post-upgrade WebSocket expiry
- read-only Claude review: no actionable findings
- gstack testing, maintainability, security, performance, migration, and red-team passes; all actionable findings fixed and final reruns clean

The optional gstack Codex export was not run because the local approval policy blocked sending the repository diff to an external service.

### DIFF
--- a/crates/agentic-server/src/auth.rs
+++ b/crates/agentic-server/src/auth.rs
@@ -1,0 +1,837 @@
+use std::collections::HashMap;
+use std::sync::Arc;
+use std::time::Duration;
+use std::time::Instant;
+
+use axum::body::Body;
+use axum::extract::State;
+use axum::http::{HeaderMap, Request, StatusCode, header};
+use axum::middleware::Next;
+use axum::response::Response;
+use jsonwebtoken::jwk::{Jwk, JwkSet, KeyAlgorithm, KeyOperations, PublicKeyUse};
+use jsonwebtoken::{Algorithm, DecodingKey, Validation, decode, decode_header};
+use serde::Deserialize;
+use serde_json::json;
+use tokio::sync::RwLock;
+use tracing::{debug, warn};
+use url::{Host, Url};
+
+const OIDC_HTTP_TIMEOUT: Duration = Duration::from_secs(10);
+const JWKS_REFRESH_COOLDOWN: Duration = Duration::from_secs(30);
+const JWKS_REFRESH_COALESCE_WINDOW: Duration = Duration::from_secs(1);
+const DEFAULT_JWKS_TTL: Duration = Duration::from_secs(300);
+const MAX_JWKS_TTL: Duration = Duration::from_secs(3600);
+const MAX_PROVIDER_RESPONSE_BYTES: usize = 1024 * 1024;
+const MAX_JWKS_KEYS: usize = 100;
+const JWT_CLOCK_SKEW_SECONDS: u64 = 60;
+
+pub(crate) const ANTHROPIC_MESSAGES_PATH: &str = "/v1/messages";
+pub(crate) const ANTHROPIC_COUNT_TOKENS_PATH: &str = "/v1/messages/count_tokens";
+
+#[derive(Clone)]
+pub struct OidcConfig {
+    issuer: Url,
+    issuer_value: String,
+    audience: String,
+    allows_loopback_http: bool,
+}
+
+impl OidcConfig {
+    /// Create an OIDC bearer-token configuration.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error when the issuer is not an absolute HTTPS URL (except
+    /// for loopback test/development issuers) or the audience is empty.
+    pub fn new(issuer: &str, audience: &str) -> Result<Self, OidcAuthError> {
+        let issuer_value = issuer.trim().to_owned();
+        let issuer = Url::parse(&issuer_value).map_err(OidcAuthError::InvalidIssuer)?;
+        let allows_loopback_http = is_loopback_http(&issuer);
+        if issuer.scheme() != "https" && !allows_loopback_http {
+            return Err(OidcAuthError::InsecureIssuer);
+        }
+        if issuer.query().is_some() || issuer.fragment().is_some() {
+            return Err(OidcAuthError::InvalidIssuerComponents);
+        }
+
+        let audience = audience.trim().to_owned();
+        if audience.is_empty() {
+            return Err(OidcAuthError::EmptyAudience);
+        }
+
+        Ok(Self {
+            issuer,
+            issuer_value,
+            audience,
+            allows_loopback_http,
+        })
+    }
+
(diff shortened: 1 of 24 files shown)


crates/agentic-server/src/handler/websocket/responses.rs
```diff
@@ -78,6 +102,11 @@ async fn responses_ws_loop(socket: WebSocket, state: AppState, headers: HeaderMa
             }
         };
 
+        if let Some(error) = websocket_identity_error(principal.as_ref()) {
+            let _ = send_ws_error(&mut sender, &error).await;
```
maralbahari: routes expiration through `WsError::to_ws_frame`, which produces `{type,status,error:{...}}`. The official Responses WebSocket schema instead requires top-level code, `message`, `param`, and `sequence_number`. `Codex-compatible` clients therefore may not recognize the expiration or reconnect with the refreshed credential. Emit a proper `ResponseErrorEvent` before closing the connection. [OpenAI WebSocket error event](https://developers.openai.com/api/reference/resources/responses/websocket-events#error)

crates/agentic-server/src/auth.rs
```diff
@@ -0,0 +1,837 @@
...
+            }
+            if refresh_state.retry_after.is_some_and(|deadline| now < deadline) {
+                return Err(OidcAuthError::JwksRefreshBackoff);
+            }
+        }
+
+        self.refresh_state
+            .lock()
+            .map_err(|_| OidcAuthError::RefreshStateUnavailable)?
+            .retry_after = Some(Instant::now() + JWKS_REFRESH_COOLDOWN);
```
maralbahari: sets `retry_after` before awaiting the provider request. If that future is dropped. For example, after a client disconnect—the cleanup and successful-state update never run, but subsequent requests still receive `JwksRefreshBackoff` and 503 responses for 30 seconds. Make the refresh cancellation-safe, such as with a drop guard or detached shared refresh task.

crates/agentic-server/src/auth.rs
```diff
@@ -0,0 +1,837 @@
...
+                "param": null,
+                "code": code
+            }
+        }),
+        AuthErrorFormat::Anthropic => json!({
+            "type": "error",
+            "error": {
+                "type": anthropic_error_type,
+                "message": message
+            }
```
maralbahari: creates an Anthropic-shaped body without `request_id`, while the response builder emits no `request-id` header. Anthropic documents a unique request-id header on every response and the same value in error bodies, which SDKs expose for diagnostics. Generate one identifier and include it in both locations. [Anthropic error and request-ID contract](https://platform.claude.com/docs/en/api/errors)

crates/agentic-server/src/auth.rs
```diff
@@ -0,0 +1,837 @@
...
+}
+
+impl IdentityClaims {
+    fn audience_allows(&self, expected: &str) -> bool {
+        match &self.aud {
+            AudienceClaim::One(audience) => audience == expected,
+            AudienceClaim::Many(audiences) if audiences.len() == 1 => audiences[0] == expected,
+            AudienceClaim::Many(audiences) => {
+                audiences.iter().any(|audience| audience == expected) && self.azp.as_deref() == Some(expected)
+            }
```
maralbahari, line 720: accepts ["agentic-api", "<anything>"] whenever azp equals agentic-api. OIDC requires rejecting ID tokens containing additional audiences that the client does not trust, but the gateway has no configuration declaring those additional values trusted. Reject multi-audience tokens unless every audience is explicitly allowlisted; also validate azp whenever it is present.

maralbahari (commented):
Thanks for the PR. left some inline comments

crates/agentic-server/src/handler/websocket/responses.rs
franciscojavierarceo: Fixed in c25567f. OIDC expiry now emits the Responses WebSocket top-level error event with type, code, message, param, and sequence_number. Generic WebSocket handler errors keep the existing nested envelope. Covered by websocket_identity_expiry_uses_responses_error_event and the live authenticated_websocket_rejects_requests_after_identity_expiry integration test.

crates/agentic-server/src/auth.rs
franciscojavierarceo: Fixed in c25567f. A scoped RefreshAttempt guard clears only its own provisional retry deadline when the refresh future is cancelled, while real provider failures retain the 30-second backoff. Refresh-gate waiting is also bounded to one second. Covered by cancelled_jwks_refresh_does_not_install_backoff, stalled_jwks_refresh_bounds_waiters, and fresh_cache_unknown_key_refreshes_once_after_cooldown.

crates/agentic-server/src/auth.rs
franciscojavierarceo: Fixed in c25567f. Anthropic-shaped authentication and authentication-dependency errors now generate one req_-prefixed UUIDv7 and return the same value in the request-id header and top-level request_id body field. Covered by anthropic_authentication_errors_include_matching_request_id.

crates/agentic-server/src/auth.rs
franciscojavierarceo: Fixed in c25567f. OIDC_AUDIENCE is now the complete trust set: every audience entry must equal the configured audience, and any present azp must also match, including scalar-audience tokens. The API, design, deployment docs, and negative integration cases were updated.

maralbahari (approved):

Issue #116, opened 2026-07-14
[Claude Code - Stage 3] Parity hardening (cassettes + edge cases)


ashwing:
Part of #113 (Claude Code integration).

Extend the Messages cassette matrix beyond Stage 2's single acceptance recording (same recorder approach as the Responses cassettes in #94) to the edge cases:

- [ ] Interleaved `thinking` + `tool_use` in one turn.
- [ ] Multi-round gateway tools (model calls a gateway tool on consecutive turns).
- [ ] Streaming `tool_use` (`input_json_delta` partial-JSON args).
- [ ] `system` prompt handling (attribution block + multi-block `system`).
- [ ] `cache_control` fidelity across transparent proxying and every server-side tool-loop round: retain each supported marker, TTL, and original block/tool boundary so cached-prefix composition is unchanged.

Session identity normalization and downstream routing metadata are tracked separately in #159.

Follows #115 (which lands the recorder + the first gateway-tool cassette).

ashwing:
All four parity items are now covered:

- ✅ interleaved `thinking` + `tool_use` in one turn — `repro_f3_next_round_preserves_thinking_and_text_blocks`
- ✅ multi-round gateway tools — `messages_loop_multi_round_sequential` + `messages_loop_parallel_tool_use` (replay the live-recorded sequential/parallel cassettes)
- ✅ streaming `input_json_delta` — `messages_stream_*` tests over the streaming cassette
- ✅ multi-block `system` attribution — added in #154 (`messages_loop_preserves_multi_block_system_across_rounds`)

On the `pause_turn` refinement raised in the #115 thread: it's a non-issue for the vLLM backend. `stop_reason` on the Anthropic Messages path is a closed `Literal["end_turn", "max_tokens", "stop_sequence", "tool_use"]` in `vllm/entrypoints/anthropic/protocol.py`, and `serving.py` only ever emits `end_turn`/`max_tokens`/`tool_use` — `pause_turn` isn't representable, so vLLM can't produce it. The loop's existing `stop_reason != "tool_use"` terminal branch already handles every value the upstream can return. If a future backend does emit `pause_turn`, we can revisit then; there's nothing to build against today.

Closing once #154 merges.

PR #154, opened 2026-07-30
test: cover multi-block system preservation across the Messages gateway loop

ashwing:
## Summary

Adds a regression guard that a multi-block Anthropic `system` field survives the `/v1/messages` gateway tool loop unchanged across rounds.

The loop rebuilds the upstream request body every round — `append_round_to_history` appends the assistant turn + `tool_result` to `messages`. It only touches `messages`, so `system` should ride through untouched, but the loop path had no coverage of that (only the transparent-proxy path asserted `system` forwarding). Claude Code sends `system` as an attribution block + instructions, so this is the shape that matters.

The test drives a two-round gateway loop with a multi-block `system` and asserts both upstream request bodies carry the identical blocks — so a future refactor that rebuilds the upstream body from a curated field subset can't silently drop `system`.

Completes the Stage 3 (#116) system-attribution parity item. Part of #113.

## Test Plan

- `cargo test -p agentic-server-core --test messages_loop_test --test messages_stream_test` (15 tests, including non-streaming and streaming multi-block `system` preservation)
- `cargo fmt -- --check`

### DIFF
--- a/crates/agentic-server-core/tests/messages_loop_test.rs
+++ b/crates/agentic-server-core/tests/messages_loop_test.rs
@@ -516,3 +516,46 @@ async fn messages_loop_handles_malformed_tool_input() {
     assert_eq!(calls, 2, "loop still ran the tool round + final");
     assert_eq!(result["stop_reason"], "end_turn");
 }
+
+// #116 (Stage 3 parity): a multi-block `system` (attribution block + instructions,
+// the shape Claude Code sends) must survive the gateway tool loop unchanged. The
+// loop only appends the assistant turn + tool_result to `messages`
+// (`append_round_to_history`) and must never touch `system` — so round 2's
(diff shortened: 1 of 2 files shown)


franciscojavierarceo (commented):
The new regression guard covers only the non-streaming `run_messages_loop` path. Claude Code sends `stream: true`, and `/v1/messages` routes those requests through the independent `run_messages_stream` implementation and its own round-history mutation. Without a matching request-body assertion in `messages_stream_test.rs`, a streaming-only refactor could still drop or reshape the multi-block `system` field while this test remains green.

I’m pushing the corresponding streaming request-capture regression test to this branch before merge.

franciscojavierarceo (approved):
Verified the non-streaming and streaming multi-block `system` regression coverage, including a mutation check against the streaming path. Local `cargo test`, formatting, and Clippy all pass.

Issue #152, opened 2026-07-30
[Reasoning retention 1/2] Reconcile complete streamed reasoning items


franciscojavierarceo:
Parent: #151

## Goal

Make the streaming path produce the same complete typed reasoning items as the non-streaming path, without changing the database schema or persistence unit.

### Event normalization

- Give `response.reasoning_text.delta/done` and `response.reasoning_summary_text.delta/done` distinct normalized payloads.
- Preserve the item ID and any index needed to update the correct in-flight field.
- Keep reasoning text and reasoning summaries semantically separate throughout normalization and accumulation.

### Streaming accumulator

- Track independent fallback state for reasoning `content` and `summary`.
- Parse reasoning items from `response.output_item.done.item`.
- Reconcile a valid completed item with matching in-flight state by item ID.
- Treat raw field presence as significant:
  - completed fields that are present—including empty arrays or `null`—are authoritative;
  - delta-derived state is used only when the corresponding completed field is omitted.
- Preserve `id`, all typed `content` entries, all `summary` entries, `encrypted_content`, and `status`.
- If `output_item.added` is absent, retain the completed item at its `output_index`.
- If the completed item is malformed, retain valid delta-derived state.
- Emit at most one reasoning item per ID and preserve ordering relative to messages and tool calls.
- Follow the existing `complete_custom_tool_call` reconciliation structure where practical.

### Storage contract

- Keep `ReasoningOutput` as the persistence unit.
- Add no reasoning-text table, column, migration, or alternate flattened representation.
- Verify a complete reasoning item survives storage serialization and deserialization with its private item-kind marker removed from the reconstructed typed item.

## Test plan

Add focused tests to the existing accumulator, normalizer, and storage test modules:

- [ ] streamed reasoning-text deltas followed by a complete reasoning item;
- [ ] streamed summary deltas preserved separately from reasoning text;
- [ ] `encrypted_content` and `status` copied from `output_item.done`;
- [ ] completed content/summary overriding incomplete delta-derived fields;
- [ ] delta fallback when completed content or summary is omitted;
- [ ] completed reasoning without a preceding `output_item.added`;
- [ ] no duplicate when deltas and a completed item are both present;
- [ ] malformed completed item retaining valid accumulated reasoning;
- [ ] original `output_index` ordering when an added event is omitted;
- [ ] complete typed storage serialization/deserialization round trip.

Use literal hand-checked fixtures. Do not hand-author or update captured replay cassettes for these unit-level cases.

[...]

PR #158, opened 2026-07-31
fix: align custom tool SSE lifecycle with OpenAI API compatibility

maralbahari:
## Summary

This PR fixes custom tool support across the gateway while preserving OpenAI’s public Responses API contract.

## What was wrong

vLLM accepts a `type: "custom"` declaration, but exposes the resulting call as a regular `function_call`. Its streaming response emits:

- `response.function_call_arguments.delta`
- `response.function_call_arguments.done`

It does not emit OpenAI-compatible custom-tool events:

- `response.custom_tool_call_input.delta`
- `response.custom_tool_call_input.done`

Forwarding vLLM’s response directly therefore leaked the internal normalized function shape instead of returning a public `custom_tool_call`.

## What changed

The gateway now:

- Normalizes custom tools into model-facing function tools with a required string `input` parameter.
- Keeps custom tools client-owned; the gateway does not execute them.
- Hides vLLM’s internal normalized function-call events.
- Restores blocking output as `custom_tool_call`.
- Emits the OpenAI-compatible streaming lifecycle:
  - `response.output_item.added`
  - `response.custom_tool_call_input.delta`
  - `response.custom_tool_call_input.done`
  - `response.output_item.done`
- Uses one stable `ctc_` item ID throughout the lifecycle.
- Converts custom call/output history into the normalized function representation only when sending continuation context upstream.

## Testing

Added OpenAI and gateway cassettes for streaming and non-streaming two-turn custom-tool flows. Tests verify:

- Gateway output matches OpenAI’s public custom-tool contract.
- Raw custom input is preserved.
- Internal `function_call` items do not leak.
- Streaming lifecycle IDs and sequence numbers remain consistent.
- `custom_tool_call_output` works correctly on the continuation turn.


crates/agentic-server-core/src/tool/custom.rs
```diff
@@ -0,0 +1,183 @@
...
+pub struct CustomHandler;
+
+impl CustomHandler {
+    #[must_use]
+    pub fn to_function_call(param: &CustomToolParam) -> FunctionTool {
+        FunctionTool {
+            type_: "function".to_owned(),
+            name: param.name.as_str().to_owned(),
+            description: param.description.clone(),
+            parameters: Some(serde_json::json!({
```
franciscojavierarceo, line 75: custom-tool grammar constraints are silently dropped here. every custom tool becomes a function that accepts an arbitrary string, while `format` and extension fields are ignored. CFG and regex tools can receive invalid input while still looking supported. we should preserve or enforce these constraints, or reject formats we can’t support.

crates/agentic-server-core/src/types/io/input.rs
```diff
@@ -254,6 +276,23 @@ impl ResponsesInput {
...
+    if let Some(suffix) = item_id.strip_prefix("ctc_").filter(|suffix| !suffix.is_empty()) {
+        return Some(format!("fc_{suffix}"));
+    }
+    Some(item_id.to_owned())
+}
+
+fn custom_output_text(output: Value) -> String {
+    match output {
+        Value::String(output) => output,
+        output => output.to_string(),
```
franciscojavierarceo: structured custom-tool outputs get stringified here. an image or file content array becomes JSON text like `"[{...}]"`, so the upstream loses the media semantics. we should preserve the string-or-content-array shape and reject unsupported values.

crates/agentic-server-core/src/executor/accumulator.rs
```diff
@@ -457,6 +422,17 @@ impl ResponseAccumulator {
...
+        match (item_type, self.in_flight.get_mut(item_id).map(|entry| &mut entry.item)) {
+            (SSEItemType::CustomToolCall, Some(InFlight::CustomToolCall { item, input })) => {
+                item.apply_done(payload, input);
+                return;
+            }
+            (SSEItemType::McpCall, Some(InFlight::McpCall { item })) => {
+                item.apply_done(payload, &mut String::new());
+                return;
+            }
+            _ => {}
```
franciscojavierarceo: `output_item.done` can’t restore an initially unnamed function call. if the name and call ID only arrive in the done event, the streaming path recognizes and hides the internal custom call, but the accumulator keeps the empty metadata. the completed response then leaks a broken plain `function_call`. we should apply the authoritative done item to function calls too.

crates/agentic-server-core/src/types/request_response.rs
```diff
@@ -152,8 +112,8 @@ impl RequestPayload {
             .as_deref()
             .map(|tools| CodexNamespaceHandler.resolve_namespace_members(tools))
             .transpose()?;
-        let tools: Option<Vec<UpstreamTool>> =
-            renamed_tools.map(|tools| tools.into_iter().flat_map(upstream_tools).collect());
+        let tools: Option<Vec<FunctionTool>> =
+            renamed_tools.map(|tools| tools.iter().flat_map(ResponsesTool::to_function_tools).collect());
```
franciscojavierarceo: the initial streaming metadata exposes the internal function conversion. `response.created` and `response.in_progress` describe a generic function, then later events switch back to `custom_tool_call`. we should rewrite `tools` and `tool_choice` back to the public custom-tool shape before sending those frames.

franciscojavierarceo (commented):
left a few inline comments.

crates/agentic-server-core/src/executor/upstream.rs
```diff
@@ -374,23 +374,29 @@ fn should_hide_upstream_event(
...
-        ) => hidden_gateway_item_ids.contains(item_id),
+        ) => hidden_public_call_item_ids.contains(item_id),
         _ => false,
     }
 }
 
+fn uses_public_call_shape(registry: &ToolRegistry, name: &str) -> bool {
+    registry
+        .lookup(name)
+        .is_some_and(|entry| entry.tool_type == ToolType::Custom || entry.tool_type.is_gateway_owned())
```
franciscojavierarceo: once a custom call appears, the rest of the upstream stream gets buffered until inference finishes. this turns the input delta into one full-input event, delays later message and reasoning events, and leaves us with an unbounded buffer. we should emit normalized custom deltas incrementally and bound any remaining deferred storage.

crates/agentic-server-core/src/types/request_response.rs
```diff
@@ -49,12 +49,11 @@ pub struct UpstreamRequest<'a> {
...
     pub instructions: Option<&'a str>,
-    /// Tools forwarded to vLLM. Namespace members are flattened to ordinary
-    /// function declarations; native custom declarations retain their freeform
-    /// wire shape.
+    /// Tools forwarded to vLLM. Function-like declarations are normalized to
+    /// ordinary function tools.
     /// Skipped when empty so vLLM does not receive an empty array.
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub tools: Option<Vec<UpstreamTool>>,
+    pub tools: Option<Vec<FunctionTool>>,
```
franciscojavierarceo: this breaks the core crate’s public Rust API: `UpstreamRequest.tools` changes type, `UpstreamTool` disappears, and `ToolChoice::Custom` is removed. i confirmed that existing downstream code no longer compiles. we should keep the public types and do the normalization at the upstream serialization boundary.

crates/agentic-server-core/src/executor/accumulator.rs
```diff
@@ -490,13 +466,9 @@ impl ResponseAccumulator {
             }
             return;
         }
-        if let (SSEItemType::McpCall, Some(InFlight::McpCall { item })) =
-            (item_type, self.in_flight.get_mut(item_id).map(|entry| &mut entry.item))
+        if let Some(output_item @ (OutputItem::CustomToolCall(_) | OutputItem::McpCall(_))) =
```
ashwing: This refactor removes the dedicated `complete_custom_tool_call` — which had an explicit branch + comment for providers that omit `output_item.added` — and folds the custom done-only case into this shared `CustomToolCall | McpCall` catch-all. The only custom accumulator unit test (`test_custom_tool_call_accumulates_freeform_input`) sends `output_item.added` → done, so the done-only path this catch-all now owns is untested.

That's the exact bug class #139 just shipped a fix + regression for on the MCP side (done-only ordering/drop). Worth a `from_sse_lines` unit test: a `custom_tool_call` `output_item.done` at `output_index: 1` with **no** preceding `output_item.added`, plus a reasoning item parked in-flight at index 0 — asserting the custom item finalizes *after* reasoning in output order. Cheap, and it locks the seam #152 is about to build on.

crates/agentic-server-core/src/tool/custom.rs
```diff
@@ -0,0 +1,183 @@
...
+    value.as_bytes().iter().fold(0xcbf2_9ce4_8422_2325_u64, |hash, byte| {
+        (hash ^ u64::from(*byte)).wrapping_mul(0x0000_0100_0000_01b3)
+    })
+}
+
+pub(crate) fn input_from_arguments(arguments: &str) -> String {
+    try_input_from_arguments(arguments).unwrap_or_else(|| arguments.to_owned())
+}
+
+pub(crate) fn try_input_from_arguments(arguments: &str) -> Option<String> {
```
ashwing, line 297: This best-effort heuristic recovers the raw custom input from a `{"input": <string>}` envelope, and the `unwrap_or_else(|| arguments.to_owned())` fallback forwards the raw JSON verbatim on any miss — extra field (`fields.len() == 1` fails), non-string `input`, or unparseable args.

Custom tools are client-executed (apply_patch et al.) with freeform payloads, so a miss silently hands the client `{"input": "..."}` instead of the patch blob. `strict: true` + `additionalProperties: false` make the model *usually* comply, but this runs on both the streaming (`CustomHandler::output_item` → synthesized input events) and blocking paths, so a mismatch corrupts input identically in both. At minimum a `tracing::debug!` on the fallback branch — a leaked envelope is otherwise invisible until a client tool breaks on it.

ashwing (commented):
Focused on the accumulator/normalizer seam since the client-contract and public-API surface is already well covered. Two things there — a coverage gap on the done-only path and the input-extraction heuristic.

crates/agentic-server-core/src/tool/custom.rs
maralbahari: Custom-tool format constraints and extension fields are now preserved in the normalized model-visible description, including complete Lark and regex definitions and explicit raw-input instructions, rather than being silently dropped.

crates/agentic-server-core/src/types/io/input.rs
maralbahari: Added a typed `ToolCallOutput` matching OpenAI’s documented string-or-content-array contract. Structured outputs now preserve `input_text`, `input_image`, and `input_file` content instead of being JSON-stringified, while unsupported shapes are rejected. See OpenAI’s [function-calling](https://developers.openai.com/api/docs/guides/function-calling),  [image-input](https://developers.openai.com/api/docs/guides/images-vision), and [file-input](https://developers.openai.com/api/docs/guides/file-inputs) documentation.

crates/agentic-server-core/src/executor/accumulator.rs
maralbahari: Fixed by adding the missing FunctionCall arm so output_item.done applies its authoritative name, call ID, and arguments to the matching in-flight function call before finalization.

crates/agentic-server-core/src/types/request_response.rs
maralbahari: Addressed using the same request-scoped mapping pattern as Codex namespaces. `ToolRegistry` builds a `custom_tool_map` alongside `namespace_map`; before emitting `response.created` and `response.in_progress`, `restore_stream_event_wire` uses it to restore normalized function declarations, forced choices, and `allowed_tools` selectors to their public `type: "custom"` shapes. This keeps initial metadata consistent with the later `custom_tool_call` lifecycle.

crates/agentic-server-core/src/executor/accumulator.rs
maralbahari: The shared done-only fallback already preserves custom calls when `output_item.added` is omitted. I added a regression test with preceding in-flight reasoning to verify the custom call is retained and finalized in `output_index` order.

crates/agentic-server-core/src/tool/custom.rs
maralbahari: Addressed by logging a debug event whenever custom arguments do not match the expected `{"input":"..."}` envelope and raw arguments are forwarded. The log includes only the payload byte length to avoid exposing client tool input.

Reviewer reply:
Added a `tracing::debug!` on the raw-argument fallback path, making envelope recovery failures visible in both streaming and blocking flows without logging the potentially sensitive custom-tool payload.

crates/agentic-server-core/src/executor/upstream.rs
maralbahari: @franciscojavierarceo  Fixed by adding a function SSE translator that runs immediately after `ResponseAccumulator` updates the authoritative in-flight function call. Normalized custom-tool argument deltas are now converted and emitted incrementally as `custom_tool_call_input.delta`, without delaying later stream events. Only unnamed calls are temporarily buffered, with a 256 KiB limit. The re-recorded OpenAI and gateway cassettes both emit the same six-delta lifecycle, and the integration test now compares the exact delta sequence, lifecycle ordering, stable item ID, and completed output.

maralbahari:
@franciscojavierarceo @ashwing 
Thanks for the review. it's ready to review again.
The changes after addressing the comments are as following:

- **Custom-tool grammar handling:** Preserved custom `format` constraints in the normalized function description so the model sees the original grammar or regex requirements instead of receiving an unconstrained string tool.

- **Structured tool outputs:** Added typed handling for text and content-array outputs, preserving OpenAI image/file content semantics and rejecting unsupported output shapes instead of stringifying them.

- **Public Rust API compatibility:** Restored `UpstreamTool` while retaining function-only upstream serialization, since custom tools are normalized to functions before being sent upstream.

- **OpenAI tool-choice contract:** Restored support for `auto`, `none`, `required`, forced function/custom choices, and typed `allowed_tools` entries.

- **Streaming response metadata:** Added a custom-tool name map, following the namespace restoration design, so `response.created` and `response.in_progress` expose the original custom tools and tool choices rather than internal normalized functions.

- **Authoritative done events:** Function calls now apply the complete `output_item.done` payload, preventing initially unnamed calls from leaking incomplete IDs, names, or arguments.

- **Malformed custom envelopes:** Added debug logging when normalized custom arguments cannot be unwrapped, making any raw-envelope fallback observable.

- **Done-only streaming regression:** Added coverage for custom calls that receive `output_item.done` without a preceding `output_item.added`, including correct ordering after reasoning items.

- **Incremental custom-tool SSE:** Added a generic function SSE translator that uses the accumulator’s authoritative in-flight state and converts normalized function deltas into incremental `custom_tool_call_input.delta` events without delaying subsequent stream events.

- **Bounded buffering:** Only unnamed function calls are temporarily buffered, with a 256 KiB limit; ordinary functions pass through and gateway-owned MCP/web-search calls retain their execution-owned lifecycle.

- **OpenAI cassette verification:** Re-recorded two-turn streaming and non-streaming cassettes for both OpenAI and the gateway. The streaming test now compares the exact six-delta sequence, lifecycle ordering, stable public item ID, completed item, and `custom_tool_call_output` continuation.

crates/agentic-server-core/src/executor/upstream.rs
maralbahari: @franciscojavierarceo 
Function-call classification and wire conversion were moved out of `executor/upstream.rs` into a dedicated `FunctionSseTranslator`. After `ResponseAccumulator` updates the authoritative internal function call, the translator uses a request-scoped `name → ToolType` map to pass through normal functions, incrementally translate custom-tool events, and suppress gateway-owned MCP/web-search frames until execution emits their public lifecycle. This replaces the previous hide, buffer, flush, and function-call classification helpers scattered throughout `executor/upstream.rs`; only initially unnamed calls are buffered, with a 256 KiB limit.

these changes made the PR  pushed out of the scope but it's cleaner and helps in up coming tool type support.

crates/agentic-server-core/src/executor/function_sse.rs
```diff
@@ -0,0 +1,606 @@
...
+/// model-visible name-to-type mapping.
+#[derive(Debug, Default)]
+pub(super) struct FunctionSseTranslator {
+    tool_types: HashMap<String, ToolType>,
+    active: HashMap<String, FunctionCallShape>,
+    pending_unnamed: HashMap<String, PendingFunctionCall>,
+    pending_bytes: usize,
+}
+
+impl FunctionSseTranslator {
```
maralbahari, line 64: Function-call SSE translation refactor

- Detecting normalized function calls that represent gateway-owned tools.
- Hiding their internal `function_call` events.
- Buffering calls whose names were unavailable in `output_item.added`.
- Flushing ordinary function events once their ownership was known.
- Deferring events after a gateway-owned call to preserve output ordering.
- Synthesizing the public custom-tool lifecycle after inference completed.

This logic was spread across `emit_upstream_stream_event`, `defer_after_gateway_call`, `record_first_hidden_gateway_output_index`, `defer_or_flush_function_event`, `should_hide_upstream_event`, `flush_pending_function_events`, and `drop_pending_function_events` in `executor/upstream.rs`.

The refactor moves function-call classification and public wire translation into a dedicated `FunctionSseTranslator`. The translator receives only a request-scoped `name → ToolType` map, rather than the full `ToolRegistry`, and handles normalized function calls according to their declared public type:

- Ordinary functions and namespace members pass through unchanged.
- Custom tools are translated incrementally from `function_call_arguments.*` into `custom_tool_call_input.*`.
- Gateway-owned MCP and web-search function frames are suppressed until their execution-specific public lifecycle is emitted.
- Calls without an initial name are buffered only until their type can be resolved, with a 256 KiB safety limit.

`ResponseAccumulator` still processes the original upstream frame first. This keeps the normalized `FunctionToolCall` as the authoritative internal representation used for dispatch and model continuation. The translator then reads that accumulated call state to produce client-facing frames, avoiding a second argument accumulator or changes to internal output items.

crates/agentic-server-core/src/executor/function_sse.rs
```diff
@@ -0,0 +1,606 @@
...
+        item_id: &str,
+        name: &str,
+        output_index: u32,
+        call: Option<AccumulatedFunctionCall<'_>>,
+    ) -> ExecutorResult<FunctionSseTranslation> {
+        if self.active_key(item_id, output_index).is_some() {
+            return Ok(FunctionSseTranslation::default());
+        }
+
+        let pending = self.take_pending(item_id);
```
franciscojavierarceo: `pending_unnamed` is keyed only by the original `item_id`, so when `output_item.done` supplies a new stable ID we lose the buffered `output_item.added` and argument deltas. i reproduced this and only `output_item.done` reached the client. we should resolve pending calls by `item_id` or `output_index` and add this case as a regression test.

crates/agentic-server-core/src/executor/function_sse.rs
```diff
@@ -0,0 +1,606 @@
...
+                output_index,
+                name: Some(name),
+                ..
+            } => self.start_call(item_id, name, *output_index, Some(frame.clone()), call),
+            EventPayload::OutputItemAdded {
+                item_id,
+                item_type: SSEItemType::FunctionCall,
+                name: None,
+                ..
+            } => self.buffer_unnamed(item_id.clone(), frame),
```
franciscojavierarceo: an unnamed function call doesn’t establish a defer boundary while we wait to classify it. this lets later output items reach the client before an earlier custom or gateway-owned call, leaving the SSE lifecycle out of order. we should establish a provisional boundary at the unnamed call’s `output_index` and release it once the call is classified.

crates/agentic-server-core/src/tool/custom.rs
```diff
@@ -0,0 +1,445 @@
...
+        .as_deref()
+        .map(str::trim)
+        .filter(|value| !value.is_empty())
+    {
+        fragments.push(description.to_owned());
+    }
+
+    fragments.push("Provide the raw tool input in the `input` string field.".to_owned());
+
+    if let Some(format) = &param.format {
```
franciscojavierarceo: the Lark or regex grammar is only copied into the function description while the generated `input` schema still accepts any string. that doesn’t preserve the constrained-decoding behavior of an OpenAI custom tool grammar. we should enforce or validate the grammar, or reject grammar formats we can’t support instead of treating them as prompt guidance. [OpenAI’s grammar documentation](https://developers.openai.com/api/docs/guides/function-calling#context-free-grammars)

crates/agentic-server-core/src/types/io/tools.rs
```diff
@@ -103,16 +142,42 @@ impl<'de> Deserialize<'de> for ToolChoice {
...
             )),
         }
     }
 }
 
+impl ToolChoice {
+    /// Converts client-facing custom-tool selectors to the function-tool shape
+    /// used by the normalized upstream tool declarations.
+    #[must_use]
+    pub(crate) fn normalized_for_upstream(&self) -> Self {
```
franciscojavierarceo, line 159: `normalized_for_upstream` rewrites every custom selector as a function selector without checking the corresponding declaration. a request with a function named `echo` and `tool_choice: {"type":"custom","name":"echo"}` is therefore accepted as a function choice instead of rejected. we should validate selectors against the declaration map before normalizing them.

maralbahari: Custom selectors are now validated against the request’s custom-tool declaration map before normalization, so a custom selector cannot silently target a function with the same name; valid custom selectors and `allowed_tools` entries are rewritten to functions only after validation.

crates/agentic-server-core/src/tool/custom.rs
maralbahari: Custom-tool grammar formats are no longer copied into the function description as prompt guidance. `CustomHandler::validate` now rejects any `format` declaration because function normalization cannot preserve OpenAI’s constrained-decoding semantics. `RequestPayload::to_upstream_request()` invokes `ResponsesTool::validate()` before normalization, and regression tests cover both Lark and regex formats. Unformatted custom tools remain supported.

crates/agentic-server-core/src/executor/function_sse.rs
maralbahari: Unnamed calls now establish a provisional defer boundary at their `output_index`; it is released and queued frames are flushed after client-owned classification, or retained when the call is gateway-owned.

crates/agentic-server-core/src/executor/function_sse.rs
maralbahari: Pending calls are now resolved by either the original `item_id` or `output_index`, with a regression test confirming buffered added and delta frames survive when `output_item.done` supplies a new stable ID.

franciscojavierarceo (approved):

PR #161, opened 2026-08-02
fix: preserve Claude Code Messages transport fidelity

franciscojavierarceo:
## Summary

- preserve the original Messages query string and open-ended Anthropic/Claude Code request headers across every server-side gateway tool round
- preserve upstream Messages status, body, request IDs, retry guidance, and rate-limit metadata while keeping local connection failures in valid Anthropic JSON envelopes
- forward documented upstream SSE error events and terminate without a synthetic successful message stop
- verify multi-block system attribution and a valid four-breakpoint, correctly ordered cache-control layout survive transparent proxying and native gateway tool loops
- run the real Claude Code CLI through agentic-server against recorded vLLM streaming responses and a local web-search replay service
- pin Claude Code 2.1.218, PyYAML 6.0.3, Node 24, and immutable action SHAs in the dedicated GitHub Actions check

The E2E job uses localhost services and a placeholder token. It requires no Anthropic credential, GPU, model download, or paid inference. This change does not add conversation persistence or other server-side client state; the request-scoped transport context is reused only for the gateway-owned tool loop.

For non-streaming requests, response metadata comes from the terminal upstream round. For streaming requests, HTTP response metadata comes from the initial upstream response because later rounds occur after the client response has started.

The visible reasoning double-render remains a vLLM-side issue tracked separately; this PR closes the agentic-api transport and error-fidelity gaps found by running Claude Code.

Related to #116.

## Test Plan

- cargo test
- cargo clippy --all-targets -- -D warnings
- cargo fmt --all -- --check
- pre-commit 4.4.0: pre-commit run --all-files
- pinned replay-server unit tests: 7 passed
- pinned Claude Code smoke test: capture valid: messages=2 transports=2 searches=1
- bash -n scripts/claude-code-smoke.sh


crates/agentic-server-core/src/proxy.rs
```diff
@@ -92,7 +92,13 @@ impl ProxyState {
...
 }
 
-fn filter_request_headers(headers: &HeaderMap, config: &Config, auth: ProxyAuth) -> reqwest::header::HeaderMap {
+/// Build the request headers forwarded to an upstream API.
+///
+/// Hop-by-hop and origin-specific headers are removed, all other headers stay
+/// open-ended, and the configured credential is injected only when the client
+/// did not supply one.
+#[must_use]
+pub fn upstream_request_headers(headers: &HeaderMap, config: &Config, auth: ProxyAuth) -> reqwest::header::HeaderMap {
```
ashwing, line 102: Anchoring on the forwarding entry point — the issue is the `REQUEST_DROP_EXTRA` list (`["host", "content-length"]`): `accept-encoding` isn't dropped there and isn't hop-by-hop, so open-ended forwarding now sends the client's `accept-encoding: gzip, br` upstream.

The proxy path gets away with this because it streams the upstream bytes back verbatim (`ProxyBody::Stream`) and the *client* decompresses. But #161 routes the gateway-tool loop through `response_lines` / `fetch_response_json_with_headers`, which parse the body in-process — and our `reqwest` is built `default-features = false` with no `gzip`/`brotli` feature, so it won't auto-decompress. If vLLM (or an intermediary) honors the forwarded `accept-encoding`, the loop parses compressed bytes and every SSE/JSON parse fails.

The old `extract_client_key` path sent only `Content-Type`, so upstream always replied uncompressed — this is a regression from the switch to open-ended forwarding. Add `accept-encoding` to `REQUEST_DROP_EXTRA` (the gateway advertises an encoding it can't decode). `content-encoding` on the response side is worth the same thought.

ashwing (commented):
First pass — the transport-fidelity direction is right, and priming the first upstream request before the handler commits an HTTP 200 is the correct fix for preserving streaming error status. One correctness issue on the open-ended header forwarding, plus the test gap around it.

crates/agentic-server/tests/messages_test.rs
```diff
@@ -277,6 +338,82 @@ async fn messages_with_web_search_tool_routes_to_native_loop() {
...
+    assert_eq!(requests.len(), 1);
+    assert_eq!(requests[0].uri, "/v1/messages?beta=true");
+    assert_eq!(requests[0].headers["anthropic-version"], "2023-06-01");
+    assert_eq!(
+        requests[0].headers["anthropic-beta"],
+        "future-beta-unknown,interleaved-thinking-2025-05-14"
+    );
+    assert_eq!(requests[0].headers["x-claude-code-session-id"], "session-loop");
+    assert_eq!(requests[0].headers["x-claude-code-agent-id"], "agent-loop");
+    assert_eq!(requests[0].headers["x-api-key"], "anthropic-key");
```
ashwing, line 430: These assert the allowlisted headers (`anthropic-version`, `anthropic-beta`, `x-claude-code-*`, `x-api-key`) survive — good — but nothing asserts that headers which would break in-process parsing get stripped. Once `accept-encoding` is dropped, add a negative case here: client sends `accept-encoding: gzip`, assert it's absent from `requests[0].headers`. That pins the parse-path contract so a later header-forwarding change can't silently reintroduce it.

crates/agentic-server/src/handler/http/messages.rs
```diff
@@ -35,30 +37,12 @@ async fn proxy_messages(
...
-/// Render an executor error as the Anthropic error envelope
-/// (`{"type":"error","error":{"type","message"}}`), consistent with the proxy
-/// path (E14).
-fn messages_error_response(err: &ExecutorError) -> Response {
+/// Preserve upstream Messages errors verbatim; render local executor failures
+/// as an Anthropic error envelope, consistent with the proxy path (E14).
+fn messages_error_response(err: ExecutorError) -> Response {
+    if let ExecutorError::LLMRequest { status, body } = err {
+        return (status, [(http::header::CONTENT_TYPE, "application/json")], body).into_response();
+    }
```
maralbahari, line 58: treats every LLMRequest as an upstream response, but connection failures and timeouts use the same variant with plain bodies such as upstream unavailable. Gateway-tool requests therefore return Content-Type: application/json with non-JSON text, breaking Anthropic SDK error parsing. Anthropic specifies that API errors are always JSON envelopes. Use a distinct transport-error variant, or only forward bodies that parse as [Anthropic errors](https://platform.claude.com/docs/en/api/errors)

crates/agentic-server-core/tests/messages_stream_test.rs
```diff
@@ -207,6 +259,37 @@ async fn messages_stream_presents_one_message_and_hides_gateway_tool() {
     );
 }
 
+#[tokio::test]
+async fn messages_stream_preserves_error_event_after_gateway_tool_round() {
```
maralbahari, line 264: this test simulates a second-round HTTP 400 response, not a documented SSE error received after HTTP 200.

When the upstream sends:

```text
event: error
data: {"type":"error","error":{"type":"overloaded_error","message":"Overloaded"}}
```

`MessagesStreamAccumulator::push` falls through to the wildcard branch and drops the event. The loop then calls `finish()`, which emits `message_stop`. Clients can therefore interpret partial or empty output as a successful completion instead of handling or retrying the error.
Please forward `type: "error"` events and terminate the synthetic stream. https://platform.claude.com/docs/en/build-with-claude/streaming.

crates/agentic-server-core/tests/fixtures/claude-code-cache-control-request.json
```diff
@@ -0,0 +1,84 @@
...
+  "model": "qwen3",
+  "max_tokens": 1024,
+  "stream": false,
+  "system": [
+    {
+      "type": "text",
+      "text": "<attribution>claude-code-session</attribution>",
+      "cache_control": {
+        "type": "ephemeral",
+        "ttl": "5m"
```
maralbahari: contains five explicit breakpoints, while Anthropic allows four. It also places `5m` entries before later `1h` entries; longer TTLs must precede shorter TTLs in the effective `tools → system → messages` prompt order. The replay server accepts this invalid shape, so the tests provide false compatibility confidence. [Anthropic prompt-caching constraints](https://platform.claude.com/docs/en/build-with-claude/prompt-caching)

maralbahari (commented):
I left some inline comments.

couldnt add this as inline but worth addressing:

[inference.rs (line 88)](https://github.com/vllm-project/agentic-api/blob/971afa93bf0842decd6329cc1390b375fd070c44/crates/agentic-server-core/src/executor/inference.rs#L88) retains only status/body, and [messages.rs (line 44)](https://github.com/vllm-project/agentic-api/blob/971afa93bf0842decd6329cc1390b375fd070c44/crates/agentic-server/src/handler/http/messages.rs#L44) reconstructs only Content-Type. A 429 therefore loses retry-after, while request-id and rate-limit headers disappear from both error and successful loop responses. Preserve filtered upstream response headers, at least from the terminal round. [Anthropic rate-limit headers](https://platform.claude.com/docs/en/api/rate-limits), [request IDs](https://platform.claude.com/docs/en/api/errors)

PR #166, opened 2026-08-05
Feat/conversion endpoints

### LINKED ISSUE #155: Complete the OpenAI-compatible Conversations API
## Problem statement / motivation

`agentic-api` already has most of the internal machinery needed for durable conversation state: ordered item persistence, response records, rehydration, automatic input/output item appends, streaming and non-streaming continuation, and conversation isolation.

The public API surface is still incomplete and is not wire-compatible with the current OpenAI Conversations API:

- Only `POST /v1/conversations` is registered.
- Conversation creation ignores the standard `metadata` and initial `items` fields and always returns empty metadata.
- Conversation retrieve, update, and delete operations are missing.
- Conversation item create, list, retrieve, and delete operations are missing.
- Responses requests use the local `conversation_id` field rather than the standard `conversation` reference.
- A Responses request with `store=true` currently calls `ConversationStore::get_or_create()`, allowing an unknown client-supplied conversation ID to create a new conversation implicitly.
- `tenant_id` columns exist, but conversation and item operations are not tenant-scoped.

This means the existing stateful behavior is useful internally, but OpenAI SDK clients cannot rely on the standard Conversations resource contract. It also prevents us from cleanly treating an authorized Conversation ID as the Responses-side source of a normalized session coordinate for downstream llm-d/vLLM policy.

Relevant current code:

- `crates/agentic-server/src/app.rs::build_router`
- `crates/agentic-server/src/handler/http/conversations.rs::conversations`
- `crates/agentic-server-core/src/types/request_response.rs::RequestPayload`
- `crates/agentic-server-core/src/executor/rehydrate.rs::rehydrate_conversation`
- `crates/agentic-server-core/src/executor/rehydrate.rs::from_conversation`
- `crates/agentic-server-core/src/executor/modes/conversation.rs::ConversationHandler`
- `crates/agentic-server-core/src/storage/conversation.rs::ConversationStore`
- `crates/agentic-server-core/src/storage/models/item.rs::create_in_tx_with_next_conversation_seq`

Official API references:

- https://developers.openai.com/api/reference/resources/conversations/methods/create
- https://developers.openai.com/api/reference/resources/conversations/methods/retrieve
- https://developers.openai.com/api/reference/resources/conversations/methods/update
- https://developers.openai.com/api/reference/resources/conversations/methods/delete
- https://developers.openai.com/api/reference/resources/conversations/subresources/items/methods/list
- https://developers.openai.com/api/reference/resources/responses/methods/create

## Proposed solution

Complete the Conversations API while preserving the existing executor and storage boundaries.

### Conversation resource

Implement the standard resource operations and request/response shapes:

- `POST /v1/conversations`
  - Parse and persist `metadata`.
  - Accept and persist initial `items` in the supplied order.
  - Remove the nonstandard requirement around a `store` field.
- `GET /v1/conversations/{conversation_id}`
- `POST /v1/conversations/{conversation_id}` for metadata updates
- `DELETE /v1/conversations/{conversation_id}`

### Conversation items

Implement the standard item operations:

- `POST /v1/conversations/{conversation_id}/items`
- `GET /v1/conversations/{conversation_id}/items`
- `GET /v1/conversations/{conversation_id}/items/{item_id}`
- `DELETE /v1/conversations/{conversation_id}/items/{item_id}`

Match the current OpenAI schemas, ordering, pagination parameters, list envelope, object types, errors, and deletion responses. Reuse the existing `items` table and sequence-number allocation rather than creating a parallel item representation.

### Responses integration

- Add the standard `conversation` reference to the Responses request type and make it the canonical wire field.
- Preserve `conversation_id` only as a temporary input alias for existing agentic-api callers; reject requests that provide both aliases with different values.
- Return the standard `conversation` representation in blocking responses and streaming response objects instead of emitting `conversation_id`.
- Continue rejecting `conversation` together with `previous_response_id`.
- Require a referenced conversation to exist. Do not implicitly create it from a Responses request.
- Preserve the current behavior in which the conversation's ordered items are prepended before inference and the response's new input/output items are appended after completion.
- Preserve conversation persistence when the Responses request sets `store=false`; `store` governs stored-response behavior, not the lifetime of an explicitly referenced conversation.

### Tenant and lifecycle identity

- Scope conversation, item, and response lookups and mutations by the authenticated tenant.
- Prevent cross-tenant access even when a caller knows a conversation or item ID.
- Make the authorized, tenant-scoped Conversation ID available as the Responses-side source of a normalized internal session coordinate.
- Keep that coordinate as lifecycle/routing metadata. It must not be added to vLLM KV block hashing or treated as proof that reusable blocks are resident.

## Acceptance criteria

- [ ] All four Conversation resource operations are implemented with OpenAI-compatible request and response shapes.
- [ ] All four Conversation Item operations are implemented, including ordering and pagination.
- [ ] `POST /v1/conversations` preserves supplied metadata and initial items.
- [ ] `POST /v1/responses` accepts the standard `conversation` reference.
- [ ] Blocking and streaming Response objects expose the standard conversation field/shape.
- [ ] `conversation` and `previous_response_id` are rejected when supplied together.
- [ ] Unknown conversation IDs return a not-found error instead of creating a conversation implicitly.
- [ ] Input and output items from a completed Response are appended exactly once and in deterministic order.
- [ ] `store=false` Responses attached to a conversation still update that conversation.
- [ ] Conversation and item operations are tenant-scoped and have cross-tenant negative tests.
- [ ] Existing conversation continuation behavior remains covered for streaming and non-streaming requests.
- [ ] The OpenAI SDK can create, retrieve, update, delete, and manipulate items against agentic-api without custom field names.
- [ ] README and API documentation describe the implemented surface accurately.

## Test plan

Add HTTP and storage integration coverage for:

- Create with empty body, metadata, initial items, and mixed supported item types.
- Retrieve, metadata update, and deletion.
- Item create, list ordering/pagination, retrieve, and deletion.
- Conversation-not-found and item-not-found errors.
- Standard `conversation` continuation over blocking, SSE, and Responses WebSocket transports.
- `conversation` plus `previous_response_id` rejection.
- `store=false` with an existing conversation.
- Concurrent item appends and deterministic sequence allocation on SQLite and PostgreSQL.
- Tenant isolation for every read and mutation path.
- Compatibility behavior for the temporary `conversation_id` input alias.
- OpenAI SDK conformance tests using the public Conversations and Responses clients.

Existing tests in `crates/agentic-server-core/tests/stateful_conversation_integration.rs` should remain as regression coverage, but they should be migrated to the standard `conversation` field.

## Out of scope

- Implementing llm-d or vLLM session-aware routing policy in this issue.
- Treating a Conversation ID as exact KV-cache or block identity.
- Adding hard cache pinning or retention guarantees.
- Changing Anthropic Messages persistence semantics.

## Enterprise Readiness

This is the canonical enhancement for standard conversation references, creation payloads, and Conversations resource compatibility. Coordinate with #107 for ownership and the Enterprise Readiness request-fidelity work for field handling across execution paths.

Completion evidence should include standard conversation references over JSON, SSE, and Responses WebSocket; supported legacy-alias behavior and conflicting-input validation; initial items/metadata; deterministic appends; and tenant-scoped resource operations. Existing acceptance criteria remain authoritative.

Enterprise Readiness tracker: https://github.com/vllm-project/agentic-api/issues/316

Prasannajaga:
Fixes #155 by adding complete stateful Conversations API support and integration coverage.

This change adds support for:

- Creating conversations with an empty body, metadata, initial items, and mixed supported item types.
- Retrieving, updating, and deleting conversations.
- Creating, listing, paginating, retrieving, and deleting conversation items.
- Returning proper not-found errors for conversations and items.
- Continuing conversations through blocking, SSE, and Responses WebSocket transports.
- Rejecting requests that provide both `conversation` and `previous_response_id`.
- Handling `store=false` correctly when an existing conversation is provided.
- Concurrent item appends with deterministic ordering on SQLite and PostgreSQL.
- Tenant isolation across every conversation and item read or mutation path.
- Backwards compatibility for the temporary `conversation_id` input alias.
- OpenAI SDK conformance coverage using the public Conversations and Responses clients.

The existing stateful conversation tests remain as regression coverage and now use the standard `conversation` request field. This is needed so conversations can be managed independently from response generation and reused consistently across all supported transports. It also ensures that persisted conversation data is correctly ordered, isolated by tenant, and compatible with the public OpenAI SDK.

## Test Plan

Executed the relevant Rust tests covering:

```bash
cargo test --workspace
```

Formatting and lint checks:

```bash
cargo fmt -- --check
cargo clippy --all-targets -- -D warnings
```

OpenAI SDK conformance test:

```bash
AGENTIC_API_BASE_URL=http://127.0.0.1:9000/v1 \
OPENAI_API_KEY=test-key \
OPENAI_TEST_MODEL=qwen2.5-coder-0.5b-instruct \
uv run --with openai python tests/openai_sdk_conformance.py
```

The SDK test requires the gateway and the configured LLM server to be running.


crates/agentic-server-core/src/executor/modes/conversation.rs
```diff
@@ -1,7 +1,12 @@
 //! Conversation storage handler — owns all conversation store operations.
 
+#![allow(clippy::missing_errors_doc)]
```
maralbahari: it's repo policy to not ignore any clippy warning. need to remove this and add error docs.

crates/agentic-server-core/src/executor/accumulator.rs
```diff
@@ -533,6 +533,13 @@ impl ResponseAccumulator {
             previous_response_id: previous_response_id.map(str::to_string),
             conversation_id: self.conversation_id,
             instructions: instructions.map(str::to_string),
+            parallel_tool_calls: true,
```
maralbahari: we are not supporting `parallel_tool_calls` as off now.

tests/openai_sdk_conformance.py
```diff
@@ -0,0 +1,125 @@
+"""Conformance checks using the public OpenAI Python Conversations and Responses clients.
```
maralbahari: we do not check compatibility between openai and gateway in this format. we have cassettes recorder where we have a listener to record request and responses payload for both conversation and responses api. after recording cassettes from openai and gateway you compare gateway behavior to be compatible with openai not only the logical but also the the SSE lifecycle in case of streaming events.
please refer to https://github.com/vllm-project/agentic-api/blob/main/crates/agentic-server-core/tests/cassettes/README.md

maralbahari (commented):
Thanks for contributing to agentic-api! I’ve left a few inline comments covering some of the maintenance practices we follow across the repository. We also use the [rust-skills guidelines](https://github.com/leonardomso/rust-skills) as a reference for Rust best practices, including addressing Clippy warnings.

Could you please take a look at these items first? Once they’re addressed, I’ll continue reviewing the implementation and logic. If anything in the feedback is unclear, please don’t hesitate to ask. Thanks again for your contribution!

crates/agentic-server/src/app.rs
```diff
@@ -247,6 +180,20 @@ pub fn build_router_with_auth(
...
+                .post(update_conversation)
+                .delete(delete_conversation),
+        )
+        .route(
+            "/v1/conversations/{conversation_id}/items",
+            post(create_items).get(list_items),
+        )
+        .route(
+            "/v1/conversations/{conversation_id}/items/{item_id}",
+            get(retrieve_item).delete(delete_item),
```
maralbahari, line 195: I think for each route you can try to open separate PR instead of all in one PR ?

crates/agentic-server-core/src/executor/modes/conversation.rs
Prasannajaga: I have changed this

crates/agentic-server-core/src/executor/accumulator.rs
Prasannajaga: Ill remove it for now

tests/openai_sdk_conformance.py
Prasannajaga: got it! Ill remove this and follow cassettes order

crates/agentic-server/src/app.rs
Prasannajaga: is it because of the file huge file changes will be hard to track or something we follow as guidelines ? 
I would definitely do it in separate PR just curious to know

crates/agentic-server/tests/common/mod.rs
```diff
@@ -39,16 +39,71 @@ pub fn test_state(config: &Config) -> AppState {
...
-        readiness_tracker: ReadinessTracker::default(),
         shutdown_token: CancellationToken::new(),
         websocket_tracker: WebSocketTracker::default(),
         llm_api_base: config.llm_api_base.clone(),
-        skip_llm_ready_check: config.skip_llm_ready_check,
         openai_api_key: config.openai_api_key.clone(),
     }
 }
 
+#[allow(dead_code)]
```
maralbahari, line 49: These `dead_code` suppressions show that this fixture is not actually common. Every file in `tests/` is compiled as a separate integration-test crate with its own mod common, while `storage_backed_state` is used only by `conversations_test.rs.` Move the route-specific state helper next to that test, or introduce a real shared test-support crate; do not hide the unused code from clippy.

crates/agentic-server/src/app.rs
maralbahari: I suggest to split this into two focused PRs:
Conversation resource:
`GET/POST/DELETE /v1/conversations/{conversation_id}`
Only the handler, storage/model changes, tenant handling, tests, and cassettes required by those operations.

Conversation items:
`GET/POST /v1/conversations/{conversation_id}/items`
`GET/DELETE /v1/conversations/{conversation_id}/items/{item_id}`
Item serialization, pagination, storage changes, tests, and cassettes.

Changes to POST `/v1/conversations` should follow the same ownership: `metadata/resource` behavior belongs with PR 1, while initial-item behavior belongs with PR 2. PR 2 may be stacked on PR 1 if there is a genuine dependency, but each commit/PR should contain only logic needed for its route group.
The reason for splitting is concrete: this PR currently adds 2,861 lines across 41 files and combines two independently testable public API contracts. Separate PRs give each contract its own OpenAI reference recordings, review boundary, and rollback boundary.

crates/agentic-server/tests/common/mod.rs
```diff
@@ -39,16 +39,71 @@ pub fn test_state(config: &Config) -> AppState {
...
+struct TestDb {
```
maralbahari, line 50: This custom temporary database and Drop cleanup are unnecessary. Test storage is already created and migrated with one call:
```rust
let pool = create_pool_with_schema(Some("sqlite://?mode=memory")).await?
```
The function is in `agentic-server-core/src/storage/pool.rs` and is already used this way in `executor/request.rs` and `executor/modes/conversation.rs`. Use that pool directly to construct the conversation and response stores.

crates/agentic-server-core/src/storage/models/conversation.rs
```diff
@@ -25,14 +27,47 @@ pub struct Conversation {
 /// # Errors
 /// Returns `DbResult::Err` if the database insertion fails.
 pub async fn create(pool: &DbPool, id: &str) -> DbResult<Conversation> {
+    create_with_metadata(pool, id, None, None).await
```
maralbahari, line 30: create is now only a stale compatibility wrapper around the new function. Since this PR changes creation semantics and all internal call sites are in-tree, change create itself to accept tenant_id and metadata, update the callers, and name the transaction form `create_in_tx`. Please do not introduce a permanent `create_with_metadata/create_with_metadata_in_tx` parallel API.

if you find any other such surface API pattern introduced, try to avoid them

crates/agentic-server-core/src/storage/models/item.rs
```diff
@@ -131,6 +133,20 @@ pub async fn create_in_tx(
     tx: &mut DbTransaction<'_>,
     items: Vec<(String, String)>,
     conversation_id: Option<&str>,
+) -> DbResult<Vec<Item>> {
+    create_in_tx_with_tenant(tx, items, conversation_id, None).await
```
maralbahari, line 137: Same issue here: add tenant_id to the existing `create_in_tx` signature and update its call sites. Keeping `create_in_tx` only to pass None into `create_in_tx_with_tenant` leaves two creation paths and makes the old path encode an increasingly incorrect default.

crates/agentic-server-core/src/storage/models/response.rs
```diff
@@ -39,19 +41,46 @@ pub async fn create_in_tx(
     previous_response_id: Option<&str>,
     history_item_ids: Option<&str>,
     metadata: Option<&str>,
+) -> DbResult<Response> {
+    create_in_tx_with_tenant(
```
maralbahari, line 45: Please modify the existing `create_in_tx ` API to accept tenant_id and migrate its callers directly. A wrapper whose only behavior is calling `create_in_tx_with_tenant(..., None)` is unnecessary stale API surface.

crates/agentic-server/src/handler/http/conversations.rs
```diff
@@ -1,31 +1,316 @@
...
+    items: Vec<Value>,
+}
+
+#[derive(Debug, Default, Deserialize)]
+pub struct ListItemsQuery {
+    after: Option<String>,
+    limit: Option<usize>,
+    order: Option<String>,
+    #[allow(dead_code)]
+    include: Option<Vec<String>>,
```
maralbahari, line 39: This is a public API parameter, but the handler accepts it and silently ignores it. Create and retrieve do not accept `include` at all. That makes clients believe their requested representation was honored. Record OpenAI behavior for every supported `include` value and either implement it consistently on create/list/retrieve or explicitly reject unsupported values. The `dead_code` suppression must not hide an unimplemented protocol field.

crates/agentic-server-core/src/storage/conversation.rs
```diff
@@ -156,9 +381,42 @@ impl ConversationStore {
         .await
     }
 
+    #[allow(clippy::too_many_arguments)]
```
maralbahari, line 384: another clippy warning suppression.

crates/agentic-server-core/src/storage/conversation.rs
```diff
@@ -156,9 +381,42 @@ impl ConversationStore {
...
+            Some(expected_version),
+            response_id,
+            previous_response_id,
+            new_items,
+            metadata,
+        )
+        .await
+    }
+
+    #[allow(clippy::too_many_arguments)]
```
maralbahari, line 415: avoid the suppression.

crates/agentic-server-core/src/storage/conversation.rs
```diff
@@ -66,13 +109,176 @@ impl ConversationStore {
...
+        &self,
+        conversation_id: &str,
+        tenant_id: Option<&str>,
+        after: Option<&str>,
+        limit: usize,
+        descending: bool,
+    ) -> StoreResult<ConversationItemPage> {
+        let pool = self.pool()?;
+        self.get_for_tenant(conversation_id, tenant_id).await?;
+        let mut rows = item::get_items_by_conversation_for_tenant(pool, conversation_id, tenant_id).await?;
```
maralbahari, line 216: `limit` does not bound database work or memory here. This loads the entire conversation, may reverse it, scans it for the cursor, and then clones a slice of at most 100 rows. A sufficiently long conversation can make one list request allocate its complete history. Push tenant filtering, ordering, the cursor predicate, and `LIMIT limit + 1` into SQL; use the extra row to derive `has_more`.

crates/agentic-server-core/src/storage/types/item.rs
```diff
@@ -116,6 +131,22 @@ impl InOutItem {
     }
 }
 
+impl From<StorageDbItem> for ConversationItemData {
+    fn from(row: StorageDbItem) -> Self {
+        let mut item = deserialize_from_str_opt::<Value>(&row.data).unwrap_or(Value::Null);
```
maralbahari, line 136: Persisted JSON and seq are storage invariants for a conversation item. Converting malformed JSON to null and a missing sequence to `0` turns corruption into a successful but invalid API response. Make this a fallible `TryFrom<StorageDbItem>` returning `StorageError`, and propagate failures from append/list/retrieve rather than silently manufacturing values.

crates/agentic-server-core/src/storage/types/item.rs
```diff
@@ -43,6 +44,20 @@ pub enum InOutItem {
     Output(OutputItem),
 }
 
+#[derive(Debug, Clone, PartialEq)]
+pub struct ConversationItemData {
```
maralbahari, line 48: `ConversationItemData` is being used as a shortcut between the database row and the HTTP response:

- `id` preserves the database row ID for pagination and retrieval.
- `item: Value` lets the handler inject that ID and remove `_agentic_item_kind`.
- `created_at` and `sequence` carry database metadata, although the HTTP handler discards both.

This bypasses the existing typed `Item → InOutItem` conversion. It also creates two item identities: the ID already inside a typed message/tool/reasoning item and the generated database row ID that later overwrites it.

OpenAI models conversation items as a typed discriminated union, and its list response contains `List[ConversationItem]`, not arbitrary JSON values. Individual returned variants define their typed `id`, `status`, and other fields. [ConversationItem union](https://github.com/openai/openai-python/blob/main/src/openai/types/conversations/conversation_item.py), [ConversationItemList](https://github.com/openai/openai-python/blob/main/src/openai/types/conversations/conversation_item_list.py).

 Please remove `ConversationItemData`. This creates a second, untyped database-to-domain conversion path even though `models::Item` already maps stored data into the typed `InOutItem` union.

 The current implementation parses the stored payload as an arbitrary `Value`, removes the private storage marker, and overwrites the item’s existing ID with the database row ID. This can change the identity of model-generated `message`, `function_call`, `custom_tool_call`, and `reasoning` items. It also allows malformed stored data to escape the business layer without typed validation. `created_at` is discarded by the HTTP handler, while `sequence` is a database pagination detail and should not be part of the returned business object.

 Please preserve one conversion contract:

 `models::Item → TryFrom<Item> for InOutItem → typed HTTP serialization`

 Make the database conversion fallible and propagate `StorageError`. The append and retrieve operations should return `InOutItem`, and the list operation should use a typed page containing `Vec<InOutItem>`, `first_id`, `last_id`, and `has_more`. Keep `seq` inside the storage query.

If the existing input/output variants cannot represent fields that OpenAI returns—particularly generated `id` or `status` on tool outputs—extend those typed variants according to the OpenAI cassettes. Assign the canonical item ID once when the item is stored; do not repair or overwrite arbitrary JSON during reads.

A suitable direction would be:

```rust
pub struct ConversationItemPage {
    pub data: Vec<InOutItem>,
    pub first_id: Option<String>,
    pub last_id: Option<String>,
    pub has_more: bool,
}
```

The important point is that `InOutItem` remains the business contract. If it is currently insufficient to serialize the OpenAI conversation-item response, its typed variants should be completed rather than introducing `serde_json::Value` as an escape hatch.

tests/openai_sdk_conformance.py
maralbahari: The current Python recorder does not cover the conversation-management or conversation-item routes introduced here. Its `conv` mode only calls `POST /v1/conversations` and then uses the returned ID with `/v1/responses`. It never calls:

- `GET/POST/DELETE /v1/conversations/{conversation_id}`
- `GET/POST /v1/conversations/{conversation_id}/items`
- `GET/DELETE /v1/conversations/{conversation_id}/items/{item_id}`

Therefore, both split PRs (one PR `conversation/` another `conversation/items`) need new cassette scenarios.

For the conversation-management PR, record create, retrieve, metadata update, retrieve-after-update, delete, and retrieve-after-delete.

For the items PR, add a `conversation_items` recorder scenario covering create, list, retrieve, and delete. It must capture generated IDs and reuse them in later requests. Cover all supported item formats, pagination, `include`, missing IDs, invalid requests, and OpenAI error responses.

Record every scenario twice:

1. Against OpenAI as the ground truth.
2. Against agentic-api with its database, gateway, and vLLM running.

Use identical requests and prompts for both recordings. Normalize dynamic IDs and timestamps, but compare status codes, response shapes, ordering, pagination, ID relationships, and history behavior.

The following stateful prompt scenarios are especially important.

#### Items added through the API affect model context

Add this item through `/v1/conversations/{conversation_id}/items`:

`Remember exactly: MANUAL=30.`

Then call `/v1/responses` with the conversation ID:

`Using only the conversation history, reply exactly: MANUAL=<value-or-MISSING>`

This verifies that an item added through the items API is actually persisted and passed to the model.

#### Deleting an item changes conversation history

Create two turns:

- Turn 1: `Remember exactly: ALPHA=314159. Reply only OK.`
- Turn 2: `Remember exactly: BETA=271828. Reply only OK.`

List the conversation items, delete the first user message, and then continue with the conversation ID:

`Using only the visible conversation history, reply exactly: ALPHA=<value-or-MISSING>;BETA=<value-or-MISSING>`

Also ask the same question using the second turn’s `previous_response_id`.

This records whether OpenAI treats deletion differently for:

- The conversation’s current history.
- A previous-response checkpoint created before the deletion.

The gateway must reproduce OpenAI’s recorded behavior rather than assume both histories behave identically.

#### Branching after adding an item

Create this history:

- Turn 1: `Remember exactly: ROOT=10. Reply only OK.`
- Turn 2: `Remember exactly: MAIN=20. Reply only OK.`
- Directly add through the items endpoint: `Remember exactly: MANUAL=30.`

Then ask this question in three ways:

`Using only the history visible to this request, reply exactly: ROOT=<value-or-MISSING>;MAIN=<value-or-MISSING>;MANUAL=<value-or-MISSING>`

Run it:

1. With `previous_response_id` from turn 1.
2. With `previous_response_id` from turn 2.
3. With the conversation ID.

This determines which earlier turns and manually added items are visible on each branch.

Follow the existing mathematical branching examples in:

- `conv-multi-turn-single-branch-gpt-4o-nonstreaming.yaml`
- `conv-multi-branch-multi-turn-gpt-4o-nonstreaming.yaml`
- `record_text_only_cassettes.sh`
- `stateful_conversation_integration.rs`

Those tests use ordered arithmetic to make incorrect history obvious—for example, a branch from the response containing `4` should produce `5`, not the main branch’s later result.

Add repeatable recording scripts following `record_text_only_cassettes.sh` and `record_web_search_cassettes.sh`. The Python recorder must generate the YAML; cassettes must not be written or edited manually. Finally, add Rust integration tests comparing gateway behavior with the OpenAI reference.

crates/agentic-server/src/handler/http/conversations.rs
```diff
@@ -1,31 +1,316 @@
...
+use agentic_core::{InOutItem, InputItem};
 
-use super::super::common::{executor_error_response, extract_store, read_bytes};
+use super::super::common::{executor_error_response, read_bytes};
 use crate::app::AppState;
+use crate::auth::AuthenticatedPrincipal;
 
-pub async fn conversations(State(state): State<AppState>, req: Request) -> Response {
+const MAX_ITEMS_PER_REQUEST: usize = 20;
+
```
maralbahari, line 14: overall feedback on this file from architectural POV.

### Keep storage behind the executor boundary

The HTTP conversation handler currently depends on storage types such as `ConversationData`, `ConversationItemData`, and `ConversationItemPage`. Although it calls `conv_handler`, that handler passes storage results back to Axum, so the database representation still leaks into the HTTP layer.

Please preserve this layering:

```text
HTTP handler
    → typed Conversation API requests/responses
ConversationHandler in executor/modes/conversation.rs
    → internal storage operations
ConversationStore / database
```

Define the public payloads under `agentic-server-core/src/types/conversations/`. Then expose typed operations from `ConversationHandler`, for example:

```rust
pub async fn create(
    &self,
    tenant_id: Option<&str>,
    request: CreateConversationRequest,
) -> ExecutorResult<ConversationResponse>;

pub async fn retrieve(
    &self,
    tenant_id: Option<&str>,
    conversation_id: &str,
) -> ExecutorResult<ConversationResponse>;

pub async fn update(
    &self,
    tenant_id: Option<&str>,
    conversation_id: &str,
    request: UpdateConversationRequest,
) -> ExecutorResult<ConversationResponse>;

pub async fn create_items(
    &self,
    tenant_id: Option<&str>,
    conversation_id: &str,
    request: CreateItemsRequest,
) -> ExecutorResult<ConversationItemListResponse>;

pub async fn list_items(
    &self,
    tenant_id: Option<&str>,
    conversation_id: &str,
    query: ListItemsQuery,
) -> ExecutorResult<ConversationItemListResponse>;
```

`ConversationHandler` should call `ConversationStore`, validate requests, convert `StorageError` into `ExecutorError`, and map storage results into typed API responses.

The Axum handler should then remain small:

```rust
let request: CreateConversationRequest = read_json(body).await?;

match state
    .exec_ctx
    .conv_handler
    .create(tenant_id.as_deref(), request)
    .await
{
    Ok(response) => Json(response).into_response(),
    Err(error) => executor_error_response(error),
}
```

The HTTP layer should not deserialize stored metadata, build responses using `json!`, manipulate storage markers, or reference database-specific item/page types. This keeps storage private and makes the executor conversation mode the single application boundary.

crates/agentic-server-core/src/executor/modes/conversation.rs
maralbahari: please address all clippy warning. I see many of them I tried to add them inline comments but I might have missed some of them.

Prasannajaga: checking it! I was mixed with the existing ones little bit I move that cleanly now

Prasannajaga:
Hey @maralbahari , thanks for the detailed review. I’ll close this PR since it was getting a bit confusing to alter the requested changes, and I’ll create 2 new PRs as we planned. so moving this to draft

LOGO127:
AI-assisted validation note for the planned split mentioned above: the qualified upsert in this PR also addresses a reproducible failure on current main `c04ac196f441`, independently of the Conversations API expansion.

Using PostgreSQL 16.15, Rust 1.98.0, the existing public `ConversationStore` and a fresh local database:

- `get_or_create` for a new ID fails with SQLSTATE `42702`: `column reference "created_at" is ambiguous`.
- `create` followed by `get_or_create` for that existing ID fails identically.
- Both matching SQLite controls pass.

A one-line diagnostic on main, qualifying only the right-hand side as `conversations.created_at` (the approach already present in this PR), makes all four checks pass. The existing-row check also confirms that a deliberately seeded timestamp and metadata remain unchanged. I reverted that diagnostic afterward; I have not opened a competing PR or claimed this change as new work.

This surfaced while running connection-local state regression tests against PostgreSQL: ordinary responses/retention checks passed, but explicit-conversation tests failed even in their **no-session** controls. The three existing PostgreSQL schema/migration checks pass, so they do not catch this execution-path failure.

Would it be useful to land the minimal upsert correction plus its database regression separately from the larger API changes, or is it already part of a planned replacement PR? I can provide the isolated main-based test. These are agent-run local checks with final contributor review pending, not a claim that the full #166 branch, tenant behavior, deployed gateway or live-model E2E has been validated.

Issue #163, opened 2026-08-04
Compaction related issues/observations when using agentic-api with Codex-CLI


noobHappylife:
### Problem statement / motivation

While validating codex CLI (v0.146.0) against agentic-api as a Responses provider, I traced and live-tested the full compaction interaction. 

I found 2 issues, and 2 observations related to this. 

### What happens

 Codex's compaction dispatch (codex-rs/core/src/tasks/compact.rs) is:

 ```
   supports_remote_compaction() == is_openai() || azure_url_markers
   ├─ true  → remote (V2 by default in v0.146.0, see Issue 2)
   └─ false → LOCAL compact: in-process summarization turn on /v1/responses
 ```

 supports_remote_compaction() (model-provider-info/src/lib.rs:417) is config-only — provider name == "OpenAI" (case-sensitive) or Azure URL markers (codex-api/provider.rs:106). The documented agentic-api setup uses name = "agentic-api" → false → local compact:

### What happens

In codex v0.146.0, RemoteCompactionV2 is Stage::Stable, default_enabled: true (codex-rs/features/src/lib.rs:1434). With name = "OpenAI", codex uses V2: it appends a compaction_trigger item to /v1/responses input and requires the response stream to contain exactly one compaction item.

agentic-api has no compaction_trigger input variant (deserializes to Unknown), runs a normal inference turn, the model emits no compaction item, and codex fatals:

```
Error running remote compact task: Fatal error: remote compaction v2 expected exactly one compaction output item, got 0 from 2 output items
```

(collect_compaction_output, codex-rs/core/src/compact_remote_v2.rs:421)

### Workaround (V1 works)

```toml
[features]
remote_compaction_v2 = false
```

forces the V1 path (POST /v1/responses/compact), which `agentic-api` implemented. 

### Proposed solution

1) Perhaps, if we want to let the user to use remote compaction, then need to update the setup to name="OpenAI" (case sensitive). And to turn off v2 compaction now. 

2) For the v2 compaction, it seems that it's only a native concept in codex cli, and it's not found in OpenAI documentations (https://developers.openai.com/api/docs/guides/compaction, https://developers.openai.com/api/docs/guides/websocket-mode#compaction-and-creating-new-responses). 

### Additional context

Here are the 2 observations I observed, when using with Codex CLI. 

 1) — Assistant messages are dropped; many consecutive user messages remain
[...]

PR #183, opened 2026-08-18
feat: Support Codex CLI remote compaction V2

noobHappylife:
## Summary

This adds server-side support for Codex CLI's remote compaction V2 to the Responses API, so server side compaction of agentic-api works with codex out of the box instead of fall back to client-side (local) compaction. Addressed #163 

In codex CLI (tested v0.146.0) remote compaction V2 is the default path. The client appends a bare {"type":"compaction_trigger"} item to a streaming /v1/responses request.

What I changed:

- Added a compaction_trigger variant to InputItem; previously, it deserialized as Unknown. Also added ResponsesInput::has_compaction_trigger().
- Added an executor branch that detects the trigger, removes it from model input, and runs the existing compact_items pipeline the same summarization pipeline used by /v1/responses/compact.
- Added OutputItem::Compaction and the corresponding SSEItemType::Compaction mapping.
- Streamed the generated checkpoint through response.output_item.added and response.output_item.done, followed by response.completed.
- Routed trigger-bearing HTTP requests through the executor instead of directly proxying them upstream.
- Stored retained user messages as request inputs and the generated checkpoint exactly once as response output, preventing duplicate checkpoints from dropping retained context on later turns.

Note: on the codex side, remote compaction only activates when the provider is name = "OpenAI" (case-sensitive) — that check is config-only on their end, documented in #163.

## Test Plan

- Added seven unit tests.
- Full workspace test suite: 695 passed, 0 failed, 9 ignored.
- cargo clippy --all-targets -- -D warnings passed.
- cargo fmt -- --check passed.
- Live smoke test against vLLM using google/gemma-4-26B-A4B-it: a streaming request containing compaction_trigger returned:
  response.created
  → response.in_progress
  → response.output_item.added
  → response.output_item.done (compaction)
  → response.completed

  The compaction item contained a model-generated checkpoint summary.

- Manual Codex test: configured agentic-api as the model provider, ran /compact in the TUI, and verified both the remote V2 compaction flow and a subsequent turn using the checkpoint as context.

### DIFF
--- a/crates/agentic-server-core/src/executor/accumulator.rs
+++ b/crates/agentic-server-core/src/executor/accumulator.rs
@@ -479,7 +479,7 @@ impl ResponseAccumulator {
                 text: String::with_capacity(256),
             }),
             SSEItemType::WebSearchCall if !item_id.is_empty() => Some(InFlight::WebSearchCall { item: None }),
-            SSEItemType::WebSearchCall => None,
+            SSEItemType::Compaction | SSEItemType::WebSearchCall => None,
             SSEItemType::McpCall => McpCall::try_from(payload).ok().map(|item| InFlight::McpCall { item }),
             SSEItemType::McpListTools => McpListTools::try_from(payload)
                 .ok()
--- a/crates/agentic-server-core/src/executor/accumulator.rs
+++ b/crates/agentic-server-core/src/executor/accumulator.rs
@@ -20,8 +20,8 @@ use crate::executor::function_sse::{FunctionSseTranslation, FunctionSseTranslato
 use crate::types::event::{MessageStatus, ResponseStatus};
 use crate::types::io::output::McpListTools;
 use crate::types::io::{
-    ApplyDone, CustomToolCall, FunctionToolCall, OutputItem, OutputMessage, OutputTextContent, ReasoningOutput,
-    ReasoningTextContent, ResponseUsage,
+    ApplyDone, CompactionItem, CustomToolCall, FunctionToolCall, OutputItem, OutputMessage, OutputTextContent,
+    ReasoningOutput, ReasoningTextContent, ResponseUsage,
 };
 use crate::types::io::{McpCall, WebSearchCall};
 use crate::types::request_response::{IncompleteDetails, ResponsePayload};
@@ -38,6 +38,7 @@ enum InFlight {
     WebSearchCall { item: Option<WebSearchCall> },
     McpCall { item: McpCall },
     McpListTools { item: McpListTools },
+    Compaction { item: CompactionItem },
 }
 
 impl std::fmt::Debug for InFlight {
@@ -50,6 +51,7 @@ impl std::fmt::Debug for InFlight {
             Self::WebSearchCall { .. } => write!(f, "InFlight::WebSearchCall {{ .. }}"),
             Self::McpCall { .. } => write!(f, "InFlight::McpCall {{ .. }}"),
             Self::McpListTools { .. } => write!(f, "InFlight::McpListTools {{ .. }}"),
+            Self::Compaction { .. } => write!(f, "InFlight::Compaction {{ .. }}"),
         }
     }
 }
@@ -87,6 +89,7 @@ impl InFlight {
             Self::WebSearchCall { item } => item.map(OutputItem::WebSearchCall),
             Self::McpCall { item } => Some(OutputItem::McpCall(item)),
             Self::McpListTools { item } => Some(OutputItem::McpListTools(item)),
+            Self::Compaction { item } => Some(OutputItem::Compaction(item)),
         }
     }
 }
@@ -479,7 +482,10 @@ impl ResponseAccumulator {
                 text: String::with_capacity(256),
             }),
             SSEItemType::WebSearchCall if !item_id.is_empty() => Some(InFlight::WebSearchCall { item: None }),
-            SSEItemType::Compaction | SSEItemType::WebSearchCall => None,
+            SSEItemType::Compaction => CompactionItem::try_from(payload)
+                .ok()
+                .map(|item| InFlight::Compaction { item }),
+            SSEItemType::WebSearchCall => None,
             SSEItemType::McpCall => McpCall::try_from(payload).ok().map(|item| InFlight::McpCall { item }),
             SSEItemType::McpListTools => McpListTools::try_from(payload)
                 .ok()
@@ -532,6 +538,7 @@ impl ResponseAccumulator {
                 (InFlight::CustomToolCall { item, input }, _) => item.apply_done(payload, input),
                 (InFlight::McpCall { item }, _) => item.apply_done(payload, &mut String::new()),
                 (InFlight::McpListTools { item }, _) => item.apply_done(payload, &mut String::new()),
+                (InFlight::Compaction { item }, _) => item.apply_done(payload, &mut String::new()),
                 (InFlight::WebSearchCall { item }, Some(OutputItem::WebSearchCall(mut call))) => {
                     if call.id.is_empty() {
                         call.id = in_flight_key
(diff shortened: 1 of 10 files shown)


crates/agentic-server-core/src/executor/engine.rs
```diff
@@ -93,13 +96,20 @@ impl<T> Drop for AbortOnDrop<T> {
     }
 }
 
+#[allow(
+    clippy::too_many_lines,
+    reason = "round-loop orchestrator keeps the full gateway tool lifecycle in one function"
+)]
```
maralbahari: Please remove this `#[allow(clippy::too_many_lines)]` and refactor instead of suppressing the warning. Follow the existing helper pattern in this file, such as `execute_and_emit_round_output_calls` and `finalize_loop`. For example, keep `run_until_gateway_tools_complete` as the dispatcher and move the normal round loop into a focused `run_gateway_tool_loop` helper. The central executor should not require a permanent Clippy exemption.

crates/agentic-server-core/src/executor/accumulator.rs
```diff
@@ -479,7 +479,7 @@ impl ResponseAccumulator {
                 text: String::with_capacity(256),
             }),
             SSEItemType::WebSearchCall if !item_id.is_empty() => Some(InFlight::WebSearchCall { item: None }),
-            SSEItemType::WebSearchCall => None,
+            SSEItemType::Compaction | SSEItemType::WebSearchCall => None,
```
maralbahari: Please handle `SSEItemType::Compaction` through the existing typed in-flight accumulator instead of returning `None` here.

The closest examples to follow are `McpCall`, `McpListTools`, and `FunctionToolCall`: each has a `TryFrom<&EventPayload>` implementation, an `InFlight` variant, handling in `start_output_item` and `complete_call_item`, and conversion into its final `OutputItem`.

Compaction should follow the same lifecycle:

1. Add `InFlight::Compaction`.
2. Implement `TryFrom<&EventPayload>` for `CompactionItem`.
3. Create it in `start_output_item`.
 4. Populate it from `OutputItemDone` in `complete_call_item`.
 5. Finalize it as `OutputItem::Compaction`.
 6. Include it in `in_flight_matches_call_type` and the completed-item fallback.

Recognizing `SSEItemType::Compaction` while excluding it from `ResponseAccumulator` leaves the streaming support incomplete and forces the engine to implement a separate output-item lifecycle.

crates/agentic-server-core/src/executor/engine.rs
```diff
@@ -211,6 +221,87 @@ async fn run_until_gateway_tools_complete(
...
+        emit_compaction_stream_events(accumulator, sender, &payload)?;
+    }
+    Ok((payload, ctx))
+}
+
+/// Emit the lifecycle and compaction-item events codex's V2 collector consumes:
+/// `response.created`/`response.in_progress`, then one `output_item.added` and
+/// `output_item.done` pair carrying the `compaction` item. `response.completed`
+/// is produced by the normal terminal framing from the returned payload.
+fn emit_compaction_stream_events(
```
maralbahari: `emit_compaction_stream_events` can be omitted entirely. The existing `ResponseAccumulator` and gateway event-plan design should handle the output-item lifecycle; `run_compaction_trigger` should contain only the compaction operation and the resulting `RequestContext`/payload state updates.

The closest example to follow is the gateway-generated MCP discovery flow:
- `mcp_list_tools_event_plans` in `executor/gateway.rs` creates plans for gateway-generated output items. - `emit_gateway_start_events` emits `response.output_item.added`.
 - `emit_gateway_completed_events` emits the completed item and `response.output_item.done`.
 - `emit_mcp_discovery_lifecycle` in `executor/upstream.rs` composes those helpers without manually calling `synthetic_event` or `emit_sse_frame`.

Please add an equivalent compaction event-plan constructor in `gateway.rs`, with the compaction item as both `started_output` and `completed_output`. Because compaction has no item-specific `*.completed` event, refactor `emit_gateway_completed_events` so `OutputItem::Compaction` skips that intermediate event but still emits `response.output_item.done`.

Once `CompactionItem` is handled through `InFlight::Compaction` in `ResponseAccumulator`, the added/done pair will also follow the same typed accumulation lifecycle as the other output items.

The direct calls to `synthetic_event`, `serialize_to_value`, and `emit_sse_frame` in `engine.rs` are therefore unnecessary. Serialization, output indices, sequence numbering, accumulator processing, and SSE forwarding already belong to the shared gateway and accumulator layers.

Please keep `run_compaction_trigger` limited to running `compact_items`, updating `ctx.new_input_items`, and constructing or updating the resulting response state. It should pass the compaction output to the shared event machinery rather than defining a separate streaming lifecycle in `engine.rs`.

crates/agentic-server-core/src/types/io/input.rs
```diff
@@ -242,6 +248,11 @@ impl InputItem {
     pub(crate) fn is_unknown(&self) -> bool {
         matches!(self, Self::Unknown)
     }
+
+    #[must_use]
+    pub fn is_compaction_trigger(&self) -> bool {
```
maralbahari: `is_compaction_trigger`  is only used inside `agentic-server-core`, so please make it `pub(crate)` rather than exposing another public API method.

crates/agentic-server-core/src/executor/engine.rs
```diff
@@ -93,13 +96,20 @@ impl<T> Drop for AbortOnDrop<T> {
     }
 }
 
+#[allow(
+    clippy::too_many_lines,
```
maralbahari: Please remove this `#[allow(clippy::too_many_lines)]` and refactor instead of suppressing the warning. Follow the existing helper pattern in this file, such as `execute_and_emit_round_output_calls` and `finalize_loop`. For example, keep `run_until_gateway_tools_complete` as the dispatcher and move the normal round loop into a focused `run_gateway_tool_loop` helper. The central executor should not require a permanent Clippy exemption.

maralbahari (commented):
Thanks for the work. left some inline comments

crates/agentic-server-core/src/executor/accumulator.rs
noobHappylife: Addressed. Compaction now follows the existing typed accumulator lifecycle:
- Added InFlight::Compaction.
- Added TryFrom<&EventPayload> and ApplyDone for CompactionItem.
- Added handling in start_output_item, complete_call_item, finalization, type matching, and the completed-item fallback.
I also added coverage for both the normal added/done lifecycle and the done-only fallback.

crates/agentic-server-core/src/executor/engine.rs
noobHappylife: Addressed. I removed emit_compaction_stream_events entirely and added compaction_event_plans alongside the existing gateway event-plan constructors.
The compaction item is used as both started_output and completed_output. emit_gateway_completed_events now skips the item-specific intermediate event for compaction, while still emitting response.output_item.done. run_compaction_trigger is now stream-agnostic and only performs compaction and updates the response/request state. The dispatcher passes its output through the shared lifecycle machinery. 

I also added a test that feeds the emitted added/done events through ResponseAccumulator to verify the layers integrate correctly.

crates/agentic-server-core/src/types/io/input.rs
noobHappylife: Addressed. InputItem::is_compaction_trigger is now pub(crate). The externally used ResponsesInput::has_compaction_trigger remains public.

crates/agentic-server-core/src/executor/engine.rs
noobHappylife: Addressed. I removed the Clippy suppression and extracted the normal round loop into run_gateway_tool_loop. run_until_gateway_tools_complete is now a small dispatcher between the compaction-trigger path and the existing gateway tool loop.

maralbahari (approved):

PR #186, opened 2026-08-20
feat: add client tool search support

haoshan98:
## Summary

Add client-executed tool search support to the Responses API.

- Model `tool_search_call` as a first-class item for blocking and streaming responses.
- Normalize public `tool_search` declarations into upstream-compatible function tools.
- Support deferred standalone functions and namespace members.
- Preserve public tool-search state across replay, persistence, and compaction.
- Add provider-parity cassettes for OpenAI, direct upstream, HTTP/SSE, and WebSocket flows.


crates/agentic-server-core/src/executor/engine.rs
```diff
@@ -562,7 +642,7 @@ impl ExecuteRequest {
             tools = self.payload.tools.as_ref().map_or(0, Vec::len),
             "executor received responses request"
         );
-        let ctx = rehydrate_conversation(self.payload, &self.exec_ctx).await?;
+        let ctx = rehydrate_for_execution(self.payload, &self.exec_ctx).await?;
```
maralbahari: Why is the established `rehydrate_conversation` entry point replaced by a nearly synonymous rehydrate_for_execution wrapper? The new name hides a tool-search preparation side effect and leaves two APIs that return contexts with different readiness invariants. Restore the existing rehydration call and perform tool-search preparation explicitly through the handler/registry seam after rehydration.

crates/agentic-server-core/src/executor/engine.rs
```diff
@@ -481,6 +518,49 @@ fn run_stream(ctx: RequestContext, exec_ctx: Arc<ExecutionContext>, auth: Option
...
+            conversation_id: ctx.conversation_id.clone(),
+            model: ctx.enriched_request.model.clone(),
+            previous_response_id: ctx.original_request.previous_response_id.clone(),
+            instructions: ctx.original_request.instructions.clone(),
+        }
+    }
+}
+
+impl StreamFailureContext {
+    fn failed_payload(&self, error: &crate::executor::error::ExecutorError) -> ResponsePayload {
```
maralbahari: use on top file import for `crate::executor::error::ExecutorError`

crates/agentic-server-core/src/executor/compaction.rs
```diff
@@ -198,12 +201,15 @@ pub(crate) async fn compact_items(
     let ctx = RequestContext {
         original_request,
         enriched_request,
+        tool_search_state: None,
+        tool_search_private_request: None,
+        tool_search_loaded_tools: None,
```
maralbahari: a tool type context should not be handled in requestContext in this format. please revisit the design. follow ARCHITECTURE.md and add a tool search handler under `agentic-server-core/src/tools`

crates/agentic-server-core/src/executor/accumulator.rs
```diff
@@ -281,11 +286,15 @@ impl ResponseAccumulator {
         line: &str,
         translator: &mut FunctionSseTranslator,
     ) -> ExecutorResult<Option<FunctionSseTranslation>> {
-        let Some(frame) = self.process_sse_line(line) else {
+        let Some(frame) = normalize_sse_line(line) else {
```
maralbahari: preserve the original code `self.process_sse_line`.
This duplicates the normalization, terminal capture, and event accumulation performed by self.process_sse_line, allowing the translator path to drift from normal accumulator behavior. Pre-accumulation validation is reasonable, but it should not bypass the established processing seam. Extract a private normalized-frame helper used by both paths, while keeping process_sse_line as the canonical parse/accumulate entry point.

crates/agentic-server-core/src/executor/accumulator.rs
```diff
@@ -155,8 +155,13 @@ impl ResponseAccumulator {
...
     /// Returns `ExecutorError::ParseError` if JSON parsing fails or required fields are missing.
     pub fn from_json(body: &str, conversation_id: Option<&str>) -> ExecutorResult<Self> {
-        let mut json: serde_json::Value = deserialize_from_str(body).map_err(ExecutorError::JsonError)?;
+        let json: serde_json::Value = deserialize_from_str(body).map_err(ExecutorError::JsonError)?;
+        Self::from_value(json, conversation_id)
+    }
 
+    /// Rehydrate a parsed non-streaming response without parsing the body a
+    /// second time after raw protocol validation.
+    pub(super) fn from_value(mut json: serde_json::Value, conversation_id: Option<&str>) -> ExecutorResult<Self> {
```
maralbahari: The logic in from_value belongs in from_json; adding a pub(super) constructor exposes raw serde_json::Value specifically so another module can mutate protocol JSON before accumulation. ARCHITECTURE.md line 288 explicitly freezes this surface. Preserve the existing API signature and, if parsing must be shared, use a private free function or typed normalization rather than a new cross-module constructor.

crates/agentic-server-core/src/executor/function_sse.rs
```diff
@@ -161,7 +419,7 @@ impl FunctionSseTranslator {
 
     fn translate_delta(
         &mut self,
-        _item_id: &str,
+        item_id: &str,
```
maralbahari, line 244: why this change is needed?

crates/agentic-server-core/src/tool/names.rs
```diff
@@ -0,0 +1,153 @@
+use std::collections::HashMap;
+use std::collections::hash_map::Entry;
+
+use crate::types::tools::{CodexNamespaceMember, ResponsesTool};
+
+use super::ToolError;
+
+pub const MODEL_VISIBLE_NAMESPACE_MEMBER_PREFIX: &str = "agentic_ns__";
```
maralbahari: mcp on `agentic-api` gateway is not behaving like codex. codex behavior of converting mcp into namespaces are custom to codex and it is separated from gateway's mcp.

crates/agentic-server-core/src/tool/normalize.rs
```diff
@@ -28,6 +28,19 @@ impl ResponsesTool {
                     "function tool config serialization failed".to_owned(),
                 )),
             ),
+            Self::ToolSearch(param) => {
```
maralbahari: please follow the `ARCHITECTURE.md` to create a ToolSearchHandler interface that implements normalize and validation functions.

crates/agentic-server-core/src/storage/conversation.rs
```diff
@@ -89,7 +90,24 @@ impl ConversationStore {
     /// Returns an error if a stored item is missing its sequence number or if the database query fails.
     pub async fn rehydrate_snapshot(&self, conversation_id: &str) -> StoreResult<ConversationSnapshot> {
         let pool = self.pool()?;
-        let rows = item::get_items_by_conversation(pool, conversation_id).await?;
+        let mut tx = pool.begin().await?;
+        if DatabaseBackend::from_connection(tx.as_mut()) == DatabaseBackend::Postgres {
```
maralbahari: please revert back to original. 
rehydrate_snapshot previously used` item::get_items_by_conversation` directly. This change adds a pool checkout, transaction, runtime backend branch, PostgreSQL isolation statement, second query over conversation turns, and commit to every rehydration solely to recover tool-search metadata. That is a hot-path regression. Please restore the original item query and move tool-search-specific metadata retrieval behind its handler/storage seam, or use a targeted latest-response lookup that does not expand every conversation rehydration.

crates/agentic-server-core/src/executor/request.rs
```diff
@@ -26,6 +28,16 @@ pub struct RequestContext {
     /// Enriched request with rehydrated conversation history injected into `.input`.
     /// This is the request forwarded to the LLM.
     pub enriched_request: RequestPayload,
+    /// Pure request-scoped tool-search views prepared after full rehydration.
+    /// Public state remains separate from the private model request.
+    pub tool_search_state: Option<ToolSearchState>,
+    /// Private inference request prepared once for active tool search.
+    /// Blocking and streaming execution consume this exact instance while
+    /// `enriched_request` remains the public representation.
+    pub tool_search_private_request: Option<Box<RequestPayload>>,
```
maralbahari: RequestContext now directly owns tool_search_state, tool_search_private_request, and tool_search_loaded_tools, requiring tool-specific initialization and branching throughout rehydration, compaction, persistence, and execution. The repeated None initializers are evidence that the tool concern is leaking into the generic request lifecycle. Per ARCHITECTURE.md line274 , keep this context limited to per-turn request and identifier state. Encapsulate search state and private lowering behind a request-scoped ToolSearchHandler/registry-owned object.

crates/agentic-server-core/src/types/tools/params.rs
```diff
@@ -143,11 +145,40 @@ pub struct CustomToolParam {
     pub extra: HashMap<String, Value>,
 }
 
+/// Only client-executed tool search is part of the public gateway contract.
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum ToolSearchExecution {
+    #[default]
+    Client,
```
maralbahari, line 155: add a TODO to support server execution type for gateway built-in tool.

crates/agentic-server-core/src/types/tools/params.rs
```diff
@@ -143,11 +145,40 @@ pub struct CustomToolParam {
...
+    pub description: String,
+    pub parameters: serde_json::Map<String, Value>,
+}
+
 /// Parameters for a gateway MCP built-in tool declaration.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct McpToolParam {
     pub server_label: String,
     #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub server_description: Option<String>,
```
maralbahari: remove this is custom to codex not mcp openai contract

crates/agentic-server-core/src/types/tools/params.rs
```diff
@@ -143,11 +145,40 @@ pub struct CustomToolParam {
...
+    Completed,
+}
+
+/// Parameters for a client-executed tool-search declaration.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct ToolSearchToolParam {
+    pub execution: ToolSearchExecution,
+    pub description: String,
+    pub parameters: serde_json::Map<String, Value>,
```
maralbahari: OpenAI defines description and parameters as optional, but this public wire type requires both. Consequently, a documented client declaration such as {"type":"tool_search","execution":"client"} fails deserialization before state preparation. Keep these fields optional and apply any description/schema required by vLLM only when constructing the private synthetic function. See the official [tool-search declaration schema](https://developers.openai.com/api/reference/cli/__sdk_schema?declaration=%28resource%29+beta.responses+%3E+%28model%29+beta_tool_search_tool+%3E+%28schema%29&selected=%28resource%29+beta.responses).

crates/agentic-server-core/src/types/tools/params.rs
```diff
@@ -143,11 +145,40 @@ pub struct CustomToolParam {
...
+}
+
+/// Terminal tool-search calls and outputs are accepted only after completion.
+/// Streaming `in_progress` state belongs to event lifecycle payloads rather
+/// than this persisted public-item status.
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum ToolSearchStatus {
+    #[default]
+    Completed,
```
maralbahari, line 165: ToolSearchStatus is used to deserialize replayed calls and client outputs but accepts only completed. OpenAI permits in_progress, completed, and incomplete for both objects, so documented payloads currently fail during deserialization. Add all variants and let the behavioral layer decide that only a completed output may load definitions. See the [call schema](https://developers.openai.com/api/reference/cli/__sdk_schema?declaration=%28resource%29+beta.responses+%3E+%28model%29+beta_response_tool_search_call+%3E+%28schema%29&selected=%28resource%29+beta.responses) and [output schema](https://developers.openai.com/api/reference/cli/__sdk_schema?declaration=%28resource%29+beta.responses+%3E+%28model%29+beta_response_tool_search_output_item+%3E+%28schema%29&selected=%28resource%29+beta.responses).

crates/agentic-server-core/src/executor/engine.rs
```diff
@@ -153,22 +149,20 @@ async fn run_gateway_tool_loop(
...
-            if let OutputItem::CustomToolCall(call) = item {
-                debug!(
-                    response_id = %ctx.response_id,
-                    call_id = %call.call_id,
-                    name = %call.name,
-                    input_bytes = call.input.len(),
-                    "custom tool call requires client execution"
-                );
-            }
+        if matches!(payload.status.as_str(), "error" | "failed" | "incomplete") {
```
maralbahari: This return precedes execute_and_emit_round_output_calls and applies to every gateway tool, not just unfinished tool search. An incomplete response can still contain an individually completed MCP or web-search call; it will now be neither executed nor processed through public_output_items, potentially exposing the normalized internal function_call and leaving no persisted call/output pair. For incomplete, discard only unfinished search placeholders, process completed gateway calls, and then return the original incomplete terminal status without starting another inference round.

crates/agentic-server-core/src/types/request_response.rs
```diff
@@ -106,6 +115,116 @@ where
...
+    /// Validate the retained public tool-search contract without lowering or
+    /// executing any declaration.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`ToolError::Config`] for invalid cardinality, reserved-name or
+    /// serial-execution conflicts, malformed replay items, and unsupported
+    /// dynamically returned declarations.
+    pub(crate) fn tool_search_readiness(&self) -> Result<ToolSearchReadiness, ToolError> {
+        self.tool_search_readiness_for_input(&self.input)
```
maralbahari: This block owns activation, cardinality, parallel-execution, reserved-name, namespace, replay-order, and dynamically loaded-tool policy. The architecture defines types/ as wire shapes and shape conversion, with tool meaning and behavior owned by tool. Move this state-readiness policy into ToolSearchState or a tool-search handler and leave only serde/shape responsibilities here.

crates/agentic-server-core/src/types/request_response.rs
```diff
@@ -42,6 +42,15 @@ fn default_true() -> bool {
     true
 }
 
+/// Structural readiness result returned after validating public tool-search
+/// declarations and replay-item wire shapes. The deterministic request-scoped
+/// state builder consumes this result and owns ordered-history semantics.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum ToolSearchReadiness {
```
maralbahari: please create a `tool_search.rs` in `src/tool` and keep all related enums interfaces in its place along with `ToolSearchHandler` interface

crates/agentic-server-core/src/tool/registry.rs
```diff
@@ -198,17 +213,21 @@ impl ToolRegistry {
...
+        let resolved_tools = CodexNamespaceHandler::resolve_namespace_members_after_validation(tools);
         McpHandler::validate_server_labels(&resolved_tools)?;
 
         for (index, tool) in resolved_tools.iter().enumerate() {
             match tool {
                 ResponsesTool::Function(p) => {
                     insert_unique_tool_entries(&mut entries, |resolved| insert_function_entry(resolved, p))?;
                 }
+                ResponsesTool::ToolSearch(_) => {
+                    tracing::debug!("client-executed tool_search declaration is omitted from the dispatch registry");
```
maralbahari: tool registry is not about client-executed tool or not. is about establishing a lookup table of available tool either owned by client or owned by gateway.

crates/agentic-server-core/src/tool/mcp/handler.rs
```diff
@@ -55,6 +55,19 @@ impl McpToolMap {
             .values()
             .any(|tool_ref| tool_ref.server_label == server_label)
     }
+
+    pub(crate) fn resolves_call_before_load(
```
maralbahari: why there are changes to mcp path? if the tool search in this PR are only to support client executable tools. means mcp path should not change

crates/agentic-server-core/src/tool/normalize.rs
```diff
@@ -97,6 +111,10 @@ impl ResponsesTool {
                 |param| FunctionHandler.normalize(&param).into_iter().take(1).collect(),
                 vec![],
             ),
+            Self::ToolSearch(_) => {
+                tracing::debug!("tool_search declaration skipped until request-scoped state preparation");
```
maralbahari: we must normalize the `tool_search` type as well to function type. as evidence on the cassettes recording, for vllm direct the `tool_search`  is sent as `function` type with name `tool_search` whereas for openai tool schema is different. the gateway agentic_api must be openai compatible meaning the same json tool type as `tool_search` type just like openai then must be recognized by gateway and then be normalized into function type before vLLM inference. please follow the pattern of creating ToolSearchHandler with `normalize` and `validate` function.

crates/agentic-server-core/src/executor/engine.rs
maralbahari: This unconditionally skips `execute_and_emit_round_output_calls`/`classify_round` for
*any* upstream response reporting one of these three statuses — not just tool-search
cases. (`upstream.rs` strips only unfinished *tool-search* function-call ids for this
same status set — see 1.4 — but this engine-level bypass applies to all requests,
which is the inconsistency worth fixing.)

A normal (non-tool-search) request declares a gateway-owned tool (`web_search` or an
MCP tool). The model emits a fully-formed, completed tool call in the same round the
upstream marks `status: "incomplete"` (e.g. truncated by `max_output_tokens` right
after finishing the call). Previously, the engine always executed gateway-owned calls
in `current_output` regardless of `payload.status`, then applied `classify_round`
(which could still mark the round `Incomplete` for its own reasons). Now the completed
gateway call is returned to the client un-executed — no tool output, no error, just a
dangling call with nothing resolving it, because the whole round is short-circuited
before `execute_and_emit_round_output_calls` runs.

crates/agentic-server-core/src/types/tools/params.rs
haoshan98: It's a valid field under https://api.openai.com/v1/responses

crates/agentic-server-core/src/storage/models/response.rs
```diff
@@ -67,6 +67,33 @@ pub async fn get(pool: &DbPool, id: &str) -> DbResult<Option<Response>> {
...
+/// # Errors
+/// Returns `DbResult::Err` if the database query fails.
+pub async fn get_conversation_turn_for_item(
+    pool: &DbPool,
+    conversation_id: &str,
+    item_id: &str,
+) -> DbResult<Option<Response>> {
+    let escaped_item_id = item_id.replace('!', "!!").replace('%', "!%").replace('_', "!_");
+    let history_suffix = format!("%\"{escaped_item_id}\"]");
+    sqlx::query_as::<_, Response>(
```
maralbahari, line 79: I understand why we need metadata from the response associated with the captured conversation version: compaction can remove the `tool_search_call`/`tool_search_output` history, and selecting the latest response would be unsafe if another turn is persisted after rehydration.

However, this query reconstructs the response-to-item relationship by applying `LIKE` to the serialized `history_item_ids` JSON. This is unindexed, depends on the exact JSON encoding, and suggests that the storage schema is missing an explicit relationship.

Could we model this association directly—for example, with a `response_id`/`turn_id` on stored items or a normalized `response_items` relation—and resolve it using an indexed lookup? At minimum, `ConversationSnapshot` could retain the last item ID already loaded by `rehydrate_snapshot`, avoiding the additional `get_id_by_conversation_sequence` query. I would keep the exact-version behavior but avoid introducing CRUD based on textual matching over JSON.

crates/agentic-server/src/handler/http/responses.rs
```diff
@@ -53,11 +54,13 @@ pub async fn responses(State(state): State<AppState>, req: Request) -> Response
         Err(e) => return e,
     };
 
+    let has_tool_search_state = ToolRegistry::request_has_tool_search_state(&payload);
     let should_execute = payload.store
         || payload.previous_response_id.is_some()
         || payload.conversation_id.is_some()
         || payload.input.contains_compaction()
         || payload.input.has_compaction_trigger()
+        || has_tool_search_state
```
maralbahari, line 59: why tool search state needs to be part of `should_execute`? sounds unnecessary

crates/agentic-server-core/src/executor/engine.rs
```diff
@@ -97,41 +98,42 @@ impl<T> Drop for AbortOnDrop<T> {
...
             let event_plans = compaction_event_plans(&payload.output, 0);
             emit_gateway_start_events(&event_plans, stream_accumulator, stream_sender)?;
             emit_gateway_completed_events(&payload.output, &event_plans, stream_accumulator, stream_sender)?;
         }
-        return Ok((payload, ctx));
+        return Ok((payload, ctx, registry));
     }
 
-    run_gateway_tool_loop(ctx, exec_ctx, auth, stream_upstream, stream).await
+    run_gateway_tool_loop(ctx, registry, exec_ctx, auth, stream_upstream, stream).await
```
maralbahari: registry is being passed down to multiple-functions in `src/executor` without that function even needing to access registry all they do is to pass it down to another function. perhaps just pass to exactly where it is needed. it's hard to track now.
like the example in this function. or `persist_if_needed`

crates/agentic-server-core/src/executor/engine.rs
```diff
@@ -97,41 +98,42 @@ impl<T> Drop for AbortOnDrop<T> {
...
-) -> ExecutorResult<(ResponsePayload, RequestContext)> {
+) -> ExecutorResult<(ResponsePayload, RequestContext, ToolRegistry)> {
     let mut executors = exec_ctx.gateway_executors.request_scoped();
-    let registry: ToolRegistry = match ctx.enriched_request.tools.as_mut() {
-        Some(tools) => ToolRegistry::build_with_handlers(tools, &mut executors).await?,
-        None => ToolRegistry::default(),
-    };
-    let mut combined_output: Vec<OutputItem> = registry
+    let mut registry = registry
+        .build_prepared_with_handlers(ctx.enriched_request.tools.as_mut(), &mut executors)
```
maralbahari: this function builds the registry here so only the entries (tool lookup table) is created only here but why is it receiving registry:ToolRegistry instance argument?

crates/agentic-server-core/src/tool/registry.rs
```diff
@@ -268,23 +296,196 @@ impl ToolRegistry {
...
+            state.map_or(&empty, ToolSearchState::withheld_function_names),
+        )
+    }
+
+    /// Ensure tool-search requests went through the request-scoped preparation seam.
+    pub(crate) fn ensure_request_prepared(&self, request: &RequestPayload) -> Result<(), ToolError> {
+        ensure_request_prepared(request, self.tool_search.is_some())
+    }
+
+    pub(crate) fn normalize_response_output(
```
maralbahari: this function should be part of `src/executor` module. not sure its relevency to registry?

crates/agentic-server-core/src/tool/registry.rs
```diff
@@ -268,23 +296,196 @@ impl ToolRegistry {
...
             custom_tool_map,
             mcp_tool_map,
             mcp_list_tools_items,
         })
     }
 
+    /// Prepare the request's private inference projection, build the normal
+    /// dispatch table, and retain the public tool-search projection in this
+    /// request-scoped registry.
+    pub(crate) fn prepare_request(
```
maralbahari: I think this function can also exist in `executor` module after registry is created by executor module then only call `registry.install_tool_search_state()` instead of creating registry instance  in a function that is part of registry.

crates/agentic-server-core/src/storage/conversation.rs
```diff
@@ -105,6 +105,27 @@ impl ConversationStore {
...
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if either targeted database lookup fails.
+    pub async fn response_metadata_at_version(
+        &self,
+        conversation_id: &str,
+        version: ConversationVersion,
+    ) -> StoreResult<Option<ResponseMetadata>> {
+        let ConversationVersion::LastSequence(sequence) = version else {
```
franciscojavierarceo: `ConversationVersion::Empty` returns no metadata here, but a WebSocket `generate:false` prewarm with empty input persists response metadata without advancing the item sequence. the next conversation request therefore loses the prewarmed tool-search state, or can inherit stale metadata from an earlier item-bearing turn. we should track the latest response independently of item sequence so zero-item turns remain addressable.

crates/agentic-server-core/src/tool/tool_search.rs
```diff
@@ -0,0 +1,2207 @@
...
+}
+
+fn tool_has_deferred_definition(tool: &ResponsesTool) -> bool {
+    match tool {
+        ResponsesTool::Function(function) => function.defer_loading == Some(true),
+        ResponsesTool::Namespace(namespace) => namespace.tools.iter().any(
+            |member| matches!(member, CodexNamespaceMember::Function(function) if function.defer_loading == Some(true)),
+        ),
+        ResponsesTool::ToolSearch(_)
+        | ResponsesTool::Mcp(_)
```
franciscojavierarceo: MCP and custom declarations with `defer_loading:true` are accepted but treated as immediately available, so the model can invoke them before the client returns them from tool search. the [Responses tool schema](https://developers.openai.com/api/reference/cli/__sdk_schema?declaration=%28resource%29+beta.responses+%3E+%28model%29+beta_tool+%3E+%28schema%29&selected=%28resource%29+beta.responses) supports deferred MCP and custom tools. we should implement withholding for these variants or reject unsupported deferred kinds before discovery and normalization.

crates/agentic-server-core/src/tool/tool_search.rs
```diff
@@ -0,0 +1,2207 @@
...
+        let mut normalized = param.clone();
+        normalized.description = Some(
+            param
+                .description
+                .as_deref()
+                .filter(|description| !description.trim().is_empty())
+                .unwrap_or(DEFAULT_DESCRIPTION)
+                .to_owned(),
+        );
+        normalized.parameters = Some(
```
franciscojavierarceo, line 122: a declared `tool_search.parameters` schema without top-level `"type":"object"`, such as `oneOf`, is silently replaced with the built-in `query` schema. other object schemas are forced into `strict:true` lowering without strict-schema validation. this can produce arguments that do not match the client’s declaration. we should preserve the declared schema or return a clear 400 when the upstream lowering cannot support it.

crates/agentic-server-core/src/tool/tool_search.rs
```diff
@@ -0,0 +1,2207 @@
...
+    let value: Value = match deserialize_from_str(body) {
+        Ok(value) => value,
+        Err(_) => return Ok(()),
+    };
+    let status = value
+        .get("status")
+        .and_then(Value::as_str)
+        .map_or(ResponseStatus::Completed, |status| status.parse().unwrap_or_default());
+    let discard_unfinished = matches!(status, ResponseStatus::Error | ResponseStatus::Incomplete);
+    for item in value.get("output").and_then(Value::as_array).into_iter().flatten() {
```
franciscojavierarceo, line 903: this validates each tool-search call independently, so an upstream response containing two completed calls is exposed and persisted. continuation then rehydrates `call1, call2` before either client output and rejects `call2` because `call1` remains unresolved. we should reject more than one search call before exposing the response, or support multiple pending calls throughout replay.

crates/agentic-server-core/src/tool/tool_search.rs
```diff
@@ -0,0 +1,2207 @@
...
+
+    /// Normalize native and synthetic upstream tool-search calls into the
+    /// canonical public output item.
+    pub(crate) fn normalize_response_output(
+        registry: &ToolRegistry,
+        output: &mut Vec<OutputItem>,
+        status: ResponseStatus,
+        unfinished_stream_item_ids: &HashSet<String>,
+    ) -> Result<(), ToolError> {
+        let discard_unfinished = matches!(status, ResponseStatus::Error | ResponseStatus::Incomplete);
```
franciscojavierarceo: `ResponseStatus::Incomplete` is treated like an error here, which drops incomplete or still-in-progress search calls from the terminal response and persistence. streaming clients can observe `response.output_item.added` and then receive a terminal response without that item, while blocking clients never see it. we should preserve the public item with its documented incomplete status while keeping it non-replayable as a completed call.

crates/agentic-server-core/src/storage/models/response.rs
```diff
@@ -67,6 +67,33 @@ pub async fn get(pool: &DbPool, id: &str) -> DbResult<Option<Response>> {
...
+        "SELECT * FROM responses \
+         WHERE conversation_id = $1 \
+           AND previous_response_id IS NULL \
+           AND history_item_ids LIKE $2 ESCAPE '!' \
```
franciscojavierarceo: this leading-wildcard `LIKE` over serialized `history_item_ids` cannot use an index and depends on the exact JSON encoding. compacted continuation metadata lookup therefore grows linearly with the number of stored conversation turns. we should store an indexed `last_item_id` or normalize the response-to-item relationship.

crates/agentic-server-core/tests/relayed_stream_validation_test.rs
```diff
@@ -282,7 +282,7 @@ fn strict_relay_decoder_accepts_compatible_recorded_responses_streams() {
     );
     assert_eq!(
         rejected.len(),
-        10,
+        13,
```
franciscojavierarceo: the [current Rust job](https://github.com/vllm-project/agentic-api/actions/runs/34138820938/job/101795919389) still finds 10 incompatible recorded streams, so changing this expectation to 13 fails `strict_relay_decoder_accepts_compatible_recorded_responses_streams` and stops the suite. we should reconcile this baseline with the checked-in cassettes and rerun the test before merging.

franciscojavierarceo (commented):
the earlier zero-item persistence, indexed lookup, deferred-tool validation, schema preservation, and incomplete-call fixes are present. i found one remaining verification issue in the current Rust job, noted inline. this was a source and CI review; i haven't rerun the full suite locally.

maralbahari (approved):

franciscojavierarceo (approved):

PR #189, opened 2026-08-20
fix: harden MCP and Agentic API config startup

franciscojavierarceo:
## Summary

Follow-up to #180.

- defer configured MCP connection and tool discovery until a tool is used
- prevent unavailable MCP services from blocking Agentic API startup
- replace process-global MCP host allowlist state with per-executor configuration
- make generated config writes synced, atomic, and non-overwriting
- add regression tests for unavailable MCP servers and custom allowlists

## Test Plan

- cargo test -p agentic-server-core tool:: --lib
- cargo clippy -p agentic-server-core --all-targets -- -D warnings
- cargo fmt --all
- git diff --check

Related: #180.


franciscojavierarceo (commented):
Self-review notes — clean PR, achieves all three goals (deferred discovery, per-executor allowlist, atomic config writes) without correctness issues.

Two observations for the record:

**Per-request tool rediscovery:** `mcp_clients` is shared across request clones via `Arc<RwLock<…>>`, but `self.mcp` (discovered handlers) is a plain `HashMap` cloned per request. This means `discover_tools` (a `tools/list` RPC) runs once per request per configured server, not once at startup. The client connection cache mitigates the cost, but it's a latency trade-off worth keeping in mind — if it matters later, the handler cache could move behind `Arc<RwLock<…>>` too.

**Minor: unreachable guard in `mcp_server_tools`.** The `let Some(entry) = self.mcp_configs.get(server_label).cloned() else { … }` branch can't fire — it's guarded by `contains_key` above and `&mut self` prevents concurrent mutation. Could be an `expect` if you prefer, but no functional issue.

Tests cover the critical scenarios. Config atomicity pattern is correct.

Issue #181, opened 2026-08-14
Allow parallel function calling for gateway-owned Responses tools


maralbahari:
### Problem statement / motivation

The Responses API accepts `parallel_tool_calls`, but currently rejects
`parallel_tool_calls: true` whenever a gateway-owned tool such as
`web_search_preview` or `mcp` is declared. When the field is omitted, requests
containing such tools are normalized to `parallel_tool_calls: false`.

This restriction no longer matches the executor's capabilities:

- Multiple model-generated function calls are parsed and preserved.
- Gateway ownership is resolved independently for every call.
- Gateway-owned calls are already executed concurrently with a bounded
  concurrency of five and a per-call timeout.
- Multiple client-owned calls and mixed gateway/client rounds are already
  supported.
- `/models` advertises `supports_parallel_tool_calls: true`.

The relevant current implementation is:

- Request restriction: [`RequestPayload::to_upstream_request`](https://github.com/vllm-project/agentic-api/blob/ee07248c7e75bdee9ba0cc6c29fb3fb6930b2f27/crates/agentic-server-core/src/types/request_response.rs#L123-L134)
- Concurrent gateway dispatch: [`execute_output_calls`](https://github.com/vllm-project/agentic-api/blob/ee07248c7e75bdee9ba0cc6c29fb3fb6930b2f27/crates/agentic-server-core/src/executor/gateway.rs#L210-L232)
- Existing concurrent web-search test: [`execute_runs_multiple_web_search_calls_concurrently`](https://github.com/vllm-project/agentic-api/blob/ee07248c7e75bdee9ba0cc6c29fb3fb6930b2f27/crates/agentic-server-core/tests/web_search_tool_test.rs#L1655-L1699)
- Advertised model capability: [`supports_parallel_tool_calls`](https://github.com/vllm-project/agentic-api/blob/ee07248c7e75bdee9ba0cc6c29fb3fb6930b2f27/crates/agentic-server/src/handler/http/models.rs#L37-L45)

As a result, the public API rejects an execution mode that the tool loop can
already process.

The `parallel_tool_calls` field should control whether the model may produce
multiple tool invocations in one inference round. Gateway execution concurrency
should remain a separate, bounded server-side policy.

### 1. Permit explicit opt-in for gateway-owned tools

Change request normalization so the presence of a gateway-owned tool does not
cause `parallel_tool_calls: true` to be rejected:

| Request | Effective initial behavior |
| --- | --- |
[...]

PR #214, opened 2026-08-27
feat: add parallel tool calling support

maralbahari:
## Summary

This PR adds end-to-end parallel tool calling support to the typed Responses API (#181) and aligns gateway behavior
with the OpenAI reference behavior captured in matched cassettes.

Previously, the gateway forced `parallel_tool_calls: false` upstream and executed gateway-executed built-in tool
calls serially. This change:

- forwards the client's `parallel_tool_calls` model-generation preference for function, custom, built-in, and mixed
  tool declarations, while preserving the existing `false` default when the field is omitted;
- executes gateway-executed built-in tool calls concurrently within an inference round using a bounded sliding
  window (`tools.max_concurrent_gateway_calls`, default `5`), while preserving model call order in public output;
- scopes parallel-safety to each model-visible tool name, so a handler can serialize calls to itself without blocking
  unrelated tools in the same round; MCP and web search explicitly support same-tool parallel execution;
- isolates timeouts and individual execution failures as tool call outputs, allowing concurrently executing sibling
  calls to complete; and
- preserves streaming event order and contiguous sequence numbers while synthesizing the correct web-search and MCP
  lifecycle events around concurrent execution.

The tool registry and execution path were also reworked to make this behavior explicit. `ToolOwnership` now separates
client-executed tools from gateway-executed built-in tools, and `GatewayBinding` carries the executor and its same-tool
concurrency policy. `GatewayExecutor` owns the started and completed public-output projections for its tool type, and
`GatewayRound` owns bounded dispatch, per-call timeouts, result ordering, and output shaping. For MCP, typed handler
metadata replaces the duplicated registry side table that could drift out of sync with the main tool entries.

### OpenAI parity and correctness fixes found during cassette comparison

The parallel-call recordings exposed several independent bugs in addition to the main feature:

1. **Missing function call outputs were accepted.** A continuation could omit the
   `function_call_output` or `custom_tool_call_output` for a pending client-executed call and the gateway would
   continue with a `200`. Rehydration now detects unresolved call IDs in both stored-response and conversation
   history and rejects the continuation with a `400 invalid_request_error`, matching the reference behavior.

2. **One web-search call could not batch multiple queries.** The normalized web-search schema only accepted a
   singular `query`, despite the public `web_search_call.action` already supporting `queries`. The schema now accepts
   either form, and one call can fan out multiple independent provider requests concurrently and aggregate their
   results into one web-search output item.

3. **Mixed built-in rounds could expose an MCP call as a raw `function_call`.** This occurred when MCP and another
   gateway-executed built-in tool ran in the same round. Handler-owned public-output projection and explicit registry
   ownership now ensure MCP and web-search calls retain their correct public item types in both streaming and
   non-streaming responses.

4. **`mcp_list_tools` discovery was repeated on every turn.** The matched cassettes established an otherwise
   undocumented OpenAI behavior: an MCP server's discovery item and lifecycle are exposed on the first response in a
   stored chain, and later `previous_response_id` turns behave as though that discovery is cached rather than emitting
   it again. The gateway now retains `mcp_list_tools` in item history and lets the request-scoped registry group current
   and historical discovery records by `server_label`. Discovery is emitted only when that server has no prior list
   record, and the streaming path consumes the lifecycle after the first inference round.

   `mcp_list_tools` is history metadata, not model context: `InOutItem::into_input_items` preserves it for registry
   lifecycle decisions, while `ResponsesInput::model_input` and compaction remove it before inference. Likewise, the
   original model-facing `function_call` and `function_call_output` items are the single persisted source for executed
   web-search and MCP calls; public `web_search_call` and `mcp_call` items are no longer lossily reconstructed into a
   second copy of model input during rehydration.

### OpenAI and gateway cassette coverage

Each scenario was recorded against both the OpenAI reference endpoint and the gateway in streaming and non-streaming
modes with `parallel_tool_calls: true`.

| Cassette suite | Cases covered |
| --- | --- |
| `parallel-builtin-only` | Three-turn stored-response chain: two parallel web-search calls, with two queries batched inside each call; two calls to the same MCP tool with different arguments; and a mixed web-search + MCP round. Also verifies that `mcp_list_tools` is emitted only on the first turn. |
| `parallel-mixed` | Three-way round containing a gateway-executed web search, a client-executed function call, and a custom tool call, followed by a turn that depends on the submitted custom-tool output. |
| `parallel-client` | Three parallel client-executed calls across function and custom tool types; correlation and submission of every call output on the continuation; two calls to the same function with different arguments; and a final multi-result follow-up. |
| `parallel-failures` | A failed function call alongside a successful sibling; a continuation that omits one of two required function call outputs; and a mixed built-in + client-executed round whose client call output is omitted. The omission cases verify rejection rather than silently continuing with a dangling call ID. |

### Automated integration and unit coverage

- `parallel_tool_calls_test` loads the OpenAI and gateway `parallel-builtin-only` cassettes for both response modes and
  compares the three-turn request chain, normalized query/argument contracts, two-call grouping per turn, public
  output types and order, terminal statuses, SSE lifecycle counts, contiguous sequence numbers, and first-turn-only
  MCP discovery.
- Responses request tests cover `parallel_tool_calls: true`, `false`, and omitted behavior for client-executed,
  built-in, and mixed tool declarations, including an HTTP gateway-to-upstream integration check.
- Gateway-round tests verify real overlap for parallel-safe same-tool calls, serialization for handlers that do not
  opt into same-tool parallelism, stable model call ordering, per-call timeout isolation, and missing-provider error
  isolation.
- Web-search tests verify singular-query compatibility and multi-query fan-out/aggregation.
- Pending-call and rehydration tests cover resolved and unresolved function/custom calls for both
  `previous_response_id` and conversation continuations.
- MCP history tests verify that list-tools metadata survives storage and rehydration, is excluded from model and
  compaction input, is ignored for servers absent from the active request, and is emitted only once for blocking and
  streaming continuations.
- MCP/web-search lifecycle tests verify the public output item types and completed/failed streaming event sequences.

Commands used for focused verification:

```bash
cargo build --release
cargo test -p agentic-server-core --test parallel_tool_calls_test
cargo fmt --all -- --check
```


crates/agentic-server-core/src/executor/engine.rs
```diff
@@ -324,30 +375,30 @@ async fn execute_and_emit_ordered_output_calls(
...
+                    |item| matches!(item, OutputItem::FunctionCall(call) if registry.is_gateway_owned_name(&call.name)),
                 )
                 .count()
                 .saturating_add(start)
         });
     let first_gateway_run_len = first_gateway_run_end.saturating_sub(first_gateway_index.unwrap_or(0));
     emit_gateway_start_events(&event_plans[..first_gateway_run_len], stream_accumulator, stream_sender)?;
 
-    let gateway_results = execute_output_calls(output_items, registry).await?;
+    let gateway_results = GatewayRound::new().execute(output_items, registry).await?;
```
franciscojavierarceo: [`gateway_event_plans`](https://github.com/vllm-project/agentic-api/blob/8c38e30dd2226df31624cd902b893f86fa8f8c17/crates/agentic-server-core/src/executor/gateway.rs#L259-L280) skips `Gateway(None)` calls such as `file_search` and `code_interpreter`, while the streaming engine still counts them here. this leaves `event_plans` shorter than the gateway-call count and causes an out-of-bounds slice panic. we need to retain a positional plan for every gateway-owned call or key plans by output index.

crates/agentic-server/src/main.rs
```diff
@@ -262,6 +270,7 @@ fn build_config(llm_api_base: String, common: &CommonArgs, file: &FileConfig) ->
             mcp_servers: file.mcp_servers.clone(),
             mcp_allowed_hosts,
             messages_gateway_tool_aliases: file.messages_gateway.tool_aliases.clone(),
+            max_concurrent_gateway_calls: max_concurrent_gateway_calls as usize,
```
franciscojavierarceo: `tools.max_concurrent_gateway_calls` accepts `0` from TOML because the nonzero check only applies to environment-provided values. that reaches `.buffered(0)`, which polls no calls, so the request hangs before the per-call timeout can start. we need to validate the resolved value or represent it as `NonZeroUsize`.

crates/agentic-server-core/src/tool/web_search.rs
```diff
@@ -128,31 +132,61 @@ impl WebSearchHandler {
...
+        let provider = self
+            .provider
+            .as_ref()
+            .ok_or_else(|| ToolError::Config("web_search spec-only handler cannot execute tools".to_owned()))?;
         let args = WebSearchArguments::from_json(arguments)?;
         let config = serde_json::from_value::<WebSearchToolParam>(config.clone())
             .map_err(|e| ToolError::Config(format!("invalid web_search config: {e}")))?;
-        let response = self.provider.search(&args, &config).await?;
+        let queries = args.all_queries();
+        let responses =
```
franciscojavierarceo: the `queries` array has no length bound, and `try_join_all` starts every provider request concurrently inside one gateway slot. one model-generated call can therefore bypass `max_concurrent_gateway_calls`, launch thousands of outbound requests, and aggregate all results in memory. we need to enforce `maxItems` and execute searches through a bounded stream or shared semaphore.

crates/agentic-server-core/src/executor/gateway.rs
```diff
@@ -145,109 +105,134 @@ fn execution_error_output(call: &FunctionToolCall, message: &str) -> ExecutorRes
...
+        let calls = function_calls(output_items);
+        let gateway_calls = registry.gateway_owned(&calls);
+
+        futures_stream::iter(
+            gateway_calls
+                .into_iter()
+                .cloned()
+                .map(|call| self.run_one(call, registry)),
+        )
+        .buffered(max_concurrent_gateway_calls())
```
franciscojavierarceo: `buffered` preserves result order internally, so a slow first call prevents completed later calls from leaving the queue and stops the window from refilling. same-tool waiters also consume global slots before acquiring their permit, which can block unrelated tools for the full timeout. we need to execute unordered with original indexes, keep exclusion waiters outside the active global window, and restore output order afterward.

crates/agentic-server-core/src/executor/gateway.rs
```diff
@@ -1,27 +1,46 @@
...
+/// via [`set_max_concurrent_gateway_calls`] when `ExecutionContext` is built from
+/// config; falls back to [`DEFAULT_MAX_CONCURRENT_GATEWAY_CALLS`] otherwise (e.g.
+/// in tests that construct `ExecutionContext::new` directly).
+static MAX_CONCURRENT_GATEWAY_CALLS: OnceLock<usize> = OnceLock::new();
+
+/// Sets the process-wide gateway concurrency limit. Called once, from
+/// `ExecutionContext::from_config`. Subsequent calls are no-ops: the value is
+/// fixed for the lifetime of the process.
+pub fn set_max_concurrent_gateway_calls(value: usize) {
+    let _ = MAX_CONCURRENT_GATEWAY_CALLS.set(value);
```
franciscojavierarceo: the `OnceLock` silently discards subsequent `set` failures, so the first `ExecutionContext` fixes gateway concurrency for the entire process. later contexts can accept a different configuration while continuing to use the first value. we should store the limit on `ExecutionContext` and pass it into each `GatewayRound`.

crates/agentic-server-core/src/tool/registry.rs
```diff
@@ -299,16 +328,40 @@ impl ToolRegistry {
...
     }
 
-    pub(crate) fn mcp_tool_ref(&self, internal_name: &str) -> Option<&McpToolRef> {
-        self.mcp_tool_map.tool_ref(internal_name)
+    /// Current discovery items whose server label has no rehydrated list-tools
+    /// history. A vector of length greater than one contains the current item
+    /// followed by at least one historical item and is therefore suppressed.
+    pub(crate) fn mcp_list_tool_items(&self) -> impl Iterator<Item = &McpListTools> {
+        self.mcp_list_tools_items
+            .values()
```
franciscojavierarceo: `mcp_list_tool_items` iterates `HashMap::values()`, replacing the previous declaration-ordered representation. with multiple MCP servers, response order and streaming `output_index` assignments can change between processes. we should preserve declaration order with a `Vec` or `IndexMap`.

crates/agentic-server-core/src/executor/error.rs
```diff
@@ -125,7 +125,7 @@ impl ExecutorError {
     pub fn error_type(&self) -> &'static str {
         match self.client_visible_error() {
             Self::ConversationLocked { .. }
-            | Self::Tool(ToolError::Config(_))
+            | Self::Tool(ToolError::Config(_) | ToolError::MissingOutput { .. })
```
franciscojavierarceo, line 128: `error_param` and the error serializer make `MissingOutput` diverge from the OpenAI reference cassette: the message gains a `tool error:` prefix, `param: "input"` is omitted, and `code` becomes `"invalid_request_error"` instead of `null`. we should special-case this validation error so the public envelope matches the recorded reference.

crates/agentic-server-core/src/tool/registry.rs
```diff
@@ -299,16 +328,40 @@ impl ToolRegistry {
...
+    pub(crate) fn cache_listed_mcp_tools(&mut self, input: &ResponsesInput) {
+        if let ResponsesInput::Items(items) = input {
+            for item in items {
+                let InputItem::McpListTools(list_tools) = item else {
+                    continue;
+                };
+                // Only a currently declared MCP server can emit discovery in
+                // this request. Ignore history for labels absent from the
+                // current registry instead of turning it into a new candidate.
+                if let Some(items) = self.mcp_list_tools_items.get_mut(&list_tools.server_label) {
```
franciscojavierarceo, line 307: `cache_listed_mcp_tools` suppresses current discovery whenever history contains the same `server_label`, even when the earlier discovery failed or the declaration and tool set changed. this can hide a valid current discovery lifecycle from the client. we should suppress only an equivalent prior successful discovery.

crates/agentic-server-core/src/executor/pending_calls.rs
```diff
@@ -0,0 +1,120 @@
...
+/// appears. Namespace member calls are represented as `InputItem::FunctionCall`
+/// (their flattened name lives in the `name` field), so they're covered by
+/// the same check as plain function calls.
+pub(super) fn pending_calls(items: &[InputItem]) -> Vec<PendingCall> {
+    let mut pending = Vec::new();
+    for item in items {
+        match item {
+            InputItem::FunctionCall(call) => pending.push(call.call_id.clone()),
+            InputItem::CustomToolCall(call) => pending.push(call.call_id.clone()),
+            InputItem::FunctionCallOutput(output) => pending.retain(|call_id| *call_id != output.call_id),
```
franciscojavierarceo: `pending_calls` uses `retain`, so one output clears every pending call sharing its `call_id`; empty IDs are accepted as well. duplicate or missing IDs can therefore bypass the new one-output-per-call validation. we should reject empty and duplicate IDs and enforce one-to-one call/output matching.

maralbahari: @franciscojavierarceo  `pending_calls` now rejects empty or duplicate `call_id` values and enforces one-to-one, same-kind matching between calls and outputs. Each output removes exactly one pending entry; unmatched, repeated, or function/custom mismatched outputs return `InvalidRequest`. Added focused coverage for all invalid cases, and the full two-thread test suite passes.

crates/agentic-server-core/src/tool/registry.rs
maralbahari: @franciscojavierarceo  MCP discovery is now suppressed only when the current result is successful and history contains an equivalent successful tool set. Previous failures, changed tool definitions, and current discovery failures remain visible to the client. Comparison ignores generated lifecycle IDs while covering all public tool fields and order. Added regression tests for each case.

crates/agentic-server-core/src/executor/error.rs
maralbahari: @franciscojavierarceo this is indeed an invalid-request error: OpenAI returns HTTP 400, which the SDK exposes as `openai.BadRequestError`, while the inner error object contains `type: "invalid_request_error"` and `code: null`.

I initially assumed the gateway should repeat the error type in `code`, but to match the OpenAI response exactly, I updated the gateway to preserve HTTP 400 and `invalid_request_error` while emitting the exact message, `param: "input"`, and `code: null`.

crates/agentic-server-core/src/tool/registry.rs
maralbahari: @franciscojavierarceo `mcp_list_tools_items` now uses `IndexMap<String, Vec<McpListTools>>`, preserving MCP declaration order while retaining keyed lookup and per-server history. Blocking responses and streaming lifecycle `output_index` assignments are therefore deterministic. Added a regression test verifying insertion/declaration order is preserved.

crates/agentic-server-core/src/executor/gateway.rs
maralbahari: @franciscojavierarceo  addressed. I replaced the ordered `buffered` window with `join_all` backed by an explicit scheduler-local semaphore.

Each call now acquires its same-tool exclusion permit before acquiring global execution capacity, so exclusion waiters no longer block unrelated tools. Because `join_all` polls all planned calls while preserving input order, completed later calls immediately release their global slot without requiring index sorting afterward.

Added regression tests covering both same-tool waiter starvation and execution-window refilling behind a slow earlier call.

crates/agentic-server-core/src/tool/web_search.rs
maralbahari: @franciscojavierarceo  Addressed. Batched web search remains supported to match OpenAI behavior, but `queries` is now capped at five through both `maxItems` in the model-visible schema and runtime validation.

I also replaced `try_join_all` with an ordered bounded stream. Each provider request acquires a shared semaphore initialized from `max_concurrent_gateway_calls`, so concurrent batched calls cannot bypass the configured outbound concurrency limit. Tests cover oversized batches and concurrency across simultaneous web search calls.

crates/agentic-server/src/main.rs
maralbahari: @franciscojavierarceo Fixed. `max_concurrent_gateway_calls` is now represented as `NonZeroUsize` from TOML/environment parsing through scheduler execution, making `.buffered(0)` unrepresentable. A configured value of `0` is rejected during parsing.

I also removed the process-global `OnceLock`. Each `ExecutionContext` now owns its scheduler policy and passes it explicitly to `GatewayScheduler`. Added regression tests for TOML/environment zero rejection and enforcement of the configured scheduler window.

crates/agentic-server-core/src/executor/engine.rs
maralbahari: I addressed this by replacing the separately derived gateway-call and event-plan lists with a single `GatewayScheduler`.

The scheduler first creates one `GatewayCallPlan` for every gateway-owned call in model-output order. Each plan owns the original item index, execution binding, timeout policy, and lifecycle event plan. `Gateway(None)` is represented explicitly as `MissingHandler`, so calls such as unconfigured `file_search` and `code_interpreter` retain their positional slot and produce an error `function_call_output` instead of being skipped.

Execution uses bounded concurrency through `max_concurrent_gateway_calls`. Handlers that support parallel execution may overlap, while handlers that do not are protected by a per-tool-name semaphore. Each call retains its timeout, and results and SSE lifecycle events are emitted according to the planned output order regardless of completion order. The upstream `parallel_tool_calls` field remains a model-generation preference; it does not replace the gateway’s execution-safety policy.

I also removed the untyped `ToolEntry.config: serde_json::Value` path. `ToolHandler` now declares its public `ToolParams`, and `GatewayExecutor` declares its `ExecutionParams`. For example, MCP validates and normalizes an `McpToolParam`, but each discovered tool executes with its own `McpDiscoveredToolParam`.

`GatewayBinding::new` pairs the concrete executor with its correctly typed parameters before type-erasing that pair for storage in the heterogeneous registry. The scheduler therefore receives a complete binding and never handles raw configuration JSON or tool-specific type checks. Tool handlers remain responsible for their lifecycle projection through `GatewayToolEventPlan`, while the scheduler owns indexes, ordering, concurrency, and SSE framing.

I added `scheduler_retains_event_slot_for_gateway_tool_without_handler` as a regression test for the reported panic path.

franciscojavierarceo (approved):

docs/design/codex-integration.md
```diff
@@ -133,11 +133,11 @@ Responses tool shapes and execution semantics, so it can be always on.
 
 | Shape | Behavior |
 |-------|----------|
-| `function` | Client-owned by default. Preserve declaration and return matching calls to the client unless configured as gateway-owned. |
+| `function` | Client-owned. Preserve the declaration and return matching calls to the client. |
 | `namespace` | Client-owned Codex grouping for function tools. Flatten members only for upstream requests, then restore returned calls. |
 | `custom` | Client-owned freeform tool. Preserve its opaque format and forward it natively. |
```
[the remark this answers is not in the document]
franciscojavierarceo: Fixed in #230. I replaced the ownership wording with the project’s preferred client-executed and gateway-executed terminology across the Codex integration docs and related Rustdoc.

crates/agentic-server-core/src/types/io/output.rs
```diff
@@ -282,12 +282,16 @@ fn default_web_search_action_search_type() -> String {
...
 
 impl WebSearchActionSearch {
+    /// # Panics
+    ///
+    /// Panics if `queries` is empty.
     #[must_use]
-    pub fn new(query: impl Into<String>, sources: Vec<WebSearchSource>) -> Self {
+    pub fn new(queries: Vec<String>, sources: Vec<WebSearchSource>) -> Self {
+        let query = queries.first().expect("queries must have at least one entry").clone();
         Self {
```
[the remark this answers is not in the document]
franciscojavierarceo: Fixed in #230. I replaced both web-search constructors with fallible try_new APIs returning a typed WebSearchActionError, added the empty-query regression case, and kept the valid wire shape covered.

PR #216, opened 2026-08-27
Add hydrate and persist endpoints.

### LINKED ISSUE #215: Add `llm-d` coordinator integration: hydrate and persist endpoints

Let `llm-d` use `agentic-api` for Responses state without routing model traffic through the gateway. A stateful turn is exposed as two separately callable steps: `hydrate` before inference, `persist` after, so the caller makes the model
call itself.

This approach represents a separation of concerns between `agentic-api` which is responsible for all state management, hydration, etc., and `llm-d` coordinator, which is responsible for the AI request and response pipelines.

### Proposed solution

- `hydrate`: resolve `previous_response_id` and return the upstream request with history inlined and continuation and storage fields removed, plus an opaque context for the turn.
- `persist`: take that context and the model's response, store the turn, and return the response envelope.

Both compose the functions the in-process executor already uses, so rehydration and persistence have one
implementation rather than two.

The endpoints are served by a separate crate and binary `agentic-llm-d` that depends on `agentic-server-core` and not on the gateway. It serves `/internal/hydrate`, `/internal/persist`, `/health` and `/ready`.

### Alternatives considered

_No response_

### Additional context

_No response_

Mohammad-nassar10:
## Summary

Exposes the two halves of a stateful Responses turn as separate endpoints, so an external orchestrator (the llm-d coordinator) can make the inference call itself:

- `POST /v1alpha/responses/hydrate`: expands `previous_response_id` into a stateless upstream request, plus a sealed context.
- `POST /v1alpha/responses/persist`: takes that context and the model's response (a JSON body, or the SSE frames a streaming caller relayed), stores the turn, and returns the envelope carrying the stored `resp_` id.

Both compose existing core operations (`rehydrate_conversation`, `upstream_request`, `decode_upstream`, `commit`), so they add no parsing, storage, or request building of their own. `decode_upstream` is the OpenAI JSON/SSE adapter; `commit` owns the shared validation and persistence. The in-process path is unchanged.

The endpoints ship as a new crate and binary, `agentic-llm-d`, depending on `agentic-server-core` alone. The wire protocol (`SplitContext`, the sealing, the splittability check) lives there, so core stays unaware of a specific consumer.

The context is HMAC-signed with an expiry and an audience, and the routes require a shared workload token in `x-agentic-workload-token`, leaving `Authorization` free for end-user identity.

## Test Plan

`cargo test` - new `split_execution_integration.rs` covers the two-turn replay, the error paths, and the context round-trip. Existing suites pass unchanged.
Verified on a cluster with a multi-turn conversation.

Closes #215.


crates/agentic-llm-d/src/lib.rs
```diff
@@ -0,0 +1,29 @@
...
+    pub exec_ctx: Arc<ExecutionContext>,
+}
+
+/// The whole surface: two split-execution endpoints and two probes.
+pub fn build_internal_router(state: InternalState) -> Router {
+    Router::new()
+        .route("/health", get(handler::health))
+        .route("/ready", get(handler::ready))
+        .route("/internal/hydrate", post(handler::internal_hydrate))
+        .route("/internal/persist", post(handler::internal_persist))
```
franciscojavierarceo: these routes have no workload authentication or object-level authorization. once this listener is reachable by the coordinator, any reachable workload with a known response ID can hydrate its full stored history or write a turn. we should authenticate both endpoints and scope every read and write to the authenticated tenant.

crates/agentic-server-core/src/executor/request.rs
```diff
@@ -49,6 +54,54 @@ impl RequestContext {
...
 }
 
+/// Wire form of a [`RequestContext`]: the subset that survives leaving the process.
+///
+/// Absent on purpose: `enriched_request` (the conversation, already in flight as
+/// the request — rebuilt on return) and `new_input_items` (derived).
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct SplitContext {
+    /// Response id reserved for this turn (`resp_` prefix).
+    pub response_id: String,
```
franciscojavierarceo: `SplitContext` is accepted as caller-authored authority: persist never proves it came from hydrate and directly trusts its `response_id`, original request, continuation link, and effective configuration. a caller can skip hydrate and forge durable records; an empty `response_id` also returns success because persistence silently skips the write. we should use a signed, expiring context or a one-time server-side handle, and reject malformed reserved IDs.

crates/agentic-server-core/src/executor/split.rs
```diff
@@ -0,0 +1,107 @@
...
+/// # Errors
+/// [`ExecutorError::InvalidRequest`] for a non-terminal response or a cut-short
+/// stream, a parse error for an invalid body, or a storage error.
+pub async fn persist(
+    context: SplitContext,
+    upstream: UpstreamBody<'_>,
+    exec_ctx: &ExecutionContext,
+) -> ExecutorResult<ResponsePayload> {
+    let ctx = RequestContext::from(context);
+    let payload = payload_from_upstream(&ctx, upstream)?;
```
franciscojavierarceo: at the new HTTP persistence boundary, a response containing only `{"id":"x"}` defaults to `Completed` with empty output, while malformed output entries are silently dropped. this lets an invalid model response persist a completed turn with no valid output. we should deserialize a strict terminal response here and reject missing status, missing output, and malformed output items.

crates/agentic-server-core/src/executor/split.rs
```diff
@@ -0,0 +1,107 @@
...
+        payload.status.parse::<ResponseStatus>().unwrap_or_default(),
+        ResponseStatus::Completed | ResponseStatus::Incomplete
+    ) {
+        return Err(ExecutorError::InvalidRequest(format!(
+            "upstream response status '{}' cannot be persisted",
+            payload.status
+        )));
+    }
+
+    persist_if_needed(
```
franciscojavierarceo: this insert makes `/internal/persist` non-idempotent. if the transaction commits but the response is lost, retrying the exact request hits the same primary key and returns HTTP 500 even though the turn was stored. we should use the reserved response ID as an idempotency key: return the existing result for an identical retry and `409` for mismatched reuse.

crates/agentic-server-core/src/executor/split.rs
```diff
@@ -0,0 +1,107 @@
...
+    context: SplitContext,
+    upstream: UpstreamBody<'_>,
+    exec_ctx: &ExecutionContext,
+) -> ExecutorResult<ResponsePayload> {
+    let ctx = RequestContext::from(context);
+    let payload = payload_from_upstream(&ctx, upstream)?;
+
+    // The in-process flow silently skips non-terminal statuses; here that would
+    // return an envelope whose id can never be continued, so reject it.
+    if !matches!(
```
franciscojavierarceo: `response.failed` is a terminal upstream result, but it becomes `ResponseStatus::Error` and is rejected here as `InvalidRequest`, so `/internal/persist` reports a model failure as HTTP 400. we should return the normalized failed response or a dedicated upstream-error status, and reserve 400 for malformed or non-terminal input.

crates/agentic-llm-d/src/main.rs
```diff
@@ -0,0 +1,50 @@
...
+        db_url: cli.db_url,
+        postgres: PostgresConfig::default(),
+        sqlite: SqliteConfig::default(),
+        tools: ToolRuntimeConfig::default(),
+    };
+
+    let shutdown = CancellationToken::new();
+    let on_signal = shutdown.clone();
+    tokio::spawn(async move {
+        if tokio::signal::ctrl_c().await.is_ok() {
```
franciscojavierarceo: this only handles Ctrl-C, while Kubernetes terminates containers with SIGTERM. the default SIGTERM action bypasses the graceful-shutdown future and can cut off in-flight hydrate or persist requests. we should reuse the gateway's SIGTERM-aware shutdown path and bounded drain timeout.

crates/agentic-llm-d/src/main.rs
```diff
@@ -0,0 +1,50 @@
...
+
+use clap::Parser;
+use tokio_util::sync::CancellationToken;
+
+use agentic_core::config::{Config, PostgresConfig, SqliteConfig, ToolRuntimeConfig};
+use agentic_llm_d::runner;
+
+#[derive(Parser)]
+#[command(name = "agentic-llm-d", about = "agentic-api backend mode for the llm-d coordinator")]
+struct Cli {
```
maralbahari, line 12: Could we reuse `agentic-server`’s existing common CLI/configuration instead of defining another `Cli` here? `agentic-server` already supports `--gateway-host`, `--gateway-port`, and `--db-url`. Duplicating these options with different names and defaults risks configuration drift. Could the llm-d backend be exposed as an `agentic-server` subcommand, or could both binaries share the same common argument structure?

docs/design/agentic-llm-d.md
```diff
@@ -0,0 +1,62 @@
...
+response envelope carrying the reserved `resp_` identifier. Nothing is re-emitted for a streamed turn, since the caller
+has already sent the frames on.
+
+`SplitContext` is the wire form of the in-process `RequestContext`. It omits the enriched request, which is already in
+flight as the request body, and the derived input items; both are rebuilt when the context comes back. Callers treat it
+as opaque.
+
+## Composition
+
+Neither step reimplements the flow. `hydrate` calls `rehydrate_conversation` and `upstream_request_json`, and `persist`
```
maralbahari: can we reconsider the boundary between split execution and upstream handling?

 The llm-d integration is intended to expose only hydration and persistence while the coordinator owns inference. However, implementing it currently requires changing the existing `agentic-server-core` upstream path by introducing `UpstreamBody`, `payload_from_upstream`, `absorb_line`, and `ResponseAccumulator::saw_terminal_frame`. This couples the split-execution API to the current OpenAI Responses JSON/SSE representation and increases the regression surface for the normal `agentic-server` inference path.

Could core instead expose a transport-neutral prepare/commit API built around the existing `rehydrate_conversation` and `persist_response` operations? The commit operation would also handle the missing shared behavior—context reconstruction, terminal-state validation, reserved-ID injection, and conditional persistence.

 Upstream-specific decoding could then live behind an adapter interface, with the existing OpenAI JSON/SSE implementation as one adapter. The llm-d handlers would decode their input through that adapter and pass a normalized `ResponsePayload` to the core commit operation. This would keep persistence independent of upstream wire formats and make support for future upstream protocols additive rather than requiring further changes to `upstream.rs` and the accumulator.

I do not think the accumulator itself should move entirely into `agentic-llm-d`, because the regular server path already depends on it. The concern is that response decoding, upstream transport, and persistence are currently joined in the same abstraction. Separating those responsibilities would give us a cleaner and more extensible integration boundary.

crates/agentic-server-core/src/executor/upstream.rs
```diff
@@ -32,27 +32,80 @@ pub(super) struct StreamPayload {
...
     let body = fetch_response_json(upstream_json, &url, &exec_ctx.client, auth).await?;
 
-    let acc = ResponseAccumulator::from_json(&body, ctx.conversation_id.as_deref())?;
+    payload_from_upstream(ctx, UpstreamBody::Json(&body))
+}
+
+/// A complete upstream response, in whichever form the caller received it.
+#[derive(Debug, Clone, Copy)]
+pub enum UpstreamBody<'a> {
+    Json(&'a str),
```
maralbahari, line 64: I left a design-level comment in `docs/design/agentic-llm-d.md`. This abstraction currently couples split persistence to the OpenAI JSON/SSE wire format; I think response decoding should sit behind an upstream adapter boundary.

docs/design/agentic-llm-d.md
```diff
@@ -0,0 +1,62 @@
...
+
+The endpoints are served by a separate crate and binary that depends on `agentic-server-core` and not on the gateway.
+It serves `/internal/hydrate`, `/internal/persist`, `/health` and `/ready`, and nothing else: the passthrough proxy,
+`/v1`, the WebSocket transport, upstream readiness probing and vLLM subprocess management are all absent, so the
+internal endpoints cannot be exposed on a listener that also serves `/v1`. Readiness reports whether storage answers,
+since the coordinator owns the model fleet.
+
+## Discussion points
+
+The `/internal` endpoints carry no credential and trust their caller, so restricting them is a network-layer concern
```
maralbahari: relying solely on network isolation does not seem sufficient for endpoints with this level of authority. `/internal/hydrate` can return complete persisted conversation history, while `/internal/persist` accepts an unsigned, caller-controlled `SplitContext` and writes to storage. Any compromised pod, SSRF path, or misconfigured network policy could potentially read history, forge context fields, inject turns, replay requests, or exhaust storage.

`agentic-server` supports optional inbound OIDC bearer-token authentication, but this binary has no equivalent application-layer authentication or authorization. Before merging, could we define and enforce the service-to-service trust model—for example, mTLS/workload identity or bearer-token authentication?

 We should also consider making `SplitContext` tamper-evident, such as an HMAC-signed opaque token with an expiry and audience, or storing the context server-side behind a short-lived, one-time identifier. Replay/idempotency behavior for `persist` should also be defined. Network policy can remain defense in depth, but it should not be the only security control.

crates/agentic-server-core/src/executor/request.rs
```diff
@@ -49,6 +54,54 @@ impl RequestContext {
     }
 }
 
+/// Wire form of a [`RequestContext`]: the subset that survives leaving the process.
+///
+/// Absent on purpose: `enriched_request` (the conversation, already in flight as
+/// the request — rebuilt on return) and `new_input_items` (derived).
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct SplitContext {
```
maralbahari: Core should not model the private protocol of one consuming crate.

`SplitContext` appears to be specific to the llm-d HTTP boundary rather than part of the executor’s general domain model. It exists to serialize state between `/internal/hydrate` and `/internal/persist`, and its wire representation is consumed only by the llm-d integration.

 Could we move `SplitContext`, its conversions, and the associated wire-level tests into `agentic-llm-d`? The same likely applies to `Hydration`, `UpstreamBody`, and the split-specific orchestration. `agentic-server-core` should expose only the narrow, transport-neutral operations needed for rehydration, response normalization, ID assignment, validation, and persistence.

 This preserves the intended dependency direction—`agentic-llm-d` depends on core—without making core aware of a specific consumer’s private protocol. The split integration tests currently under `agentic-server-core/tests` should move with that boundary, while tests for the generic core operations can remain in core.

crates/agentic-llm-d/src/lib.rs
Mohammad-nassar10: Agreed. On authentication, the coordinator already has the caller's bearer token, it just doesn't forward it. If it did, these endpoints could use the same `require_oidc` check `/v1` uses. `OidcAuthenticator::authenticate` could move to core and both crates share it. Does that sound right? I'll do it in this PR if so. Or if you'd prefer mTLS on the coordinator instead?

crates/agentic-server-core/src/executor/request.rs
Mohammad-nassar10: Fixed, the context is now signed and expiring, and persist verifies it. The empty `response_id` case is covered too.

crates/agentic-server-core/src/executor/split.rs
Mohammad-nassar10: Fixed, thanks!

crates/agentic-server-core/src/executor/split.rs
Mohammad-nassar10: Fixed.

crates/agentic-server-core/src/executor/split.rs
Mohammad-nassar10: Fixed.

crates/agentic-llm-d/src/main.rs
Mohammad-nassar10: Fixed.

crates/agentic-llm-d/src/main.rs
Mohammad-nassar10: The only option that really overlaps is `--db-url/DATABASE_URL`, and that already matches. Host and port have to differ since the two run as separate processes, and "gateway" would be the wrong word here.
A shared arg struct would make this crate depend on agentic-server, and a subcommand would put backend mode back inside the gateway binary, which is what the separate crate is avoiding.
I did change the default bind address to 0.0.0.0 to match.

docs/design/agentic-llm-d.md
Mohammad-nassar10: Excellent point. Core now has `upstream_request`, `decode_upstream`, and `commit`. `commit` takes a normalized `ResponsePayload` and does the validation, id, and storage. `decode_upstream` is the OpenAI JSON/SSE adapter. It stays in core because it needs the accumulator, that shouldn't move, so `UpstreamBody` stayed with it as its input type.

crates/agentic-server-core/src/executor/request.rs
Mohammad-nassar10: Good point, Done. now nothing in core is named after `llm-d`. `SplitContext`, `Hydration`, `ensure_splittable` moved into `agentic-llm-d`.

docs/design/agentic-llm-d.md
Mohammad-nassar10: `SplitContext` is now an HMAC-signed opaque token with expiry and audience, and replay is defined, a retry returns the stored turn.
The trust model is the one still open: `OidcAuthenticator::authenticate` is framework-free, so it could move to `core` with both crates sharing it, or we authenticate the coordinator with mTLS. Which preferred?
Same point [here](https://github.com/vllm-project/agentic-api/pull/216#discussion_r3874734056).

maralbahari: I dont think the endpoints `/internal/hydrate`, and `/internal/persist` should  not move to the `core` server as these endpoints are for `llm-d` crates specific so it needs to handle its own authorization.

Mohammad-nassar10: Agreed, nothing moves to core. `/internal/*` now requires a bearer token `INTERNAL_API_TOKEN`.

crates/agentic-llm-d/src/lib.rs
Mohammad-nassar10: Related point [here](https://github.com/vllm-project/agentic-api/pull/216#discussion_r3877727784).

crates/agentic-llm-d/src/main.rs
```diff
@@ -0,0 +1,86 @@
...
+    db_url: Option<String>,
+
+    /// Seals the context passed between hydrate and persist. Required: without
+    /// it a caller could forge a context and write turns under any id.
+    #[arg(long, env = "INTERNAL_SIGNING_KEY", hide_env_values = true)]
+    signing_key: String,
+
+    /// Shared secret every `/internal` caller must present as a bearer token.
+    /// Required: these endpoints read and write conversation history.
+    #[arg(long, env = "INTERNAL_API_TOKEN", hide_env_values = true)]
```
franciscojavierarceo: both trust-boundary secrets accept arbitrary strings. i confirmed an empty signing key can produce a valid signed context, and a one-character `INTERNAL_API_TOKEN` authenticates successfully. we should validate independently generated high-entropy values at startup before binding the listener.

crates/agentic-server-core/src/executor/upstream.rs
```diff
@@ -32,27 +34,125 @@ pub(super) struct StreamPayload {
...
+    /// Frames of a streamed response, already relayed by the caller.
+    Sse(&'a str),
+}
+
+fn absorb_line(acc: &mut ResponseAccumulator, ctx: &RequestContext, line: &str) {
+    if let Some(frame) = acc.process_sse_line(line) {
```
franciscojavierarceo, line 70: the JSON path is strict now, but SSE still silently drops malformed data frames and accepts a later terminal frame. i reproduced `data: not-json` followed by `response.completed` being stored as completed with empty output. we should make SSE decoding fallible at this boundary and reject malformed data frames.

crates/agentic-llm-d/src/handler.rs
```diff
@@ -0,0 +1,155 @@
...
+    match hydrate(payload, &state).await {
+        Ok(hydration) => axum::Json(hydration).into_response(),
+        Err(error) => error_response(error),
+    }
+}
+
+/// Rehydrates the turn and builds the request the caller forwards to a model.
+#[allow(clippy::result_large_err)] // `ExecutorError` is core's; boxing it is not ours to decide
+async fn hydrate(request: RequestPayload, state: &InternalState) -> agentic_core::executor::ExecutorResult<Hydration> {
+    ensure_splittable(&request)?;
```
franciscojavierarceo, line 97: `ensure_splittable` runs before rehydration restores effective tools. a continuation that omits `tools` can therefore inherit a gateway-owned tool from stored metadata after it passes this check. we should validate the fully rehydrated effective request before building or sealing the upstream request.

crates/agentic-llm-d/src/context.rs
```diff
@@ -0,0 +1,129 @@
...
+    /// cannot read or alter what it carries.
+    pub context: String,
+}
+
+/// Wire form of a [`RequestContext`]. `enriched_request` and `new_input_items`
+/// are absent on purpose: both are rebuilt on return.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct SplitContext {
+    pub response_id: String,
+    pub original_request: RequestPayload,
```
franciscojavierarceo, line 30: hydrate and persist share the same 10 MiB body limit, but the signed context embeds and base64-expands the full original request. i reproduced an accepted 8 MiB hydrate body whose context made the persist body exceed 10 MiB before adding meaningful output. we should use a compact server-side handle or derive separate round-trip-safe limits.

crates/agentic-llm-d/src/lib.rs
```diff
@@ -0,0 +1,39 @@
...
+}
+
+/// The whole surface: two split-execution endpoints and two probes.
+pub fn build_internal_router(state: InternalState) -> Router {
+    // Probes stay open so an orchestrator can check liveness without the secret.
+    let probes = Router::new()
+        .route("/health", get(handler::health))
+        .route("/ready", get(handler::ready));
+    let internal = Router::new()
+        .route("/internal/hydrate", post(handler::internal_hydrate))
```
franciscojavierarceo: i don't think `/internal` is the right namespace here, it describes deployment topology and can imply a security boundary. since this contract is tightly coupled to the Responses API, we should use `/v1alpha/responses/hydrate` and `/v1alpha/responses/persist`.

Dockerfile
```diff
@@ -51,6 +53,7 @@ RUN apt-get update && \
     chmod g=u,g+s /var/lib/agentic-api
 
 COPY --from=rust-build /out/agentic-server /usr/local/bin/agentic-server
+COPY --from=rust-build /out/agentic-llm-d /usr/local/bin/agentic-llm-d
```
franciscojavierarceo, line 56: the image now contains `agentic-llm-d`, but the entrypoint always executes `agentic-server`, so setting the image command to `agentic-llm-d` still launches the wrong binary. we should let an explicit command select the binary or add a dedicated image or entrypoint for this mode.

crates/agentic-llm-d/src/lib.rs
franciscojavierarceo: the shared token closes unauthenticated workload access, but it still does not address object-level authorization. every token holder can hydrate any known response ID, and the response query is still scoped only by `id`. using `Authorization` for the workload token also leaves no channel for the OIDC token from the caller. we should separate workload authentication from end-user identity, bind the authenticated subject or tenant to stored state, and scope every read and write to it.

crates/agentic-server-core/src/executor/split.rs
franciscojavierarceo: this is still not idempotent for mismatched or concurrent reuse. if the response ID already exists, we return the new caller payload without comparing it to the stored turn, and the separate `stored` check and insert can race. i reproduced storage containing `4` while the retry returned `5`. we should make insert-or-compare atomic, return the canonical stored result for an identical retry, and return `409` for mismatched reuse.

Dockerfile
Mohammad-nassar10: Fixed, the entrypoint now takes the binary from the command, so command: ["agentic-llm-d"] runs it.

crates/agentic-server-core/src/executor/upstream.rs
Mohammad-nassar10: Fixed, thanks.

crates/agentic-llm-d/src/handler.rs
Mohammad-nassar10: Fixed, `ensure_splittable` runs again on the rehydrated request.

crates/agentic-llm-d/src/lib.rs
Mohammad-nassar10: Agreed, using `/v1alpha/responses/hydrate` and `/persist`.
Dropped `/internal` from the rest of the crate too.

crates/agentic-llm-d/src/lib.rs
Mohammad-nassar10: Agreed. The workload token moved to `x-agentic-workload-token`. Object-level authorization needs an owner on stored responses which is what in the issue #107, should we handled it here or separate PR?

crates/agentic-llm-d/src/main.rs
Mohammad-nassar10: Added a check that must be at least 32 characters and must differ, checked before the listener binds. Whether a value is actually random can be part of operator requirements, stated in --help on both flags. Is that enough here, or would want something stronger?

crates/agentic-llm-d/src/context.rs
Mohammad-nassar10: right, anything over ~7.5 MiB hydrates fine and can then never be persisted, since the context alone base64-expands past the limit.
A server-side handle would fix this: persist carries an id instead of the payload. It makes hydrate stateful though so maybe can be handled in a follow-up PR. For example to cap hydrate at a round-trip-safe size here, then the handle as a follow-up. Works, or prefer the handle now?

crates/agentic-server-core/src/executor/split.rs
Mohammad-nassar10: The mismatched case is fixed: a reused id returns 409, not the caller's payload. The race isn't - `stored()` is still a read before the write, so concurrent retries can both pass it.

The canonical stored result is the harder one: we persist the turn's items and metadata, not the envelope (`status`, `usage`, `incomplete_details` and `error` aren't kept) so it has to be stored before it can be returned. And telling an identical retry from a mismatched one needs a request digest alongside it. Both are schema changes, so I'd do those as a follow-up. WDYT?

franciscojavierarceo (approved):
Synced with main. Merging this and addressing the remaining review findings in a focused follow-up PR.

PR #219, opened 2026-08-28
fix: reconcile complete streamed reasoning items

### LINKED ISSUE #152: [Reasoning retention 1/2] Reconcile complete streamed reasoning items

StevenWang-CY:
## Summary

Fixes #152.

The streaming accumulator previously merged reasoning text and summary events into one unindexed buffer and did not reconcile reasoning `response.output_item.done` events. A streamed response could consequently omit completed content, summaries, `encrypted_content`, status, or a reasoning item whose added event was absent.

This change:

- gives reasoning text and reasoning summary events distinct typed payloads with their output and part indexes;
- constructs reasoning items through `ReasoningOutput::try_from` for both added and done-only lifecycle events;
- routes actual reasoning text and summary done events directly through `ReasoningOutput::apply_done`;
- uses the complete reasoning item for every field it contains, including empty arrays and `null` values;
- preserves `content` or `summary` collected from preceding part-done events when the complete item omits that array;
- retains valid part-done state when a complete item is malformed;
- verifies complete typed reasoning items survive the existing JSON storage round trip without leaking the private storage marker.

The Responses event contract defines reasoning done-event text as the full completed text. `ResponseAccumulator` therefore owns event lifecycle routing without retaining another reasoning content model or reconstructing terminal fields from deltas. No database schema or migration changes are required.

## Test Plan

- `cargo test -p agentic-server-core reasoning_output_` — 3 passed
- `cargo test -p agentic-server-core reasoning_done_text_is_authoritative_even_when_empty` — 1 passed
- `cargo test -p agentic-server-core executor::accumulator` — 54 passed
- `cargo test -p agentic-server-core --test event_normalizer_test` — 40 passed
- `cargo test -p agentic-server-core storage` — 69 passed; 3 PostgreSQL environment tests ignored
- `cargo test -p agentic-server-core` — passed
- `cargo test --workspace -- --test-threads=8` — passed
- `cargo clippy --workspace --all-targets -- -D warnings` — passed
- `cargo fmt --all -- --check` — passed
- `uvx pre-commit==4.4.0 run --all-files` — passed

### DIFF
--- a/crates/agentic-server-core/src/executor/accumulator.rs
+++ b/crates/agentic-server-core/src/executor/accumulator.rs
@@ -7,7 +7,6 @@
 //! runs on a blocking thread while the async task continues reading from the
 //! network — keeping the tokio executor thread free between chunk arrivals.
 
-use std::collections::BTreeMap;
 use std::pin::Pin;
 use std::sync::mpsc;
 
@@ -29,83 +28,17 @@ use crate::types::request_response::{IncompleteDetails, ResponsePayload};
 use crate::utils::common::{deserialize_from_str, deserialize_from_value_opt};
 use crate::utils::uuid7_str;
 
-#[derive(Default)]
-struct ReasoningBuffers {
-    content: BTreeMap<u32, String>,
-    summary: BTreeMap<u32, String>,
-}
-
-impl ReasoningBuffers {
-    fn push_content_delta(&mut self, content_index: u32, delta: &str) {
-        self.content.entry(content_index).or_default().push_str(delta);
-    }
-
-    fn take_content(&mut self, content_index: u32) -> String {
-        self.content.remove(&content_index).unwrap_or_default()
-    }
-
-    fn push_summary_delta(&mut self, summary_index: u32, delta: &str) {
-        self.summary.entry(summary_index).or_default().push_str(delta);
-    }
-
-    fn take_summary(&mut self, summary_index: u32) -> String {
-        self.summary.remove(&summary_index).unwrap_or_default()
-    }
-}
-
-fn apply_reasoning_buffers(item: &mut ReasoningOutput, buffers: ReasoningBuffers, output_index: u32) {
-    for (content_index, mut buffer) in buffers.content {
-        let done = EventPayload::ReasoningTextDone {
-            text: String::new(),
-            item_id: item.id.clone(),
-            output_index,
-            content_index,
-        };
-        item.apply_done(&done, &mut buffer);
-    }
-
-    for (summary_index, mut buffer) in buffers.summary {
-        let done = EventPayload::ReasoningSummaryTextDone {
-            text: String::new(),
-            item_id: item.id.clone(),
-            output_index,
-            summary_index,
-        };
-        item.apply_done(&done, &mut buffer);
-    }
-}
-
 /// Tracks a single output item currently being streamed, together with its
 /// accumulated text/arguments buffer.
 enum InFlight {
-    Message {
-        item: OutputMessage,
-        text: String,
-    },
-    Reasoning {
-        item: ReasoningOutput,
-        buffers: ReasoningBuffers,
-    },
-    FunctionCall {
-        item: FunctionToolCall,
-        arguments: String,
-    },
-    CustomToolCall {
-        item: CustomToolCall,
-        input: String,
-    },
-    WebSearchCall {
-        item: Option<WebSearchCall>,
-    },
-    McpCall {
-        item: McpCall,
-    },
-    McpListTools {
-        item: McpListTools,
-    },
-    Compaction {
-        item: CompactionItem,
-    },
+    Message { item: OutputMessage, text: String },
+    Reasoning { item: ReasoningOutput },
+    FunctionCall { item: FunctionToolCall, arguments: String },
+    CustomToolCall { item: CustomToolCall, input: String },
+    WebSearchCall { item: Option<WebSearchCall> },
+    McpCall { item: McpCall },
+    McpListTools { item: McpListTools },
+    Compaction { item: CompactionItem },
 }
 
 impl std::fmt::Debug for InFlight {
@@ -124,12 +57,9 @@ impl std::fmt::Debug for InFlight {
 }
 
 impl InFlight {
-    fn finalize(self, output_index: u32) -> Option<OutputItem> {
+    fn finalize(self) -> Option<OutputItem> {
         match self {
-            Self::Reasoning { mut item, buffers } => {
-                apply_reasoning_buffers(&mut item, buffers, output_index);
-                Some(OutputItem::Reasoning(item))
-            }
+            Self::Reasoning { item } => Some(OutputItem::Reasoning(item)),
             Self::FunctionCall { mut item, arguments } => {
                 if !arguments.is_empty() && item.arguments.is_empty() {
                     item.arguments = arguments;
@@ -324,12 +254,11 @@ impl ResponseAccumulator {
 
     /// Finalizes all streaming items in upstream `output_index` order.
     pub(crate) fn finalize_all(&mut self) {
-        self.completed.extend(self.in_flight.drain(..).filter_map(|(_, entry)| {
-            entry
-                .item
-                .finalize(entry.output_index)
-                .map(|item| (entry.output_index, item))
-        }));
+        self.completed.extend(
+            self.in_flight
+                .drain(..)
+                .filter_map(|(_, entry)| entry.item.finalize().map(|item| (entry.output_index, item))),
+        );
         self.completed.sort_by_key(|(output_index, _)| *output_index);
         self.output
             .extend(self.completed.drain(..).map(|(_, output_item)| output_item));
@@ -423,9 +352,6 @@ impl ResponseAccumulator {
     /// frame (e.g. [`StreamTee`](future)) can call this directly without
     /// re-parsing from a raw line.
     pub(crate) fn process_event(&mut self, frame: &EventFrame) {
-        if self.route_reasoning_event(frame) {
-            return;
-        }
         match (&frame.event_type, &frame.payload) {
             (SSEEventType::ResponseCreated, EventPayload::Response { id, .. }) if !id.is_empty() => {
                 self.response_id.clone_from(id);
@@ -435,6 +361,22 @@ impl ResponseAccumulator {
             }
             (SSEEventType::OutputItemDone, payload @ EventPayload::OutputItemDone { .. }) => {
                 self.complete_output_item(payload);
+            }
+            (
+                SSEEventType::ReasoningTextDone,
+                payload @ EventPayload::ReasoningTextDone {
+                    item_id, output_index, ..
+                },
+            )
+            | (
+                SSEEventType::ReasoningSummaryTextDone,
+                payload @ EventPayload::ReasoningSummaryTextDone {
+                    item_id, output_index, ..
+                },
+            ) => {
+                if let Some(item) = self.in_flight_reasoning_mut(item_id, *output_index) {
+                    item.apply_done(payload, &mut String::new());
+                }
             }
             (
                 SSEEventType::FunctionCallArgumentsDelta,
@@ -503,67 +445,6 @@ impl ResponseAccumulator {
         }
     }
 
-    fn route_reasoning_event(&mut self, frame: &EventFrame) -> bool {
-        match (&frame.event_type, &frame.payload) {
-            (
-                SSEEventType::ReasoningTextDelta,
-                EventPayload::ReasoningTextDelta {
-                    delta,
-                    item_id,
-                    output_index,
-                    content_index,
-                },
-            ) => {
-                if let Some((_, buffers)) = self.in_flight_reasoning_mut(item_id, *output_index) {
-                    buffers.push_content_delta(*content_index, delta);
-                }
-            }
-            (
-                SSEEventType::ReasoningTextDone,
-                payload @ EventPayload::ReasoningTextDone {
-                    item_id,
-                    output_index,
-                    content_index,
-                    ..
-                },
-            ) => {
-                if let Some((item, buffers)) = self.in_flight_reasoning_mut(item_id, *output_index) {
-                    let mut buffer = buffers.take_content(*content_index);
-                    item.apply_done(payload, &mut buffer);
-                }
-            }
-            (
-                SSEEventType::ReasoningSummaryTextDelta,
-                EventPayload::ReasoningSummaryTextDelta {
-                    delta,
-                    item_id,
-                    output_index,
-                    summary_index,
-                },
-            ) => {
-                if let Some((_, buffers)) = self.in_flight_reasoning_mut(item_id, *output_index) {
-                    buffers.push_summary_delta(*summary_index, delta);
-                }
-            }
-            (
-                SSEEventType::ReasoningSummaryTextDone,
-                payload @ EventPayload::ReasoningSummaryTextDone {
-                    item_id,
-                    output_index,
-                    summary_index,
-                    ..
-                },
-            ) => {
-                if let Some((item, buffers)) = self.in_flight_reasoning_mut(item_id, *output_index) {
-                    let mut buffer = buffers.take_summary(*summary_index);
-                    item.apply_done(payload, &mut buffer);
-                }
-            }
-            _ => return false,
-        }
-        true
-    }
-
     fn start_output_item(&mut self, payload: &EventPayload) {
         let EventPayload::OutputItemAdded {
             item_id,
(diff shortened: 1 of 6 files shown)


crates/agentic-server-core/src/executor/accumulator.rs
```diff
@@ -28,17 +29,112 @@ use crate::types::request_response::{IncompleteDetails, ResponsePayload};
 use crate::utils::common::{deserialize_from_str, deserialize_from_value_opt};
 use crate::utils::uuid7_str;
 
+#[derive(Default)]
+struct ReasoningFallback {
```
maralbahari: `ReasoningFallback` creates a second representation of state already owned by `ReasoningOutput`.

 `ReasoningOutput` already contains the typed `content` and `summary` fields, is initialized through `TryFrom<&EventPayload>`, and implements `ApplyDone`. However, the new accumulator path stores reasoning events in `ReasoningFallback` and later performs separate manual reconciliation in `complete_reasoning_item`. Consequently, the updated `ApplyDone for ReasoningOutput` implementation is not used by the production accumulator path.

 The accumulator should only coordinate the event lifecycle:

 - On `OutputItemAdded`, construct the item with `ReasoningOutput::try_from(payload)`.
 - On `ReasoningTextDone`, `ReasoningSummaryTextDone`, or `OutputItemDone`, find the corresponding in-flight item and call `ReasoningOutput::apply_done`.
 - If deltas must be buffered before their done event, the accumulator may hold the required temporary buffer, but it should not maintain another `content`/`summary` model or implement reconciliation semantics.

 Please extend `ReasoningOutput::try_from` and `ReasoningOutput::apply_done` to support all required added and completed event forms. In particular, `ApplyDone` should treat the complete `OutputItemDone` payload as authoritative and update `content`, `summary`, `encrypted_content`, and `status` there.

 If `OutputItemAdded` must initialize more than the ID, the normalized `EventPayload` should retain the necessary item fields so `TryFrom` can consume them. The accumulator should only locate the corresponding item and dispatch the event.

 This keeps `ReasoningOutput` as the single source of truth, follows the same conversion/application pattern used by the other output-item types, and avoids distributing reasoning behavior across `ReasoningFallback`, `process_reasoning_event`, `complete_reasoning_item`, and `ReasoningOutput::apply_done`. The tests should validate the `TryFrom`/`ApplyDone` lifecycle rather than establish a separate fallback mechanism.

StevenWang-CY: Addressed in ea645c0.

The accumulator now owns only per-index raw text buffers and lifecycle routing. ReasoningTextDone, ReasoningSummaryTextDone, and OutputItemDone are dispatched to ReasoningOutput::apply_done. ReasoningOutput::try_from handles both OutputItemAdded and done-only OutputItemDone construction.

Completed-item field presence, nullable arrays, malformed completion handling, content and summary replacement, encrypted_content, and status are reconciled inside ReasoningOutput. OutputItemAdded already carries the only field needed for initialization, the ID, so the normalized payload did not need another field.

I also added direct TryFrom and ApplyDone lifecycle tests, including indexed content and summary ordering, authoritative completed fields, omitted-field fallback, and malformed-completion preservation.

Local validation passed:

- cargo test -p agentic-server-core reasoning_output_
- cargo test -p agentic-server-core executor::accumulator
- cargo test -p agentic-server-core
- cargo test --workspace -- --test-threads=8
- cargo clippy --workspace --all-targets -- -D warnings
- cargo fmt --all -- --check
- uvx pre-commit==4.4.0 run --all-files

The updated head is now running through CI.

crates/agentic-server-core/src/executor/accumulator.rs
```diff
@@ -21,24 +22,90 @@ use crate::types::event::{MessageStatus, ResponseStatus};
...
-    ReasoningOutput, ReasoningTextContent, ResponseUsage,
+    ReasoningOutput, ResponseUsage,
 };
 use crate::types::io::{McpCall, WebSearchCall};
 use crate::types::request_response::{IncompleteDetails, ResponsePayload};
 use crate::utils::common::{deserialize_from_str, deserialize_from_value_opt};
 use crate::utils::uuid7_str;
 
+#[derive(Default)]
+struct ReasoningBuffers {
```
maralbahari: `ReasoningFallback` has effectively been renamed to `ReasoningBuffers`; the fallback behavior still exists. `apply_reasoning_buffers` also creates synthetic `ReasoningTextDone` and `ReasoningSummaryTextDone` events during finalization, leaving reasoning reconstruction duplicated between the accumulator and `ReasoningOutput::apply_done`.

 The accumulator should only route real lifecycle events. `OutputItemAdded` and done-only `OutputItemDone` should use `ReasoningOutput::try_from`, while actual reasoning done events should locate the in-flight item and call `apply_done`. Because the done payload contains the authoritative final text, deltas do not need to be retained for final response construction.

For example, the flow could remain close to the original `main` implementation, adapted for the new event variants:

```rust
enum InFlight {
    // ...
    Reasoning {
        item: ReasoningOutput,
    },
}
```

```rust
SSEItemType::Reasoning => ReasoningOutput::try_from(payload)
    .ok()
    .map(|item| InFlight::Reasoning { item }),
```

```rust
(
    SSEEventType::ReasoningTextDone,
    payload @ EventPayload::ReasoningTextDone {
        item_id,
        output_index,
        ..
    },
) => {
    let key = self.in_flight_reasoning_key(item_id, *output_index);

    if let Some(InFlight::Reasoning { item }) = key
        .as_deref()
        .and_then(|key| self.in_flight.get_mut(key))
        .map(|entry| &mut entry.item)
    {
        item.apply_done(payload, &mut String::new());
    }
}

(
    SSEEventType::ReasoningSummaryTextDone,
    payload @ EventPayload::ReasoningSummaryTextDone {
        item_id,
        output_index,
        ..
    },
) => {
    let key = self.in_flight_reasoning_key(item_id, *output_index);

    if let Some(InFlight::Reasoning { item }) = key
        .as_deref()
        .and_then(|key| self.in_flight.get_mut(key))
        .map(|entry| &mut entry.item)
    {
        item.apply_done(payload, &mut String::new());
    }
}
```

The existing `complete_reasoning_item` can likewise call `item.apply_done(payload, ...)` for an in-flight item or `ReasoningOutput::try_from(payload)` for a done-only item. With that structure, `ReasoningBuffers`, `apply_reasoning_buffers`, and the synthetic done-event construction can be removed. Reasoning delta events can still be relayed to the client without becoming a second source of truth for the accumulated response.

maralbahari (changes requested):

crates/agentic-server-core/src/executor/accumulator.rs
StevenWang-CY: Addressed in 9535d3b.

`ReasoningBuffers` and `apply_reasoning_buffers` are removed. The in-flight reasoning variant now owns only `ReasoningOutput`. Actual `ReasoningTextDone` and `ReasoningSummaryTextDone` events locate that item and call `apply_done` directly; reasoning deltas are not retained as a second response model. `OutputItemDone` calls `apply_done` for an in-flight item or `ReasoningOutput::try_from` for a done-only item.

The accumulator tests now exercise real added, reasoning-done, and output-item-done lifecycles. Direct `TryFrom` and `ApplyDone` tests remain in the typed output module. Full core and workspace tests, workspace clippy with warnings denied, rustfmt, and all pre-commit hooks pass on the updated head.

crates/agentic-server-core/src/executor/accumulator.rs
```diff
@@ -519,7 +515,7 @@ impl ResponseAccumulator {
         self.usage = usage;
     }
 
-    fn complete_call_item(&mut self, payload: &EventPayload) {
+    fn complete_output_item(&mut self, payload: &EventPayload) {
```
maralbahari: why the name has changed?

crates/agentic-server-core/src/executor/accumulator.rs
```diff
@@ -573,6 +573,67 @@ impl ResponseAccumulator {
...
+            item.apply_done(payload, &mut String::new());
+            return;
+        }
+
+        if let Ok(completed) = ReasoningOutput::try_from(payload) {
+            self.upsert_completed_reasoning(*output_index, completed);
+        }
+    }
+
+    fn upsert_completed_reasoning(&mut self, output_index: u32, item: ReasoningOutput) {
```
maralbahari: `upsert_completed_reasoning` should not be necessary. It adds a separate reconciliation path solely for repeated `OutputItemDone` events without a corresponding in-flight item. Under the expected Responses streaming lifecycle, each reasoning item is created once, updated by its done events, and finalized once:

 ```text
 OutputItemAdded
     → ReasoningOutput::try_from

 ReasoningTextDone / ReasoningSummaryTextDone
     → in_flight_reasoning_mut
     → ReasoningOutput::apply_done

 OutputItemDone
     → ReasoningOutput::apply_done

 finalize
     → OutputItem::Reasoning(item)
 ```

 By the time `finalize` runs, the in-flight `ReasoningOutput` already contains the reconciled content, summary, status, and encrypted content. There should therefore be only one item to emit.

 The upsert path handles repeated orphan `OutputItemDone` events, which are outside this lifecycle and are not handled specially for other output-item types. If done-only compatibility is required, a valid `ReasoningOutput::try_from(payload)` can use the existing generic completed-item path. Please remove this helper and the `repeated_done_only_reasoning_is_upserted_by_id` test unless there is a documented upstream that requires duplicate done-event reconciliation.

crates/agentic-server-core/src/executor/accumulator.rs
StevenWang-CY: Reverted in 2f6ba22. The rename was unnecessary; `OutputItemDone` continues through the existing `complete_call_item` helper, with reasoning handled inside that path.

crates/agentic-server-core/src/executor/accumulator.rs
StevenWang-CY: Addressed in 2f6ba22. `upsert_completed_reasoning`, the separate reasoning completion helper, and the duplicate-done test are removed. An in-flight reasoning item now receives `apply_done` inside the existing completion path; a valid done-only item is built with `ReasoningOutput::try_from` and enters the same generic completed-item branch as the other output types. Full core and workspace tests, clippy with warnings denied, rustfmt, and pre-commit pass.

maralbahari (approved):

crates/agentic-server-core/src/executor/accumulator.rs
```diff
@@ -32,7 +32,7 @@ use crate::utils::uuid7_str;
 /// accumulated text/arguments buffer.
 enum InFlight {
     Message { item: OutputMessage, text: String },
-    Reasoning { item: ReasoningOutput, text: String },
+    Reasoning { item: ReasoningOutput },
```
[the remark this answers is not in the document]
StevenWang-CY: Addressed in dab479f. The accepted lifecycle uses the done payload as the source of truth: the OpenAI event schema defines `text` as the full completed reasoning text, and [vLLM emits it from its accumulated state](https://github.com/vllm-project/vllm/blob/06569a8696076eeae9558928b00f035ded8f8b60/vllm/entrypoints/openai/responses/streaming_events.py#L979-L994). `final_text` is removed; `ReasoningOutput::apply_done` now clears any supplied buffer and applies only the done payload. The new regression test covers empty authoritative done text for both reasoning content and summary. The accumulator remains limited to lifecycle routing.

crates/agentic-server-core/src/types/io/output.rs
```diff
@@ -587,21 +612,62 @@ pub trait ApplyDone {
...
+                };
+
+                if !raw_item.contains_key("content") {
+                    completed.content = std::mem::take(&mut self.content);
+                }
+                if !raw_item.contains_key("summary") {
+                    completed.summary = std::mem::take(&mut self.summary);
+                }
+                *self = completed;
+            }
```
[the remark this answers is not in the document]
StevenWang-CY: The PR description was broader than the lifecycle and is now corrected. Earlier reasoning part-done events can establish only `content` and `summary`, so those arrays are preserved when the complete item omits them. `encrypted_content` and `status` come from the single authoritative `output_item.done`; there is no earlier valid done event from which to preserve either field. Carrying them forward would affect only duplicate terminal events, which are outside the accepted lifecycle, and could retain stale terminal state. The updated description now names only `content` and `summary` for omitted-field preservation.

crates/agentic-server-core/src/types/io/output.rs
```diff
@@ -587,21 +612,62 @@ pub trait ApplyDone {
...
+            _ => {}
         }
     }
 }
 
+fn insert_at_part_index<T>(parts: &mut Vec<T>, part_index: u32, part: T) {
+    let index = usize::try_from(part_index).unwrap_or(usize::MAX).min(parts.len());
+    parts.insert(index, part);
+}
```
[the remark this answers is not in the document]
StevenWang-CY: I checked this against the wire model and the recorded corpus and am keeping the typed `Vec` representation. `content_index` and `summary_index` address parts in contiguous arrays, and valid SSE lifecycles deliver those semantic events in sequence. All 34 recorded reasoning done events are contiguous and ordered. A sparse sequence such as 5 followed by 3 omits array elements 0–2 and 4, so neither a `Vec<Option<T>>` nor placeholders can produce a valid reasoning item without fabricating wire content. The complete `output_item.done` item remains authoritative; clamping is only best-effort recovery for a malformed sparse stream, now documented beside the helper. Adding a second indexed part store for that invalid lifecycle would conflict with the single-source-of-truth structure requested in the maintainer review.

PR #225, opened 2026-08-31
fix: forward Responses reasoning configuration

StevenWang-CY:
## Summary

Stateful Responses requests deserialize through `RequestPayload`. Before this change, that type did not model `reasoning`, so the executor silently dropped the complete reasoning configuration before constructing the upstream request. This affected stored requests, continuations, gateway-tool requests, context-managed requests, and Responses WebSocket mode; the `store: false` transparent proxy path already forwarded the original body unchanged.

This PR adds a typed `ReasoningConfig` for the current Responses fields (`context`, `effort`, `generate_summary`, `mode`, and `summary`) and borrows it into `UpstreamRequest`. The optional nested object is boxed so requests without reasoning do not inflate executor futures. The same per-request configuration now survives rehydration, automatic compaction of the input, streaming, and repeated gateway-tool rounds.

The modeled fields match the current [OpenAI Responses API](https://developers.openai.com/api/reference/cli/resources/responses/methods/create) and the `Reasoning` object accepted by current [vLLM `ResponsesRequest`](https://github.com/vllm-project/vllm/blob/82936c409d17321d5a791796296790508605aaed/vllm/entrypoints/openai/responses/protocol.py#L137-L178).

## Recorded behavior

The cassette recorder now accepts a validated `--reasoning` JSON object and applies it to ordinary and branched Responses requests. `record_reasoning_cassettes.sh` records matching streaming and non-streaming pairs from OpenAI and through the gateway, with an optional direct-vLLM mode. The script stages and validates the complete selected suite before replacing any checked-in fixture, so a later provider failure cannot leave a partially refreshed comparison set.

The four checked-in fixtures were generated by `record_cassette.py` from live `gpt-5.6` requests. The direct and gateway-facing request bodies are exactly equal, including unmodeled keys: same deterministic prompt, `max_output_tokens: 2048`, and `reasoning: {effort: high, summary: detailed}`. Every terminal response contains substantive reasoning and the exact final answer `VALID`. Streaming coverage checks the documented reasoning-summary and message lifecycles, contiguous sequence numbers, stable item IDs and output indexes, authoritative done text, and terminal-response reconciliation before replaying both recordings through the gateway.

The four recorder-generated YAML fixtures account for 2,737 of the PR's 3,557 added lines; they contain 260 captured SSE events. The functional production change remains confined to typed request parsing and upstream serialization.

Recording the gateway pair against the same OpenAI model exposed one pre-existing serialization inconsistency: an absent vLLM-only `cache_salt` was emitted upstream as `null`, which OpenAI rejects as an unknown parameter. The follow-up omits only `None`, matching every neighboring optional upstream field, while preserving non-empty cache salts verbatim; a regression covers both cases.

## Compatibility

- Missing or `null` reasoning remains absent upstream; an empty object remains `{}`.
- Current field values are forwarded verbatim so model-specific validation remains with the upstream service.
- Malformed non-object configurations and non-string known fields fail during typed request parsing.
- Reasoning stays a per-request generation setting. Persistence metadata, response payload shape, and internal compaction-summary policy are unchanged; the client configuration is retained for the final response call.
- A present `cache_salt` is unchanged; only an absent value is omitted instead of serialized as JSON `null`.

## Test plan

Negative controls:

- Unmodified `main` drops the five-field reasoning object from the upstream request.
- The former PR head emits `cache_salt: null`; the live OpenAI-backed gateway request fails with `Unknown parameter: 'cache_salt'`.
- An injected unmodeled request mismatch and an injected non-contiguous stream sequence each make the new integration test fail at the intended assertion.
- A forced failure on the third recording leaves all four pre-existing fixtures byte-identical and removes the staging directory; a successful four-recording stub run promotes exactly the selected suite.
- Malformed reasoning JSON, non-object JSON, unsupported recorder modes, an invalid record set, a missing OpenAI key, and an unreachable gateway fail closed.

Exact head `4270f60d5cd386ddae77bf9fbd57b686949d695b`, based on current `main` `b1dbce5e1f72c8d3058b437f371c0bbdf89fc4b5`:

- live direct-OpenAI and gateway recording: streaming and non-streaming pairs passed validation;
- four-fixture structural audit: complete request-pair equality, HTTP 200, 260 parseable SSE events, contiguous sequence numbers, completed reasoning and message items, exact `VALID` output, and zero supplied-credential matches;
- focused reasoning cassette replay: 2 passed;
- `cargo test --workspace --all-features`: 812 passed, 9 repository-declared PostgreSQL/core/doctest cases ignored, repeated at the exact SHA in a detached clean worktree;
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`: passed;
- Python suite: 111 passed, 3 installation-only cases skipped;
- all configured pre-commit hooks: passed in the detached exact-SHA worktree;
- `cargo fmt --all -- --check` and `git diff --check`: passed;
- range-diff confirms the two previously reviewed commits are unchanged by the current-main rebase; the third commit contains only the cassette guardrail hardening.
- all 13 fresh GitHub checks on the exact head passed, including DCO, Rust, PostgreSQL, Python 3.10-3.13, pre-commit, container smoke, Kubernetes, Dynamo, and both harness integrations.

### DIFF
--- a/crates/agentic-server-core/benches/executor_throughput.rs
+++ b/crates/agentic-server-core/benches/executor_throughput.rs
@@ -140,6 +140,7 @@ fn make_request(input: &str, stream: bool, prev_id: Option<String>) -> RequestPa
         stream,
         store: true,
         include: None,
+        reasoning: None,
         temperature: None,
         top_p: None,
         max_output_tokens: None,
--- a/crates/agentic-server-core/src/executor/compaction.rs
+++ b/crates/agentic-server-core/src/executor/compaction.rs
@@ -146,6 +146,7 @@ fn request_payload(model: String, input: ResponsesInput, instructions: Option<St
         stream: false,
         store: false,
         include: None,
+        reasoning: None,
         temperature: None,
         top_p: None,
         max_output_tokens: None,
--- a/crates/agentic-server-core/src/executor/modes/conversation.rs
+++ b/crates/agentic-server-core/src/executor/modes/conversation.rs
@@ -160,6 +160,7 @@ mod tests {
             stream: false,
             store: true,
             include: None,
+            reasoning: None,
             temperature: None,
             top_p: None,
             max_output_tokens: None,
--- a/crates/agentic-server-core/src/executor/modes/response.rs
+++ b/crates/agentic-server-core/src/executor/modes/response.rs
@@ -116,6 +116,7 @@ mod tests {
             stream: false,
             store: true,
             include: None,
+            reasoning: None,
             temperature: None,
             top_p: None,
             max_output_tokens: None,
--- a/crates/agentic-server-core/src/executor/rehydrate.rs
+++ b/crates/agentic-server-core/src/executor/rehydrate.rs
@@ -336,6 +336,7 @@ mod tests {
             stream: false,
             store: true,
             include: None,
+            reasoning: None,
             temperature: None,
             top_p: None,
             max_output_tokens: None,
--- a/crates/agentic-server-core/src/executor/upstream.rs
+++ b/crates/agentic-server-core/src/executor/upstream.rs
@@ -314,6 +314,7 @@ mod tests {
             stream: true,
             store: false,
             include: None,
+            reasoning: None,
             temperature: None,
             top_p: None,
             max_output_tokens: None,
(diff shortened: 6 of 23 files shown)


maralbahari (dismissed):
@StevenWang-CY please record openai cassettes with same reasnoning configuration and another cassettes with the same reasoning configuration this time on gateway use [agentic-server-core/tests/cassettes/record_cassette.py](https://github.com/vllm-project/agentic-api/blob/main/crates/agentic-server-core/tests/cassettes/record_cassette.py) for recording cassettes and checkout the [README.md](https://github.com/vllm-project/agentic-api/blob/main/crates/agentic-server-core/tests/cassettes/README.md) how to use the recorder. similar to how web_search, mcp, custom tool type cassettes are recorded for openai which is the ground truth and gateway which needs to match the behavior for both streaming and non streaming request types meaning would need to add a reasnoning_test integration test.

franciscojavierarceo (approved):

franciscojavierarceo (approved):

Issue #218, opened 2026-08-28
Feature request: Generate OpenAPI documentation for the agentic API server


franciscojavierarceo:
## Problem statement / motivation

The agentic API currently exposes OpenAI-compatible Responses endpoints and Anthropic-compatible Messages endpoints, but there is no machine-readable OpenAPI document for client authors, tooling, or deployment integrations.

A published API contract would make the project easier for the community to discover, integrate, validate, and generate clients against.

## Proposed solution

Add compile-time OpenAPI documentation generated from the Rust server, ideally exposed at a stable endpoint such as `/openapi.json`, with optional Swagger UI or another documentation frontend.

The initial contract should cover the stable HTTP surface, including:

- `/health` and `/ready`
- `/v1/models`
- `/v1/conversations`
- `/v1/responses`
- `/v1/responses/compact`
- `/v1/messages`
- `/v1/messages/count_tokens`
- authentication and common error envelopes
- JSON and SSE response shapes where applicable

The WebSocket form of `/v1/responses` can be documented separately if OpenAPI is not an appropriate representation.

A Rust-native solution such as `utoipa`/`utoipa-axum` could be evaluated, while preserving the existing proxy behavior and explicitly documenting fields that are intentionally extensible.

## Alternatives considered

- Continue maintaining prose-only API documentation.
- Hand-maintain a standalone OpenAPI file, which would be more likely to drift from the Rust implementation.
- Use a different Rust OpenAPI library such as `aide`.

## Additional context

This is intended as a community-facing feature request and does not require an immediate implementation commitment. The generated document should describe the gateway's public contract, not expose internal executor or storage details.

avinashsingh77:
@franciscojavierarceo I am interested in picking this one.

avinashsingh77:
This is implemented in https://github.com/vllm-project/agentic-api/pull/229

PR #229, opened 2026-09-01
feat: add compile-time OpenAPI spec generation via utoipa

avinashsingh77:
## Summary

Adds a feature-gated (`openapi`, default-on) OpenAPI 3.1 spec served at GET /openapi.json, covering all 8 HTTP endpoints with request/response schemas, Bearer auth, and error envelopes. 

~40 types auto-derive ToSchema; ~15 complex types (untagged enums, custom serde) get manual PartialSchema + ToSchema impls. 

Zero-cost when the feature is disabled.

## Test Plan
- [x] `cargo check --features openapi` — compiles with OpenAPI enabled
- [x] `cargo check --no-default-features` — compiles with OpenAPI disabled (zero-cost gate)
- [x] `cargo clippy --features openapi --all-targets` — zero warnings
- [x] `cargo test --features openapi` — all tests pass
- [x] `cargo test --no-default-features` — all tests pass without the feature
- [x] `spec_contains_expected_paths` — all 8 endpoint paths present
- [x] `spec_contains_core_schemas` — core schemas registered
- [x] `spec_has_security_scheme` — `bearer_auth` scheme present
- [x] `spec_serializes_to_valid_json` — valid OpenAPI 3.x JSON
- [x] `spec_schema_refs_resolve` — all `$ref` pointers resolve
- [x] `spec_validates_against_openapi_3_1_meta_schema` — formal validation against OpenAPI 3.1 meta-schema
- [x] Integration test: `GET /openapi.json` returns valid spec with all paths/schemas
- [x] Swagger UI: `GET /swagger-ui/` returns HTM

Related https://github.com/vllm-project/agentic-api/issues/218


crates/agentic-server/src/handler/http/responses.rs
```diff
@@ -46,6 +46,19 @@ fn has_gateway_tools(payload: &RequestPayload) -> bool {
         .is_some_and(|tools| tools.iter().any(|tool| !matches!(tool, ResponsesTool::Function(_))))
 }
 
+#[cfg_attr(feature = "openapi", utoipa::path(
+    post,
+    path = "/v1/responses",
+    request_body = agentic_core::types::request_response::RequestPayload,
+    responses(
+        (status = 200, description = "Non-streaming response", body = agentic_core::types::request_response::ResponsePayload),
+        (status = 200, description = "Streaming SSE response", content_type = "text/event-stream"),
```
maralbahari: OpenAPI responses are keyed by status code, so declaring 200 twice causes one response to overwrite the other. The generated specification currently keeps only the SSE response and loses the non-streaming JSON schema. Could we describe both media types under a single 200 response?

crates/agentic-server/src/handler/http/messages.rs
```diff
@@ -123,6 +123,19 @@ async fn execute_messages(
     }
 }
 
+#[cfg_attr(feature = "openapi", utoipa::path(
+    post,
+    path = "/v1/messages",
+    request_body = agentic_core::types::messages::MessagesRequest,
+    responses(
+        (status = 200, description = "Non-streaming message response"),
+        (status = 200, description = "Streaming SSE response", content_type = "text/event-stream"),
```
maralbahari: This has the same duplicate-200 issue as /v1/responses: the generated specification retains only the SSE response. Could we combine the JSON and SSE media types into one 200 response?

crates/agentic-server/src/handler/http/messages.rs
```diff
@@ -123,6 +123,19 @@ async fn execute_messages(
...
+        (status = 400, description = "Invalid request", body = crate::openapi::ApiErrorResponse),
```
maralbahari: ApiErrorResponse describes the OpenAI error envelope, but this endpoint returns the Anthropic shape: {"type":"error","error":{"type","message"}}. Could we define and use an AnthropicErrorResponse for the Messages endpoints?

crates/agentic-server-core/src/types/io/input.rs
```diff
@@ -94,13 +99,105 @@ pub enum ToolCallOutput {
...
+        }
+    }
+
+    impl utoipa::PartialSchema for InputContent {
+        fn schema() -> RefOr<Schema> {
+            OneOfBuilder::new()
+                .discriminator(Some(utoipa::openapi::schema::Discriminator::new("type")))
+                .item(Ref::from_schema_name("InputTextContent"))
+                .item(Ref::from_schema_name("InputImageContent"))
+                .item(Ref::from_schema_name("InputFileContent"))
```
maralbahari: This schema does not match InputContent: it includes InputFileContent, which is not a modeled variant and is discarded as Unknown, while omitting output_text and reasoning_text. Also, the referenced schemas do not contain or require the type discriminator. Could we model the actual tagged variants explicitly?

crates/agentic-server-core/src/types/io/input.rs
```diff
@@ -94,13 +99,105 @@ pub enum ToolCallOutput {
...
+    impl utoipa::ToSchema for InputContent {
+        fn name() -> std::borrow::Cow<'static, str> {
+            std::borrow::Cow::Borrowed("InputContent")
+        }
+    }
+
+    impl utoipa::PartialSchema for InputItem {
+        fn schema() -> RefOr<Schema> {
+            OneOfBuilder::new()
+                .discriminator(Some(utoipa::openapi::schema::Discriminator::new("type")))
```
maralbahari, line 247: The referenced schemas do not contain the type property added by Serde’s internally tagged InputItem enum. Consequently, the generated schema declares a discriminator that none of its branches requires. Could each branch include its actual type value? compaction_trigger should also be represented.

crates/agentic-server-core/src/types/tools/params.rs
```diff
@@ -252,6 +282,44 @@ pub enum CodexNamespaceMember {
     Unknown,
 }
 
+#[cfg(feature = "openapi")]
+impl utoipa::PartialSchema for ResponsesTool {
+    fn schema() -> utoipa::openapi::RefOr<utoipa::openapi::schema::Schema> {
+        utoipa::openapi::schema::OneOfBuilder::new()
+            .discriminator(Some(utoipa::openapi::schema::Discriminator::new("type")))
```
maralbahari, line 346: These referenced parameter schemas do not contain or require the type discriminator that Serde adds for ResponsesTool. The resulting schema accepts objects without type and does not associate branches with values such as function, mcp, or custom. Could we model each tagged variant with its concrete discriminator value?

crates/agentic-server/src/handler/http/conversations.rs
```diff
@@ -7,6 +7,16 @@ use agentic_core::executor::{ExecutorError, create_conversation};
 use super::super::common::{executor_error_response, extract_store, read_bytes};
 use crate::app::AppState;
 
+#[cfg_attr(feature = "openapi", utoipa::path(
+    post,
+    path = "/v1/conversations",
+    responses(
+        (status = 200, description = "Conversation created"),
```
maralbahari: This operation does not document either its request body or the returned conversation object. Since the specification is intended for client generation, could we add schemas for both?

crates/agentic-server/src/openapi.rs
```diff
@@ -0,0 +1,278 @@
+#![cfg(feature = "openapi")]
+
+use utoipa::OpenApi;
+
+#[derive(OpenApi)]
+#[openapi(
+    info(
+        title = "vLLM Agentic API",
+        description = "Stateful agentic layer for vLLM — OpenAI-compatible Responses API and Anthropic-compatible Messages API.",
+        version = "0.5.0",
```
maralbahari: Could we avoid hard-coding 0.5.0 here? utoipa can obtain the version from CARGO_PKG_VERSION, preventing the specification from becoming stale during the next release.

crates/agentic-server/src/app.rs
```diff
@@ -245,6 +245,8 @@ pub fn build_router_with_auth(
     authenticator: Option<OidcAuthenticator>,
 ) -> Router {
     let public_routes = Router::new().route("/health", get(health)).route("/ready", get(ready));
+    #[cfg(feature = "openapi")]
+    let public_routes = public_routes.merge(crate::openapi::swagger_ui_router());
```
maralbahari: This adds the OpenAPI document and Swagger UI to the public router, and the feature is enabled by default. Is unauthenticated documentation exposure intended for every default deployment, or should it be opt-in through a separate feature or runtime configuration?

crates/agentic-server/tests/openapi_test.rs
```diff
@@ -0,0 +1,71 @@
...
+
+    let body: serde_json::Value = resp.json().await.unwrap();
+
+    assert!(
+        body["openapi"].as_str().unwrap().starts_with("3."),
+        "must be OpenAPI 3.x"
+    );
+    assert_eq!(body["info"]["title"], "vLLM Agentic API");
+
+    let paths = body["paths"].as_object().expect("paths must be an object");
```
maralbahari, line 41: These tests verify that paths and schemas exist, but not that they match the wire contract. Could we assert both JSON and SSE media types under 200, the Anthropic error shape, and representative tagged-enum discriminator values? This would catch the current schema mismatches.

crates/agentic-server/src/app.rs
avinashsingh77: Exposing API docs unauthenticated by default isn't ideal for production. Added a runtime env var to ServerConfig so operators can opt in, defaulting to disabled

crates/agentic-server-core/src/types/io/tools.rs
```diff
@@ -33,19 +34,48 @@ pub enum ToolChoice {
...
+                    .schema_type(SchemaType::new(Type::String))
+                    .enum_values(Some(["auto", "none", "required"])),
+            )
+            .item(
+                ObjectBuilder::new()
+                    .property("type", ObjectBuilder::new().schema_type(SchemaType::new(Type::String)))
+                    .required("type")
+                    .property("name", ObjectBuilder::new().schema_type(SchemaType::new(Type::String)))
+                    .required("name"),
+            )
```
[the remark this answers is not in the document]
avinashsingh77: Fixed. The ToolChoice schema now correctly describes all four wire shapes:

  - String: "auto" / "none" / "required"
  - Function: {type: "function", name, namespace?}
  - Custom: {type: "custom", name}
  - AllowedTools: {type: "allowed_tools", mode, tools} — with proper refs to AllowedToolsMode and AllowedTool schema

crates/agentic-server-core/src/types/io/input.rs
```diff
@@ -94,13 +99,105 @@ pub enum ToolCallOutput {
...
+    }
+
+    impl utoipa::PartialSchema for InputContent {
+        fn schema() -> RefOr<Schema> {
+            OneOfBuilder::new()
+                .discriminator(Some(utoipa::openapi::schema::Discriminator::new("type")))
+                .item(Ref::from_schema_name("InputTextContent"))
+                .item(Ref::from_schema_name("InputImageContent"))
+                .item(Ref::from_schema_name("InputFileContent"))
+                .into()
```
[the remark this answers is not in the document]
avinashsingh77: Fixed already

crates/agentic-server-core/src/types/io/output.rs
```diff
@@ -806,6 +869,29 @@ pub enum OutputItem {
...
+            .item(utoipa::openapi::Ref::from_schema_name("OutputMessage"))
+            .item(utoipa::openapi::Ref::from_schema_name("FunctionToolCall"))
+            .item(utoipa::openapi::Ref::from_schema_name("CustomToolCall"))
+            .item(utoipa::openapi::Ref::from_schema_name("WebSearchCall"))
+            .item(utoipa::openapi::Ref::from_schema_name("McpCall"))
+            .item(utoipa::openapi::Ref::from_schema_name("McpListTools"))
+            .item(utoipa::openapi::Ref::from_schema_name("ReasoningOutput"))
+            .item(utoipa::openapi::Ref::from_schema_name("CompactionItem"))
+            .into()
+    }
```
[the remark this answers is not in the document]
avinashsingh77: OutputItem now uses the allOf pattern — each variant combines a {"type": "<tag>"} const object with a $ref to the payload schema, correctly reflecting the wire format from #[serde(tag = "type")].

Cargo.toml
```diff
@@ -43,3 +43,6 @@ tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 toml = "0.8"
 tokio-tungstenite = "0.29"
 url = "2"
+utoipa = { version = "5", features = ["chrono", "uuid"] }
+utoipa-axum = "0.2"
+utoipa-swagger-ui = { version = "9", features = ["axum"] }
```
maralbahari: every default build now downloads Swagger UI from GitHub.

`utoipa-swagger-ui` 9 without `vendored` runs a build script that shells out to `curl`. Build output from this branch:

```
SWAGGER_UI_DOWNLOAD_URL: https://github.com/swagger-api/swagger-ui/archive/refs/tags/v5.17.14.zip
reqwest feature: Err(NotPresent)
trying to download using `curl` system package
```

`openapi` is in `default` for `agentic-server`, so this hits `cargo build`, `cargo test`, and the maturin wheel build, and it panics (`.expect("failed to download Swagger UI")`) with no network or no `curl`. Breaks air-gapped builds, `cargo vendor`, and sandboxed CI.

Fix: `utoipa-swagger-ui = { version = "9", features = ["axum", "vendored"] }`

Also worth dropping `default = ["openapi"]` (`crates/agentic-server/Cargo.toml:10`): `ENABLE_OPENAPI_DOCS` already defaults to false, so it costs every build and buys nothing at runtime.

crates/agentic-server-core/src/types/io/input.rs
```diff
@@ -94,13 +99,166 @@ pub enum ToolCallOutput {
...
+        fn schema() -> RefOr<Schema> {
+            OneOfBuilder::new()
+                .discriminator(Some(utoipa::openapi::schema::Discriminator::new("type")))
+                .item(tagged_ref("message", "InputMessage"))
+                .item(tagged_ref("function_call", "InputFunctionToolCall"))
+                .item(tagged_ref("function_call_output", "FunctionToolResultMessage"))
+                .item(tagged_ref("custom_tool_call", "CustomToolCall"))
+                .item(tagged_ref("custom_tool_call_output", "CustomToolCallOutputMessage"))
+                .item(tagged_ref("reasoning", "ReasoningOutput"))
+                .item(tagged_ref("compaction", "CompactionItem"))
```
maralbahari, line 257: `InputItem` is missing `mcp_list_tools`, which breaks the `store=false` round-trip.

#214 added this variant after the PR was opened, so it is easy to miss. It never reaches the model, but it is still part of the wire contract: `OutputItem::McpListTools` appears in `response.output`, and `ToolRegistry::cache_listed_mcp_tools` (`tool/registry.rs:298`) reads it back off the request input to suppress re-emitting the discovery lifecycle. A `store=false` client replaying prior output gets duplicate `response.mcp_list_tools.*` events if it drops the item, or a spec violation if it keeps it.

Add, matching the enum ordering between `reasoning` and `compaction`:

```rust
.item(tagged_ref("mcp_list_tools", "McpListTools"))
```

crates/agentic-server-core/src/types/io/input.rs
```diff
@@ -94,13 +99,166 @@ pub enum ToolCallOutput {
...
+#[cfg_attr(feature = "openapi", derive(utoipa::ToSchema))]
 #[serde(tag = "type", rename_all = "snake_case")]
 pub enum ToolOutputContent {
     InputText(InputTextContent),
     InputImage(InputImageContent),
     InputFile(InputFileContent),
 }
 
+#[cfg(feature = "openapi")]
+mod openapi_schemas {
```
maralbahari, line 112: These hand-written schemas have no mechanism keeping them in sync with the serde attributes.

Roughly 15 manual `PartialSchema` impls transcribe `#[serde]` by hand. Two are already out of step with the types they describe, both in the same direction (schema narrower than what the deserializer accepts, so the spec rejects valid requests): `InputItem` at line 241 below, and `ResponsesTool` in `types/tools/params.rs:311`.

Neither was caught by anything. Both compiled, and all eight new tests passed: `spec_schema_refs_resolve` only catches dangling `$ref`s, and the meta-schema test only validates document structure, not whether a schema matches its Rust type.

`jsonschema` is already a dev-dependency for the meta-schema test, so a real guard is cheap: serialize one fixture per enum variant, including each serde alias, and validate it against that type's component schema with `components/schemas` as the resolution scope. That turns both misses below into test failures and stops the next one landing on rebase.

maralbahari: These hand-written schemas have no mechanism keeping them in sync with the serde attributes.

Roughly 15 manual `PartialSchema` impls transcribe `#[serde]` by hand. Two are already out of step with the types they describe, both in the same direction (schema narrower than what the deserializer accepts, so the spec rejects valid requests): `InputItem` at line 241 below, and `ResponsesTool` in `types/tools/params.rs:311`.

Neither was caught by anything. Both compiled, and all eight new tests passed: `spec_schema_refs_resolve` only catches dangling `$ref`s, and the meta-schema test only validates document structure, not whether a schema matches its Rust type.

`jsonschema` is already a dev-dependency for the meta-schema test, so a real guard is cheap: serialize one fixture per enum variant, including each serde alias, and validate it against that type's component schema with `components/schemas` as the resolution scope. That turns both misses below into test failures and stops the next one landing on rebase.

crates/agentic-server-core/src/types/tools/params.rs
```diff
@@ -252,6 +282,79 @@ pub enum CodexNamespaceMember {
...
+                )
+                .item(Ref::from_schema_name(schema))
+                .into()
+        }
+
+        utoipa::openapi::schema::OneOfBuilder::new()
+            .discriminator(Some(utoipa::openapi::schema::Discriminator::new("type")))
+            .item(tagged("function", "FunctionToolParam"))
+            .item(tagged("mcp", "McpToolParam"))
+            .item(tagged("web_search_preview", "WebSearchToolParam"))
```
maralbahari: The `ResponsesTool::WebSearch` variant accepts three aliases alongside its canonical name (`web_search`, `web_search_preview_2025_03_11`, `web_search_2025_08_26`), but the generated spec lists only `web_search_preview`. Codex CLI sends `type: "web_search"`, so a spec-validating client is told its request is invalid while the gateway accepts it. The aliases exist because real clients send them, so they belong in this variant's `type` enum.

maralbahari (commented):
Thanks for the work on this. 
Left four more inline comments. Two bits of context on where they came from:

- I ran a local `cargo build` on the branch and hit a build-time network fetch pulled in by the Swagger UI dependency. That is the one I would want sorted before merge.
- #214 landed after you opened this, and the rebase left a couple of the hand-written schemas slightly behind the types they describe.

The rest are small.

crates/agentic-server/Cargo.toml
```diff
@@ -6,7 +6,14 @@ edition.workspace = true
 license.workspace = true
 repository.workspace = true
 
+[features]
+default = []
```
franciscojavierarceo: the server's default feature set is empty, so the normal Docker build and Rust CI compile this feature out; `openapi_test.rs` is also entirely gated on `openapi`. that leaves the new implementation absent from the standard image and uncompiled by required checks. we should enable `openapi` explicitly in the release build and CI feature matrix, or make it a default feature, and update the PR description to match.

crates/agentic-server/src/handler/http/messages.rs
```diff
@@ -148,6 +164,17 @@ pub async fn messages(State(state): State<AppState>, request: Request) -> Respon
     proxy_messages(&state, parts, bytes, "/v1/messages").await
 }
 
+#[cfg_attr(feature = "openapi", utoipa::path(
+    post,
+    path = "/v1/messages/count_tokens",
+    responses(
+        (status = 200, description = "Token count result"),
```
franciscojavierarceo: `/v1/messages/count_tokens` consumes and proxies a request body, but this operation declares no `request_body` and no schema for the 200 JSON body. generated clients therefore cannot model a valid call or decode its result. we should add typed request and response schemas, or at minimum explicit free-form JSON schemas for this transparent proxy surface; the same applies to the unit JSON response on `/v1/messages` and the untyped `/v1/models` response.

crates/agentic-server-core/src/types/request_response.rs
```diff
@@ -70,6 +72,106 @@ pub enum ResponseTextFormat {
...
+        let str_type = || ObjectBuilder::new().schema_type(SchemaType::new(Type::String));
+        let bool_type = || ObjectBuilder::new().schema_type(SchemaType::new(Type::Boolean));
+        let num_type = || ObjectBuilder::new().schema_type(SchemaType::new(Type::Number));
+        let int_type = || ObjectBuilder::new().schema_type(SchemaType::new(Type::Integer));
+        let schema: RefOr<_> = ObjectBuilder::new()
+            .property("model", str_type())
+            .required("model")
+            .property("input", Ref::from_schema_name("ResponsesInput"))
+            .required("input")
+            .property("instructions", str_type())
```
franciscojavierarceo: `RequestPayload`'s `Option` fields are documented as non-null strings, numbers, and arrays. serde accepts explicit `null`, and several fields also serialize `None` as `null`, so the published schema rejects valid wire values. we should make every optional field's schema nullable and add a serde-output-to-schema fixture.

crates/agentic-server-core/src/types/io/output.rs
```diff
@@ -380,6 +428,31 @@ pub enum McpCallError {
     Unknown(Value),
 }
 
+#[cfg(feature = "openapi")]
+impl utoipa::PartialSchema for McpCallError {
+    fn schema() -> utoipa::openapi::RefOr<utoipa::openapi::schema::Schema> {
+        utoipa::openapi::schema::OneOfBuilder::new()
```
franciscojavierarceo: `McpCallError` uses `oneOf` for both `McpToolExecutionError` and a generic object. a structured tool-execution error matches both branches, so it fails `oneOf`; `Unknown(Value)` also accepts arrays, scalars, and null that match neither object branch. we should use non-overlapping branches or `anyOf` and add fixtures for the typed and arbitrary forms.

crates/agentic-server/src/openapi.rs
```diff
@@ -0,0 +1,408 @@
...
+pub struct ApiErrorResponse {
+    pub error: ApiError,
+}
+
+#[derive(serde::Serialize, utoipa::ToSchema)]
+pub struct ApiError {
+    pub message: String,
+    #[serde(rename = "type")]
+    pub error_type: String,
+    pub code: String,
```
franciscojavierarceo: `ApiError.code` is a required string here, but `ToolError::MissingOutput` intentionally emits `"code": null` and has a unit test locking that behavior. the generated error schema therefore rejects a real 400 response. we should make `code` nullable and validate the existing error fixture against `ApiErrorResponse`.

crates/agentic-server/src/openapi.rs
```diff
@@ -0,0 +1,408 @@
...
+pub struct AnthropicError {
+    #[serde(rename = "type")]
+    pub error_type: String,
+    pub message: String,
+}
+
+/// Request body for POST /v1/conversations.
+#[derive(serde::Serialize, serde::Deserialize, utoipa::ToSchema)]
+pub struct CreateConversationRequest {
+    pub store: bool,
```
franciscojavierarceo, line 253: `store: bool` makes this property required, while the handler and existing integration tests accept `{}` or no body and default `store` to `true`. generated clients become stricter than the endpoint. we should model the body and property as optional with a default of `true`, or tighten the handler and tests to this contract.

crates/agentic-server-core/src/types/io/input.rs
```diff
@@ -94,13 +99,167 @@ pub enum ToolCallOutput {
...
+            )
+            .item(Ref::from_schema_name(schema_name))
+            .into()
+    }
+
+    impl utoipa::PartialSchema for InputItem {
+        fn schema() -> RefOr<Schema> {
+            OneOfBuilder::new()
+                .discriminator(Some(utoipa::openapi::schema::Discriminator::new("type")))
+                .item(tagged_ref("message", "InputMessage"))
```
franciscojavierarceo: `tagged_ref` makes `type: "message"` required, but the custom deserializer treats a missing `type` as a message and existing tests exercise `{"role":"user","content":"history"}`. the schema currently rejects that supported shorthand. we should make the message discriminator optional and add the shorthand to the schema-validation fixtures.

crates/agentic-server-core/src/types/io/tools.rs
```diff
@@ -33,19 +34,67 @@ pub enum ToolChoice {
...
+
+        let str_type = || ObjectBuilder::new().schema_type(SchemaType::new(Type::String));
+
+        OneOfBuilder::new()
+            .item(str_type().enum_values(Some(["auto", "none", "required"])))
+            .item(
+                ObjectBuilder::new()
+                    .property("type", str_type().enum_values(Some(["function"])))
+                    .required("type")
+                    .property("name", str_type())
```
franciscojavierarceo: the schema permits empty `name` values even though `NonEmptyToolName` rejects them, and it omits the legacy `{"function":{"name":"..."}}` form that `Deserialize` still accepts. validators diverge from the runtime in both directions. we should reuse the `NonEmptyToolName` schema and add the legacy object branch, with positive and negative fixtures.

franciscojavierarceo (changes requested):
the feature currently isn't exercised by CI, and several published schemas do not match request and response shapes that the runtime and existing tests accept. requesting changes so the generated contract is usable by clients and tested whenever `openapi` is enabled.

crates/agentic-server/Cargo.toml
avinashsingh77: made  openapi a default feature so it's included in standard builds and CI

crates/agentic-server/src/handler/http/messages.rs
avinashsingh77: - count_tokens — added request_body = MessagesRequest and body = CountTokensResponse
- /v1/messages — changed 200 JSON body from untyped () to MessagesResponse
- /v1/models — changed 200 body from untyped to ModelsResponse
- Added doc-only types: MessagesResponse, CountTokensResponse, ModelsResponse, ModelObject

crates/agentic-server-core/src/types/request_response.rs
avinashsingh77: The nullable schema validates both the minimal fixture and the explicit-null fixture, and serde confirms it deserializes correctly.

crates/agentic-server-core/src/types/io/output.rs
avinashsingh77: Changed McpCallError from oneOf to anyOf, replaced the generic object branch with AnyValue to cover all JSON types, and added fixtures for string, structured error, arbitrary object, number,  array, and null forms

crates/agentic-server/src/openapi.rs
avinashsingh77: Made ApiError.code nullable in the schema and added fixtures validating both "code": "missing_field" and "code": null.

crates/agentic-server/src/openapi.rs
avinashsingh77: store now has #[serde(default = "default_true")] and #[schema(default = true)]

crates/agentic-server-core/src/types/io/input.rs
avinashsingh77: The message branch's type discriminator is now optional, so {"role":"user", "content": "history"} validates against the schema, matching the deserializer's behavior.
Added the shorthand fixture to the test.

crates/agentic-server-core/src/types/io/tools.rs
avinashsingh77: - name fields in Function/Custom branches now use NonEmptyToolName ref (enforces minLength: 1)
- Added legacy {"function": {"name": "..."}} branch to the schema
- Added positive fixtures for all 8 accepted forms including legacy
- Added negative fixture verifying both schema and serde reject empty names

crates/agentic-server/src/handler/http/messages.rs
```diff
@@ -148,6 +164,18 @@ pub async fn messages(State(state): State<AppState>, request: Request) -> Respon
     proxy_messages(&state, parts, bytes, "/v1/messages").await
 }
 
+#[cfg_attr(feature = "openapi", utoipa::path(
+    post,
+    path = "/v1/messages/count_tokens",
+    request_body = agentic_core::types::messages::MessagesRequest,
```
franciscojavierarceo: `MessagesRequest` requires `max_tokens`, but token counting doesn't: `messages_count_tokens_uses_matching_upstream_path` already sends `{"model":"test","messages":[]}` and expects 200. this schema makes generated clients require a generation limit for a valid token-count request. we should use a dedicated token-count request schema and validate that existing fixture against it.

crates/agentic-server/src/handler/http/models.rs
```diff
@@ -222,6 +239,19 @@ pub struct ModelsParams {
...
 /// (`{ "models": [...] }` with rich metadata). Without `client_version`, the
 /// upstream response is returned unchanged via [`proxy_get`].
+#[cfg_attr(feature = "openapi", utoipa::path(
+    get,
+    path = "/v1/models",
+    params(
+        ("client_version" = Option<String>, Query, description = "Codex CLI version; triggers Codex-compatible model list shape"),
+    ),
+    responses(
+        (status = 200, description = "Model list", body = crate::openapi::ModelsResponse),
```
franciscojavierarceo: when `client_version` is supplied, this handler returns `{"models":[...]}`, but `ModelsResponse` requires `object` and `data`. the documented query option therefore produces a response that fails its own schema. we should describe both response shapes and add a fixture for the Codex catalog.

franciscojavierarceo (commented):
the earlier schema fixes are present, but two endpoint schemas still reject shapes the handlers support. details inline; these findings are from source and existing test fixtures, not a fresh full-suite run.

crates/agentic-server/src/handler/http/messages.rs
avinashsingh77: Created a dedicated CountTokensRequest (requires only model + messages, optional system/tools, no max_tokens), updated the handler annotation, and added fixtures matching the existing integration test payload

crates/agentic-server/src/handler/http/models.rs
avinashsingh77: Added CodexModelsResponse / CodexModelObject for the {"models": [...]} shape, updated the handler to document both response forms, and added fixtures for OpenAI list and  Codex catalog (including empty)

franciscojavierarceo (approved):

PR #234, opened 2026-09-02
fix: preserve Responses text compatibility

franciscojavierarceo:
## Summary

Follow-up to [#231](https://github.com/vllm-project/agentic-api/pull/231), addressing compatibility issues found after it merged.

- Preserve byte-for-byte stateless Responses proxying for provider-specific `text` formats while retaining typed routing validation and duplicate-field rejection.
- Preserve JSON Schema and `text` extension-field order recursively on executor-bound requests.
- Keep strict Clippy clean under the current Rust toolchain by boxing oversized futures inherited from `main`.
- Document the raw-text routing boundary using the repository's pass-through terminology.

## Test Plan

- [x] `cargo test`
- [x] `cargo clippy --all-targets -- -D warnings`
- [x] `cargo fmt --all -- --check`
- [x] `git diff --check`

### DIFF
--- a/crates/agentic-server-core/src/types/request_response.rs
+++ b/crates/agentic-server-core/src/types/request_response.rs
@@ -1,9 +1,8 @@
 use std::borrow::Cow;
 use std::collections::HashMap;
 
-use serde::ser::SerializeMap;
 use serde::{Deserialize, Serialize};
-use serde_json::{Value, json};
+use serde_json::{Map, Value, json};
 
 use super::io::{
     FunctionTool, InputItem, InputMessage, InputMessageContent, OutputItem, ResponseUsage, ResponsesInput, ToolChoice,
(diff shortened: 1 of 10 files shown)


crates/agentic-server-core/src/types/request_response.rs
```diff
@@ -70,8 +71,115 @@ pub enum ResponseTextFormat {
...
+            map.serialize_entry(VERBOSITY_FIELD, verbosity)?;
+        }
+        for (key, value) in &self.extra {
+            map.serialize_entry(key, value)?;
+        }
+        map.end()
+    }
+}
+
+impl Serialize for ResponseTextFormat {
```
maralbahari: Is this doing anything the derive did not? I compared both (serde 1.0.229 / serde_json 1.0.151 with `preserve_order` + `raw_value`, matching this PR's Cargo.toml) and the output is byte-identical for every case reachable from a request: plain, extension fields at either level, and json_schema with key order. `#[derive(Serialize)]` with `#[serde(flatten)]` plus `skip_serializing_if = "Option::is_none"` already emits exactly this.

The only difference is the reserved-key check, and that case cannot be reached from the wire:

1. `flatten` routes a known key to its modeled field during deserialization, so `"format"` or `"verbosity"` can never land in `extra`.
2. If a client sends the key twice, deserialization fails first: `duplicate field 'format' at line 1 column 52`. Serialization never runs. This PR's own `test_duplicate_routing_field_is_rejected_before_upstream` pins the same mechanism for `store`.

The only way to reach the guard is constructing the value in Rust, which nothing does except the new test:

```
derive, to_string: {"format":{"type":"text"},"verbosity":"low","format":"override"}   // duplicate key
manual, to_string: Err("text configuration extension key `format` conflicts...")
```

Against that, `TEXT_CONFIG_FIELDS` / `TEXT_FORMAT_FIELDS` / `JSON_SCHEMA_FORMAT_FIELDS` and the two `serialize` bodies are a hand-maintained mirror of the struct definitions that the compiler does not check. Add a field to `ResponseTextConfig` and forget the impl, and it is silently dropped from the upstream request. That trades an unreachable duplicate-key path for a reachable silent-data-loss one.

Suggestion: revert both impls to `#[derive(Serialize)]` with `skip_serializing_if = "Option::is_none"`. If the invariant is still worth enforcing, it is cheaper as a type constraint, for example `extra` private behind a validating constructor.

Separately: `extra` is a `HashMap`, so unmodeled keys serialize in a different order on every request. That predates this PR, but it sits awkwardly next to enabling `preserve_order` for key-order stability, since `schema` is now ordered while extension keys still shuffle. `serde_json::Map<String, Value>` would make the guarantee cover the whole `text` object.

franciscojavierarceo: verified this against the request path and you are right: serde routes modeled keys before flatten, so the manual guards only covered programmatic construction while mirroring the structs by hand. fixed in 6f447ba by restoring derived Serialize and switching the text extension maps to serde_json::Map, which also preserves extension-field order. text_configuration_preserves_extension_field_order now covers the complete text object.

maralbahari (approved):

PR #236, opened 2026-09-03
fix: align relayed SSE validation with providers

franciscojavierarceo:
## Summary

- Move stateless relayed SSE validation into `events/validate.rs` and keep cross-event lifecycle validation in `ResponseAccumulator`.
- Reuse the accumulator's in-flight and completed item state instead of maintaining a second state machine in `executor/upstream.rs`.
- Keep strict validation scoped to caller-supplied `decode_upstream` payloads while leaving live upstream streaming lenient.
- Accept provider-compatible event shapes, including missing function-call names, `item_id` fallbacks, different terminal item IDs, omitted terminal output, and SSE data fields without a space after the colon.
- Replay recorded Responses streams in regression coverage and continue rejecting duplicate or inconsistent terminal items.
- Preserve pre-existing lenient repeated-completion behavior; only strict decoding eagerly removes completed items from the active set.
- Cover repeated item completion, unsupported item types, and web-search `item_id` fallbacks with and without terminal output.

The broader typed pipeline, shared resolver, relay, and placement work is tracked in #241 and #243–#245. Deferral of the remaining architectural review comments is proposed, pending reviewer agreement.

Follow-up to #235, incorporating @maralbahari's review feedback.

## Test Plan

- Observed both new lenient repeated-completion tests fail with two output items before the fix, then pass afterward.
- `CARGO_PROFILE_TEST_DEBUG=0 CARGO_INCREMENTAL=0 cargo test --workspace -j 2 --quiet` (passed with local TCP listeners enabled; 9 tests ignored by their existing configuration).
- `CARGO_PROFILE_TEST_DEBUG=0 CARGO_PROFILE_DEV_DEBUG=0 CARGO_INCREMENTAL=0 cargo clippy --workspace --all-targets -- -D warnings`
- `/Users/farceo/.local/bin/uv run --no-project --python 3.12 --with pre-commit pre-commit run --all-files`
- Independent read-only review of the fix found no critical or important findings.

### DIFF
--- a/crates/agentic-server-core/src/events/normalize.rs
+++ b/crates/agentic-server-core/src/events/normalize.rs
@@ -5,12 +5,14 @@ use crate::utils::common::{deserialize_from_str_opt, deserialize_from_value_opt}
 
 /// Normalize a raw SSE data line into a typed [`EventFrame`].
 ///
-/// Expects input in the form `data: {...}` (the `data: ` prefix is required).
+/// Expects an SSE `data:` field, with or without the optional space after the
+/// colon.
 /// Returns `None` for non-data lines, empty lines, and the `data: [DONE]`
 /// sentinel.
 #[must_use]
 pub fn normalize_sse_line(line: &str) -> Option<EventFrame> {
-    let data_str = line.strip_prefix("data: ")?;
+    let data_str = line.strip_prefix("data:")?;
+    let data_str = data_str.strip_prefix(' ').unwrap_or(data_str);
     if data_str == "[DONE]" {
         return None;
     }
--- a/crates/agentic-server-core/src/events/normalize.rs
+++ b/crates/agentic-server-core/src/events/normalize.rs
@@ -137,12 +137,19 @@ fn extract_output_item_added(json: &Value) -> EventPayload {
 }
 
 fn extract_output_item_done(json: &Value) -> EventPayload {
-    let item = &json["item"];
+    let mut item = json["item"].clone();
+    let item_id = output_item_id(&item);
+    if !item_id.is_empty()
+        && item["id"].as_str().is_none_or(str::is_empty)
+        && let Some(item) = item.as_object_mut()
+    {
+        item.insert("id".to_owned(), Value::String(item_id.clone()));
+    }
     EventPayload::OutputItemDone {
-        item_id: output_item_id(item),
-        item_type: SSEItemType::from(json_str(item, "type")),
+        item_id,
+        item_type: SSEItemType::from(json_str(&item, "type")),
         output_index: json_u32(json, "output_index"),
-        item: item.clone(),
+        item,
     }
 }
 
--- a/crates/agentic-server-core/src/events/normalize.rs
+++ b/crates/agentic-server-core/src/events/normalize.rs
@@ -139,11 +139,10 @@ fn extract_output_item_added(json: &Value) -> EventPayload {
 fn extract_output_item_done(json: &Value) -> EventPayload {
     let mut item = json["item"].clone();
     let item_id = output_item_id(&item);
-    if !item_id.is_empty()
-        && item["id"].as_str().is_none_or(str::is_empty)
-        && let Some(item) = item.as_object_mut()
-    {
-        item.insert("id".to_owned(), Value::String(item_id.clone()));
+    if !item_id.is_empty() && item["id"].as_str().is_none_or(str::is_empty) {
+        if let Some(item) = item.as_object_mut() {
+            item.insert("id".to_owned(), Value::String(item_id.clone()));
+        }
     }
     EventPayload::OutputItemDone {
         item_id,
--- a/crates/agentic-server-core/src/events/normalize.rs
+++ b/crates/agentic-server-core/src/events/normalize.rs
@@ -19,6 +19,13 @@ pub fn normalize_sse_line(line: &str) -> Option<EventFrame> {
     normalize_sse_value(json)
 }
 
+/// Whether a line carries an SSE data payload that should normalize to a frame.
+pub(crate) fn is_data_frame(line: &str) -> bool {
+    line.strip_prefix("data:")
+        .map(str::trim)
+        .is_some_and(|payload| !payload.is_empty() && payload != "[DONE]")
+}
+
 /// Normalizes an already parsed SSE payload.
 pub(crate) fn normalize_sse_value(json: Value) -> Option<EventFrame> {
     let event_type = json
@@ -90,6 +97,15 @@ fn json_str_opt(json: &Value, key: &str) -> Option<String> {
     json[key].as_str().map(ToString::to_string)
 }
 
+fn output_item_id(item: &Value) -> String {
+    item["id"]
+        .as_str()
+        .filter(|id| !id.is_empty())
+        .or_else(|| item["item_id"].as_str().filter(|id| !id.is_empty()))
+        .unwrap_or_default()
+        .to_owned()
+}
+
 fn json_u32(json: &Value, key: &str) -> u32 {
     u32::try_from(json[key].as_u64().unwrap_or(0)).unwrap_or(u32::MAX)
 }
@@ -109,7 +125,7 @@ fn extract_response_payload(json: &Value) -> EventPayload {
 fn extract_output_item_added(json: &Value) -> EventPayload {
     let item = &json["item"];
     EventPayload::OutputItemAdded {
-        item_id: json_str(item, "id"),
+        item_id: output_item_id(item),
         item_type: SSEItemType::from(json_str(item, "type")),
         output_index: json_u32(json, "output_index"),
         name: json_str_opt(item, "name"),
@@ -121,7 +137,7 @@ fn extract_output_item_added(json: &Value) -> EventPayload {
 fn extract_output_item_done(json: &Value) -> EventPayload {
     let item = &json["item"];
     EventPayload::OutputItemDone {
-        item_id: json_str(item, "id"),
+        item_id: output_item_id(item),
         item_type: SSEItemType::from(json_str(item, "type")),
         output_index: json_u32(json, "output_index"),
         item: item.clone(),
--- a/crates/agentic-server-core/src/events/validate.rs
+++ b/crates/agentic-server-core/src/events/validate.rs
@@ -2,15 +2,28 @@ use serde::Deserialize;
 use serde_json::{Map, Value};
 use thiserror::Error;
 
-use super::{EventFrame, SSEEventType};
+use super::{EventFrame, SSEEventType, SSEItemType};
 use crate::types::io::OutputItem;
 
 #[derive(Debug, Error)]
 #[error("{0}")]
 pub(crate) struct EventError(String);
 
+#[derive(Debug)]
+pub(crate) struct ValidatedFrame<'a> {
+    pub(crate) item: Option<ValidatedItem<'a>>,
+}
+
+#[derive(Debug)]
+pub(crate) struct ValidatedItem<'a> {
+    pub(crate) item_id: &'a str,
+    pub(crate) output_index: u32,
+    pub(crate) item_type: SSEItemType,
+    pub(crate) done_item: Option<OutputItem>,
+}
+
 /// Validates the stateless wire-format requirements of one normalized frame.
-pub(crate) fn validate_frame(frame: &EventFrame) -> Result<(), EventError> {
+pub(crate) fn validate_frame(frame: &EventFrame) -> Result<ValidatedFrame<'_>, EventError> {
     let event_name = frame
         .wire
         .event_type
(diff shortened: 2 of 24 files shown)


crates/agentic-server-core/src/events/normalize.rs
```diff
@@ -121,7 +139,7 @@ fn extract_output_item_added(json: &Value) -> EventPayload {
 fn extract_output_item_done(json: &Value) -> EventPayload {
     let item = &json["item"];
     EventPayload::OutputItemDone {
-        item_id: json_str(item, "id"),
+        item_id: output_item_id(item),
         item_type: SSEItemType::from(json_str(item, "type")),
         output_index: json_u32(json, "output_index"),
         item: item.clone(),
```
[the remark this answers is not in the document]
franciscojavierarceo: fixed in 9aaa8f1. `response.output_item.done` now canonicalizes the retained item `id` from `item_id` before deserialization while preserving a valid existing `id`. the regression test verifies authoritative completed content replaces partial streamed content.

crates/agentic-server-core/src/executor/accumulator.rs
```diff
@@ -675,6 +903,75 @@ fn in_flight_matches_call_type(item: &InFlight, item_type: SSEItemType) -> bool
...
+        InFlight::FunctionCall { .. } => "function_call",
+        InFlight::CustomToolCall { .. } => "custom_tool_call",
+        InFlight::WebSearchCall { .. } => "web_search_call",
+        InFlight::McpCall { .. } => "mcp_call",
+        InFlight::McpListTools { .. } => "mcp_list_tools",
+        InFlight::Compaction { .. } => "compaction",
+    }
+}
+
+fn output_item_type(item: &OutputItem) -> Option<&'static str> {
```
maralbahari: These two helpers, plus `in_flight_item_type` below and both copies of `ensure_supported_output_item_type`, all re-encode a mapping `SSEItemType` already owns. Its 8 variants are exactly the 8 known `OutputItem` variants, and it already has `as_str()`, `From<&str>`, and `PartialEq<&str>`.

Two additions collapse all of it:

```rust
// types/io/output.rs
impl OutputItem {
    /// The item's wire id, absent for [`OutputItem::Unknown`].
    #[must_use]
    pub fn id(&self) -> Option<&str> { ... }
}

// events/types.rs
impl TryFrom<&OutputItem> for SSEItemType {
    type Error = ();
    // ... Unknown => Err(())
}

/// Strict counterpart to the lenient `From<&str>`, which folds unknown types into `Message`.
impl std::str::FromStr for SSEItemType {
    type Err = ();
    // ... unknown => Err(())
}
```

crates/agentic-server-core/src/executor/accumulator.rs
```diff
@@ -270,11 +286,206 @@ impl ResponseAccumulator {
...
+                    ..
+                },
+            ) => {
+                self.require_in_progress(event_name)?;
+                if self.has_output_index(*output_index) || self.has_item_id(item_id) {
+                    return Err(invalid_stream(format!(
+                        "upstream stream repeats output item '{item_id}'"
+                    )));
+                }
+                if item_type.as_str() != raw_output_item_type(frame)? {
```
maralbahari: This guard is unreachable, and `raw_output_item_type` exists only to feed it.

`validate_frame` has already run `ensure_supported_output_item_type` on this same item, so `item_type` is one of the supported 8, and `SSEItemType::from` round-trips all 8 exactly (`"message"` -> `Message` -> `"message"`, and so on). The two sides can never differ. Anything unsupported errors earlier with the correct message.

crates/agentic-server-core/src/executor/accumulator.rs
```diff
@@ -270,11 +286,206 @@ impl ResponseAccumulator {
...
+                self.require_in_progress(event_name)?;
+                self.require_active_item(*output_index, item_id, item_type.as_str(), event_name)?;
+            }
+            (SSEEventType::Other, _) => self.require_in_progress(event_name)?,
+            (_, _) => {
+                self.require_in_progress(event_name)?;
+                let item_id = frame
+                    .wire
+                    .rest
+                    .get("item_id")
```
maralbahari: `item_id` and `output_index` were already pulled out of `wire.rest` by `validate_frame` a few frames up (`validate.rs:31-32`), which runs immediately before this. Re-reading them here is not just duplicate work, it produces a second, divergent set of error messages: `"...has no item id"` / `"...has no output index"` here versus `"...has no valid 'item_id'"` there. Only `validate_frame`'s can ever be reached, so these two are dead strings.

Same pattern with `output_item_identity`, which runs once in `validate_output_item` and again via `raw_output_item_type`, and with `OutputItem`, which is deserialized once in `validate_output_item` and again in `complete_call_item`.

Smallest fix: have `validate_frame` return a typed summary (`item_id`, `output_index`, `item_type`, the deserialized `OutputItem`) instead of `()`, and let `validate_strict_transition` and `complete_call_item` consume it rather than re-deriving it.

crates/agentic-server-core/src/events/validate.rs
```diff
@@ -0,0 +1,265 @@
...
+    };
+    if status == expected_status {
+        return Ok(());
+    }
+    Err(invalid(format!(
+        "upstream stream event '{event_name}' has status '{status}', expected '{expected_status}'"
+    )))
+}
+
+fn validate_output_item(frame: &EventFrame, event_name: &str, complete: bool) -> Result<(), EventError> {
```
maralbahari: The `item_id` fallback added here (and in `normalize.rs`) is not mirrored in `complete_call_item`, which deserializes the **raw** item rather than this `canonical` one. `WebSearchCall::id` is a required serde field (`types/io/output.rs:350`, no `#[serde(default)]`), so the raw deserialize fails, `InFlight::WebSearchCall { item: None }` never receives a value, and `finalize()` returns `None`.

Measured on this branch, one stream with only the done-event id field changed:

| done item id field | terminal `output` | result |
|---|---|---|
| `id` | present | `output len 1` |
| `item_id` | present | `ERR: terminal upstream response output does not match completed item events` |
| `id` | omitted | `output len 1` |
| `item_id` | omitted | **`output len 0`, silently dropped** |

So the stream is accepted and the item vanishes. `function_call` survives because `FunctionToolCall::id` has `#[serde(default)]`; `message` survives because `finalize()` keeps the added-time id. All 7 real occurrences in the cassettes are `function_call`, which is why the new replay test does not catch it.

Cleanest fix is to have `complete_call_item` consume the canonicalized item this function already builds, instead of canonicalizing twice with different rules.

crates/agentic-server-core/src/executor/upstream.rs
```diff
@@ -76,238 +76,6 @@ fn absorb_line(acc: &mut ResponseAccumulator, ctx: &RequestContext, line: &str)
...
-                return Err(ExecutorError::InvalidRequest(
-                    "terminal upstream response output does not match completed item events".to_owned(),
-                ));
-            }
-        }
-        Ok(())
-    }
-}
-
 fn ensure_supported_output_item_type(item_type: &str) -> ExecutorResult<()> {
```
maralbahari: This is now a verbatim duplicate of `events/validate.rs:90`. Both become one line once `SSEItemType` gets a strict string parse (see my note on `accumulator.rs:919`):

```rust
fn ensure_supported_output_item_type(item_type: &str) -> ExecutorResult<()> {
    item_type
        .parse::<SSEItemType>()
        .map(drop)
        .map_err(|()| ExecutorError::InvalidRequest(format!("upstream output item type '{item_type}' is unsupported")))
}
```

Also, `validate.rs`'s copy is marked `pub(crate)` but is not re-exported from `events/mod.rs`, so it is effectively private already.

crates/agentic-server-core/src/executor/accumulator.rs
```diff
@@ -545,7 +767,10 @@ impl ResponseAccumulator {
         } else {
             deserialize_from_value_opt::<OutputItem>(raw_item.clone())
         };
-        if let Some(entry) = in_flight_key.as_deref().and_then(|key| self.in_flight.get_mut(key)) {
+        if let Some(mut entry) = in_flight_key
+            .as_deref()
+            .and_then(|key| self.in_flight.shift_remove(key))
```
maralbahari: `get_mut` -> `shift_remove` changes behaviour on the lenient live path, not just the strict one.

Previously a `done` event mutated the in-flight entry and left it in `in_flight`; items reached `completed` only via `finalize_all()` at end of stream. Now `done` removes the entry and pushes it to `completed` immediately.

That is required by the new validator, which reads `in_flight` and `completed` as its state: `require_active_item` looks for an open item at an `output_index`, the terminal arm errors on `!self.in_flight.is_empty()`, and `validate_terminal_output` compares against `completed.len()`. I confirmed the dependency by restoring `get_mut`, which made every strict decode fail with `upstream stream ended with unfinished output items`. So this is not revertible on its own.

The fallout lands on the path with no `require_active_item` guard. A repeated `response.output_item.done` used to be idempotent (`get_mut` found the same entry twice). Now the second one finds nothing in `in_flight`, falls through to the "was never in flight" branch below, and pushes a duplicate:

- base `ccb24fd`: `output len 1`
- this branch: `output len 2`

Stream: `added`, `function_call_arguments.done`, `done`, `done`, `completed`, through `ResponseAccumulator::from_sse_lines`. Strict relay rejects the second `done`, so this only affects `fetch_stream_payload`.

Worth a regression test either way. The shared resolver I suggested on `validate_strict_transition` would make it fall out naturally, since the duplicate becomes a check on the lookup rather than a consequence of which collection the entry sits in.

crates/agentic-server-core/src/executor/accumulator.rs
```diff
@@ -270,11 +286,206 @@ impl ResponseAccumulator {
...
+        self.process_normalized_event(&frame);
+        Ok(Some(frame))
+    }
+
     pub(crate) fn process_normalized_event(&mut self, frame: &EventFrame) {
         self.capture_terminal_details_if_needed(frame);
         self.process_event(frame);
     }
 
+    fn validate_strict_transition(&self, frame: &EventFrame) -> ExecutorResult<()> {
```
maralbahari: This match and `process_event` destructure the same `(event_type, payload)` pairs and then resolve the same in-flight entry, so every strict frame is walked twice.

For one `response.output_text.delta`, `item_id` is read four times: `extract_payload` puts it in the typed payload, `deserialize_from_value_opt::<WireEvent>` walks the JSON again into `wire.rest`, `validate_frame` reads it back out of `wire.rest`, and this function reads it out of `wire.rest` a third time before `process_event` finally uses the typed copy.

The resolution step is the expensive half. `require_active_item` does

```rust
self.in_flight.iter().find(|(_, entry)| entry.output_index == output_index)
```

a linear scan, while `process_event` immediately does `in_flight.get_mut(item_id)`, a hash lookup on the same `IndexMap`. Every delta pays a scan plus a lookup where one lookup would serve, and deltas dominate a stream.

They also disagree on how to resolve: `require_active_item` keys on `output_index` then asserts the id, `complete_call_item` uses `in_flight_call_key` (id first, `output_index`+type fallback), and the `OutputTextDelta` / `CustomToolCallInputDelta` arms use `get_mut(item_id)` ignoring `output_index`. Those agree today only because `has_output_index`/`has_item_id` rule out the collisions that would separate them, which is a fragile invariant to rely on.

Suggest a single resolver both paths share:

```rust
fn resolve_active(
    &mut self,
    item_id: &str,
    output_index: u32,
    expected: SSEItemType,
    strict: bool,
) -> ExecutorResult<Option<&mut InFlightEntry>>
```

Lenient callers keep "skip if absent"; strict callers get the error from the same lookup that does the work. `require_active_item` and this pre-pass go away, and `validate_strict_transition` keeps only what has no counterpart in `process_event`: the lifecycle transitions and `validate_terminal_output`.

crates/agentic-server-core/src/executor/accumulator.rs
franciscojavierarceo: fixed in 554ed21. the shared completion fold uses `get_mut` again, and only `process_strict_sse_line` removes/finalizes the entry after validation and folding. that restores the previous lenient behavior without breaking strict terminal checks. both new lenient tests failed with `output.len() == 2` before the fix and now pass; strict repeated completion still errors. conflicting-content rejection and client-event deduplication remain proposed policy changes in #243, not part of this compatibility fix.

crates/agentic-server-core/src/events/validate.rs
franciscojavierarceo: i checked this against the current branch: 9aaa8f1 already writes the fallback ID into the retained `EventPayload::OutputItemDone.item`, so `complete_call_item` no longer receives the uncanonicalized object. 554ed21 adds a web-search regression matrix for valid/missing/null/empty `id`, with terminal output present and omitted. all eight cases retain one item with the expected ID and search action. sharing the validated typed item is still worthwhile in #243.

crates/agentic-server-core/src/executor/accumulator.rs
franciscojavierarceo: removed the unreachable guard and `raw_output_item_type` in 554ed21. the strict unsupported-item regression confirms that frame validation still rejects an unknown type before the lifecycle pass.

crates/agentic-server-core/src/executor/upstream.rs
franciscojavierarceo: removed the executor copy in 554ed21 and re-exported the existing validator from `events`, so JSON and streaming validation now share the same supported-type check. the broader strict `SSEItemType` conversion cleanup is proposed for #243.

crates/agentic-server-core/src/executor/accumulator.rs
franciscojavierarceo: 554ed21 removes the duplicated supported-type check. can we carry the remaining ID/kind conversions into #243 alongside the typed slot representation? that keeps this PR focused on provider compatibility and the live-path regression. leaving this thread open for your confirmation.

crates/agentic-server-core/src/executor/accumulator.rs
franciscojavierarceo: removed `raw_output_item_type` and its dead guard in 554ed21. the typed validation summary would change the normalize/validate/fold boundary, which is the core of #243. can we handle the remaining repeated extraction/deserialization there? leaving this open rather than treating the deferral as agreed.

crates/agentic-server-core/src/executor/accumulator.rs
franciscojavierarceo: the live duplicate-output regression is fixed in 554ed21 without changing the resolver contract: eager removal now happens only in the strict driver after the shared fold. can we move the shared resolver and single traversal into #243? those should land together with the slot identity rules. leaving this thread open for your confirmation.

crates/agentic-server-core/src/executor/accumulator.rs
franciscojavierarceo: implemented in 11fcad3. `OutputItem::id`, `TryFrom<&OutputItem> for SSEItemType`, and strict `FromStr` now own these conversions; the accumulator no longer re-encodes the mapping. added unit coverage for known and unknown variants.

crates/agentic-server-core/src/executor/accumulator.rs
franciscojavierarceo: implemented in 11fcad3. `validate_frame` now returns `ValidatedFrame` with the typed ID, validated index and type, and canonical deserialized done item. strict transition and completion consume that summary, so they no longer reread or deserialize the same fields.

crates/agentic-server-core/src/executor/accumulator.rs
franciscojavierarceo: implemented in 11fcad3. strict and lenient folding now share `resolve_active`, backed by direct item-ID lookup and an explicit index map, with no per-delta linear scan. explicit ID and index contradictions are rejected, missing indexes do not silently claim slot 0, strict repeated `done` is rejected, and lenient repeats are emitted only once when semantically identical while conflicting authoritative content returns an error. the constructor and async stream paths propagate that error too.

maralbahari (approved):

PR #237, opened 2026-09-03
fix: reject unusable relayed call IDs

StevenWang-CY:
## Summary

- Reject completed or incomplete split-execution responses whose `function_call` or `custom_tool_call` output has a missing/empty `call_id`, repeats one within the response, or reuses one from the rehydrated response chain or the current replayed input.
- In caller-relayed SSE, bind each tool call's first non-empty `call_id` across `response.output_item.added`, function-argument events, `response.output_item.done`, and terminal output, rejecting later changes before persistence.
- Validate before the first write so a rejected payload does not consume the reserved response ID and can be corrected and retried. Failed partial responses retain their existing returned-but-unstored behavior.

## Problem and contract

The split `commit` path deserialized an omitted `call_id` as an empty string and persisted the terminal response. It also checked neither resolved IDs from the continued history nor changes between SSE representations. The resulting response ID was accepted but could fail on the next continuation, or a client could answer the originally announced ID while a different ID was persisted.

The repository's continuation state treats call IDs as opaque, globally unique identifiers within the rehydrated sequence. The upstream types also require these IDs:

- OpenAI response function calls: https://github.com/openai/openai-python/blob/main/src/openai/types/responses/response_function_tool_call.py
- OpenAI custom tool calls: https://github.com/openai/openai-python/blob/main/src/openai/types/responses/response_custom_tool_call.py
- vLLM call-ID construction: https://github.com/vllm-project/vllm/blob/31e9c13685b484d4ba67dc4e73019b23402609c7/vllm/entrypoints/openai/responses/utils.py#L93-L108

A scan of the recorded corpus found no missing, empty, or repeated finalized call ID. Across 62 recorded SSE call lifecycles, added/done/intermediate representations never changed `call_id`. Older cassettes with terminal representation differences already violate the split validator's item-identity contract and are not newly rejected by this patch.

## Red/green reproduction

Against unmodified reviewed head `60aa5dc`, test-only controls failed because all of these payloads were accepted and persisted:

- JSON and SSE responses reusing a resolved historical function-call ID in a later custom call
- an SSE call announcing ID A in `response.output_item.added` and persisting ID B from `response.output_item.done`

A fresh counterexample against the first implementation also failed: `response.function_call_arguments.done` could replace the announced ID even when done/terminal item representations later returned to the original ID. The final implementation observes that mutation as well.

The final regressions cover JSON/SSE, function/custom cross-kind reuse, manually replayed current input, added/done/terminal and argument-event mutation, completed/incomplete/failed terminal states, initially empty IDs filled later, malicious CRLF markers, corrected retry, and unchanged failed-response persistence behavior. All rejection diagnostics use bounded indices and never echo the upstream ID.

## Compatibility and scope

- The additional history read occurs only for a completed/incomplete payload containing a new function or custom call.
- Failed responses remain returnable and unstored even if partial SSE representations disagree.
- The normal in-process upstream accumulator is unchanged; stability enforcement is limited to caller-supplied split SSE.
- No API, wire-format, or persistence schema changes are introduced.
- PR #236 changes the strict SSE validator's architecture and does not currently enforce this invariant, so merge order may require retaining this regression while resolving that file mechanically.

## Test Plan

Exact final commit `7f9bf9964209de75acebe72a0fe2fb854a12d975` was tested from a clean detached worktree:

- `cargo check --workspace --all-targets --all-features`
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- `cargo test -p agentic-llm-d --test split_execution_integration` (19 passed)
- `cargo test --workspace --all-features -- --test-threads=4` (all runnable tests passed; 5 PostgreSQL tests require `TEST_POSTGRES_URL`, and 1 doctest is repository-ignored)
- Python suite (111 passed, 3 package-install scenarios skipped by the suite)
- cassette validator (108/108 cassettes, 238 turns)
- source-install CLI end-to-end test
- strict MkDocs build
- all-file pre-commit hooks; Apple Git deadlocked in `check-added-large-files --all-files`, so that hook was rerun successfully on the complete four-file diff while every other hook passed over all files

### DIFF
--- a/crates/agentic-server-core/src/executor/persist.rs
+++ b/crates/agentic-server-core/src/executor/persist.rs
@@ -3,6 +3,8 @@
 //! Writes the completed response and output items to storage, routing to the
 //! appropriate handler based on whether the turn belongs to a conversation.
 
+use std::collections::HashMap;
+
 use crate::executor::error::{ExecutorError, ExecutorResult};
 use crate::executor::modes::{ConversationHandler, ResponseHandler};
 use crate::executor::request::{ExecutionContext, RequestContext};
@@ -85,7 +87,7 @@ pub async fn persist_turn(
 /// returned unstored, as the in-process flow does.
 ///
 /// # Errors
-/// [`ExecutorError::InvalidRequest`] for an unusable id or unfinished response,
+/// [`ExecutorError::InvalidRequest`] for unusable IDs or an unfinished response,
 /// [`ExecutorError::Conflict`] for an id already stored, or a storage error.
 pub async fn commit(
     ctx: RequestContext,
@@ -98,12 +100,16 @@ pub async fn commit(
         ));
     }
 
+    let status = payload.status.parse::<ResponseStatus>().unwrap_or_default();
     // Storing an unfinished turn would return an id that can never be continued.
-    if payload.status.parse::<ResponseStatus>().unwrap_or_default() == ResponseStatus::InProgress {
+    if status == ResponseStatus::InProgress {
         return Err(ExecutorError::InvalidRequest(format!(
             "upstream response status '{}' is not terminal",
             payload.status
         )));
+    }
+    if matches!(status, ResponseStatus::Completed | ResponseStatus::Incomplete) {
+        validate_output_call_ids(&payload.output)?;
     }
 
     persist_if_needed(
@@ -114,4 +120,26 @@ pub async fn commit(
     )
     .await?;
     Ok(payload)
+}
+
+fn validate_output_call_ids(output_items: &[OutputItem]) -> ExecutorResult<()> {
+    let mut call_ids = HashMap::new();
+    for (index, item) in output_items.iter().enumerate() {
+        let (item_type, call_id) = match item {
+            OutputItem::FunctionCall(call) => ("function_call", call.call_id.as_str()),
+            OutputItem::CustomToolCall(call) => ("custom_tool_call", call.call_id.as_str()),
+            _ => continue,
+        };
+        if call_id.is_empty() {
+            return Err(ExecutorError::InvalidRequest(format!(
+                "upstream response output[{index}] {item_type} has no valid 'call_id'"
+            )));
+        }
+        if let Some(first_index) = call_ids.insert(call_id, index) {
+            return Err(ExecutorError::InvalidRequest(format!(
+                "upstream response output[{index}] repeats 'call_id' from output[{first_index}]"
+            )));
+        }
+    }
+    Ok(())
 }
--- a/crates/agentic-server-core/src/executor/persist.rs
+++ b/crates/agentic-server-core/src/executor/persist.rs
@@ -8,8 +8,9 @@ use std::collections::HashMap;
 use crate::executor::error::{ExecutorError, ExecutorResult};
 use crate::executor::modes::{ConversationHandler, ResponseHandler};
 use crate::executor::request::{ExecutionContext, RequestContext};
+use crate::storage::InOutItem;
 use crate::types::event::ResponseStatus;
-use crate::types::io::OutputItem;
+use crate::types::io::{InputItem, OutputItem};
 use crate::types::request_response::ResponsePayload;
 use tracing::error;
 
@@ -109,7 +110,7 @@ pub async fn commit(
         )));
     }
     if matches!(status, ResponseStatus::Completed | ResponseStatus::Incomplete) {
-        validate_output_call_ids(&payload.output)?;
+        validate_output_call_ids(&ctx, &payload.output, &exec_ctx.resp_handler).await?;
     }
 
     persist_if_needed(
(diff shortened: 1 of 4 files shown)


crates/agentic-server-core/src/executor/persist.rs
```diff
@@ -98,13 +100,17 @@ pub async fn commit(
         ));
     }
 
+    let status = payload.status.parse::<ResponseStatus>().unwrap_or_default();
```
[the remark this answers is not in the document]
StevenWang-CY: No change needed here: ResponseStatus implements FromStr with Err = Infallible, and every string maps to a status; unknown values map to Error, so unwrap_or_default cannot take the default branch. Caller-supplied JSON statuses are also checked by ensure_strict_response before commit, with queued and unknown values covered as HTTP 400, while SSE terminal lifecycle is validated during normalization. This refactor only avoids parsing the same status twice and preserves the existing behavior.

crates/agentic-server-core/src/executor/persist.rs
```diff
@@ -115,3 +121,25 @@ pub async fn commit(
...
+        let (item_type, call_id) = match item {
+            OutputItem::FunctionCall(call) => ("function_call", call.call_id.as_str()),
+            OutputItem::CustomToolCall(call) => ("custom_tool_call", call.call_id.as_str()),
+            _ => continue,
+        };
+        if call_id.is_empty() {
+            return Err(ExecutorError::InvalidRequest(format!(
+                "upstream response output[{index}] {item_type} has no valid 'call_id'"
+            )));
+        }
```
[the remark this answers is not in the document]
StevenWang-CY: I am keeping the exact empty-string check. pending_calls treats call IDs as opaque identifiers: it rejects only empty values and keys and resolves outputs with the unchanged string. A whitespace-only ID can therefore round-trip without poisoning continuation, and the upstream string contracts do not specify trimming or a non-whitespace constraint. Using trim here would broaden validation beyond the reproduced failure and could reject an identifier the existing lifecycle can resolve.

crates/agentic-server-core/src/executor/persist.rs
```diff
@@ -115,3 +121,25 @@ pub async fn commit(
...
+        if !call_ids.insert(call_id) {
+            return Err(ExecutorError::InvalidRequest(format!(
+                "upstream response repeats call_id '{call_id}'"
+            )));
+        }
```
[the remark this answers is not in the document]
StevenWang-CY: Addressed in 52d6117. The duplicate diagnostic now reports the repeated output index and the first matching output index without interpolating the upstream call_id. The JSON and SSE regression uses a CRLF/control marker; both tests fail against the prior head because that marker reaches the error text and pass with the amended implementation.

crates/agentic-server-core/src/executor/persist.rs
```diff
@@ -98,13 +100,17 @@ pub async fn commit(
...
     // Storing an unfinished turn would return an id that can never be continued.
-    if payload.status.parse::<ResponseStatus>().unwrap_or_default() == ResponseStatus::InProgress {
+    if status == ResponseStatus::InProgress {
         return Err(ExecutorError::InvalidRequest(format!(
             "upstream response status '{}' is not terminal",
             payload.status
         )));
     }
+    if matches!(status, ResponseStatus::Completed | ResponseStatus::Incomplete) {
+        validate_output_call_ids(&payload.output)?;
```
franciscojavierarceo: `validate_output_call_ids` only checks the current response, so a resolved `call_id` can be reused in a later response, persisted, and then rejected when the next continuation rebuilds history. we should validate new output IDs against the rehydrated history before persistence while keeping diagnostics free of the raw upstream ID.

crates/agentic-server-core/src/executor/persist.rs
```diff
@@ -115,3 +121,25 @@ pub async fn commit(
     .await?;
     Ok(payload)
 }
+
+fn validate_output_call_ids(output_items: &[OutputItem]) -> ExecutorResult<()> {
```
franciscojavierarceo: the SSE path validates only the final accumulated payload, so a stream can announce `call_id=A` in `response.output_item.added`, change it to `B` in `response.output_item.done`, and persist B. a client responding to A can no longer continue the response chain. we should bind the first non-empty `call_id` across the added, done, and terminal representations and reject any later change before persistence.

crates/agentic-server-core/src/executor/persist.rs
StevenWang-CY: Confirmed and fixed in `7f9bf99`. `commit` now collects the new function/custom call IDs, rehydrates the prior response chain only when such calls exist, and rejects exact matches against stored calls or calls manually replayed in the current input before any write. Diagnostics report only bounded output/history/input indices and never interpolate the raw ID. The JSON and SSE history regressions fail on reviewed head `60aa5dc` because the repeated call is persisted, and pass at the final head; each also proves that a corrected retry can still use the reserved response ID.

crates/agentic-server-core/src/executor/persist.rs
StevenWang-CY: Confirmed and fixed in `7f9bf99`. The strict relay validator now binds the first non-empty call ID and observes it across `response.output_item.added`, function-argument delta/done events, `response.output_item.done`, and terminal output. A fresh adversarial test caught the argument-done overwrite after the first implementation, so that path is covered too. Empty-to-first-non-empty remains valid; changes are rejected for completed/incomplete responses, while failed partial responses keep the existing returned-but-unstored behavior. These regressions fail on `60aa5dc` and pass at the final head.

franciscojavierarceo (approved):

PR #240, opened 2026-09-04
feat: multiplex response streams over WebSocket

### LINKED ISSUE #239: Feature: support stream_id multiplexing for concurrent responses over a single WebSocket connection
### Summary

The WebSocket Responses endpoint accepts a `stream_id` field on `response.create`, but it does not actually support multiplexing multiple concurrent responses over a single connection. This is a feature request to make `stream_id` a functional multiplexing key so that several in-flight responses (e.g. forked conversations) can share one WebSocket connection and be demultiplexed by the client.

### Motivation

Clients that fork a conversation (one parent turn, multiple child turns chaining off the same `previous_response_id`) currently must open a separate WebSocket connection per concurrent child, because there is no reliable way to tell which events belong to which in-flight response on a shared socket. A working `stream_id` multiplexing scheme would let a client run N concurrent responses on a single connection, reducing connection overhead and matching the intent of the `stream_id` field the API already accepts.

### Current behavior (observed)

Testing against the vLLM agentic-api WebSocket Responses endpoint (`/v1/responses`, model `/models/models/GLM-5.2-FP8`), with one parent turn (`store: true`) followed by two forks fired concurrently on the same connection — each with a distinct `stream_id` (`forkA`, `forkB`) and each chaining `previous_response_id=<parent>`:

1. **`stream_id` is not echoed on any event.** Across 123 streamed events, every event had `stream_id: null`. The value sent on `response.create` is accepted but never returned, so it cannot be used to route events.
2. **Only wrapper events carry a response id.** `response.created` / `response.in_progress` / `response.completed` embed the full `response` object (with `id`). The high-frequency streaming events — `response.reasoning_text.delta`, `response.output_text.delta`, `response.output_item.added`, etc. — carry only `item_id`, `output_index`, `content_index`, and a per-response `sequence_number`. They carry **no** `response.id` and **no** `stream_id`, so a delta cannot be attributed to a response.
3. **Concurrent requests are serialized, not interleaved.** The two forks did not interleave: forkA ran to completion (`sequence_number` 0→50, `response.completed`) before forkB began (its `sequence_number` reset to 0). Timestamps confirm no overlap (forkA `created_at≈…494` → completed `≈…503`; forkB started `≈…503` → completed `≈…517`).

Note: fork inheritance itself works — both children correctly recalled the parent context (`pineapple`, `zebra`). The gap is specifically in multiplexing.

### Why this blocks multiplexing

For a client to multiplex concurrent responses on one socket it needs a demux key present on **every** event. Today:
- `stream_id` is dropped, and
- delta events carry neither `stream_id` nor `response.id`.

So if the server ever did interleave two responses, their delta tokens would be un-attributable. Combined with the server serializing requests, there is currently no throughput or demux benefit to sending multiple `response.create` messages on one connection.

### Requested behavior

1. Echo the client-supplied `stream_id` on **every** streamed event for that response (all `response.*` events, including every `*.delta`), so clients can demultiplex.
2. Allow multiple `response.create` requests to be genuinely in-flight concurrently on a single WebSocket connection, with their events interleaved and each tagged by `stream_id`.
3. (Alternative/if `stream_id` is not adopted) Include the owning `response.id` on every event as a demux key.

### Environment

- Endpoint: vLLM agentic-api WebSocket Responses endpoint (`/v1/responses`), reached via `kubectl port-forward svc/agentic-api -n agentic-api 19000:9000` → `ws://localhost:19000/v1/responses`
- Model: `/models/models/GLM-5.2-FP8`

### Reproduction

Each `response.create` envelope:

```json
{
  "type": "response.create",
  "model": "/models/models/GLM-5.2-FP8",
  "input": [{"role": "user", "content": "<text>"}],
  "max_output_tokens": 512,
  "stream_id": "<forkA|forkB>"
}
```

1. Open one WebSocket connection.
2. Send a parent `response.create` with `"store": true`, input `"Remember these two facts: the fruit is pineapple and the animal is zebra. Acknowledge with OK."`. Read to `response.completed`; capture `response.id`.
3. On the **same** connection, send two `response.create` messages back-to-back without waiting:
   - `stream_id:"forkA"`, `previous_response_id:<parent>`, input `"What fruit did I ask you to remember? Reply with only that word."`
   - `stream_id:"forkB"`, `previous_response_id:<parent>`, input `"What animal did I ask you to remember? Reply with only that word."`
4. Read all events on the single socket and attempt to route each event to forkA/forkB by its `stream_id`.

**Observed:** every event has `stream_id: null` (cannot route); delta events carry no `response.id` either; the two responses arrive serialized (forkA fully completes before forkB starts).
**Expected (feature):** every event carries the originating `stream_id`; the two responses may interleave; the client can demultiplex both concurrent streams over the one connection.

franciscojavierarceo:
## Summary

- add typed `stream_id` routing to the Responses WebSocket API and echo it on every event, including local and execution errors
- run distinct stream lanes concurrently while preserving FIFO execution within the same stream and the backward-compatible default lane
- bound outstanding requests, aggregate input, event queues, upstream responses, gateway outputs, and MCP discovery/materialization so multiplexing cannot grow memory without limit
- document the WebSocket contract and add unit/integration coverage for interleaving, ordering, capacity, shutdown, cancellation, and resource limits

## Test Plan

- [x] `cargo test --workspace` (all runnable tests pass; PostgreSQL-only tests remain ignored without `TEST_POSTGRES_URL`)
- [x] `cargo build --workspace --bins`
- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- [x] `cargo fmt --all -- --check`
- [x] `git diff --check`
- [x] `uvx pre-commit run --all-files`
- [x] gstack review, Rust-specific review, adversarial review, and final testing/performance review report no remaining findings

Closes #239

### DIFF
--- a/crates/agentic-server-core/src/executor/gateway.rs
+++ b/crates/agentic-server-core/src/executor/gateway.rs
@@ -1,33 +1,54 @@
 use std::num::NonZeroUsize;
+use std::sync::Arc;
 use std::time::Duration;
 
 use futures::future::join_all;
-use tokio::sync::Semaphore;
+use tokio::sync::{OwnedSemaphorePermit, Semaphore};
 
 use crate::config::DEFAULT_MAX_CONCURRENT_GATEWAY_CALLS;
 use crate::events::SSEEventType;
 use crate::executor::error::{ExecutorError, ExecutorResult};
 use crate::executor::gateway_accumulator::{GatewayStreamAccumulator, StreamEvent, emit_sse_frame, synthetic_event};
 use crate::executor::request::RequestContext;
+use crate::executor::response_budget::ExecutorResponseBudget;
+use crate::tool::handler::MAX_GATEWAY_TOOL_OUTPUT_BYTES;
 use crate::tool::{GatewayBinding, ToolError, ToolOutput, ToolOwnership, ToolRegistry};
 use crate::types::io::output::{FunctionToolCall, GatewayCallStatus, McpCallStatus};
 use crate::types::io::{InputItem, OutputItem, ResponsesInput};
 use crate::types::request_response::ResponsePayload;
 use crate::utils::common::{serialize_to_string, serialize_to_value};
 
+pub(super) const MAX_CONCURRENT_MATERIALIZATIONS: usize = 16;
+
 /// Request-independent execution policy owned by one [`ExecutionContext`].
 ///
-/// The nonzero type makes `.buffered(0)` unrepresentable. Distinct execution
-/// contexts retain their own limits instead of sharing process-global state.
-#[derive(Debug, Clone, Copy)]
+/// The nonzero type prevents a zero-capacity per-request semaphore. Distinct
+/// execution contexts retain their own limits instead of sharing process-global state.
+#[derive(Debug, Clone)]
 pub(crate) struct GatewaySchedulerPolicy {
     max_concurrent_calls: NonZeroUsize,
+    materialization_permits: Arc<Semaphore>,
 }
 
 impl GatewaySchedulerPolicy {
     #[must_use]
-    pub(crate) const fn new(max_concurrent_calls: NonZeroUsize) -> Self {
-        Self { max_concurrent_calls }
+    pub(crate) fn new(max_concurrent_calls: NonZeroUsize) -> Self {
+        // Each permit protects one handler whose output is capped at
+        // MAX_GATEWAY_TOOL_OUTPUT_BYTES. The policy is cloned with an
+        // ExecutionContext, so this bound is shared by concurrent requests
+        // (including WebSocket lanes) instead of being recreated per turn.
+        Self {
+            max_concurrent_calls,
+            materialization_permits: Arc::new(Semaphore::new(MAX_CONCURRENT_MATERIALIZATIONS)),
+        }
+    }
+
+    /// Acquires one process-wide materialization slot shared by cloned execution contexts.
+    pub(crate) async fn acquire_materialization_permit(self) -> OwnedSemaphorePermit {
+        self.materialization_permits
+            .acquire_owned()
+            .await
+            .expect("materialization semaphore is never closed")
     }
 }
 
@@ -212,13 +233,21 @@ impl GatewayScheduler {
 
     /// Executes planned calls under the bounded-concurrency policy and records
     /// each tool's completed public lifecycle on the same slot.
+    #[cfg(test)]
     pub(super) async fn execute(&mut self) -> ExecutorResult<Vec<GatewayCallResult>> {
+        self.execute_with_budget(&ExecutorResponseBudget::new()).await
+    }
+
+    pub(super) async fn execute_with_budget(
+        &mut self,
+        response_budget: &ExecutorResponseBudget,
+    ) -> ExecutorResult<Vec<GatewayCallResult>> {
         let execution_slots = Semaphore::new(self.policy.max_concurrent_calls.get());
         let results = join_all(
             self.calls
                 .iter()
                 .cloned()
-                .map(|call| self.run_one(call, &execution_slots)),
+                .map(|call| self.run_one(call, &execution_slots, response_budget)),
         )
         .await
         .into_iter()
@@ -232,7 +261,12 @@ impl GatewayScheduler {
         Ok(results)
     }
 
-    async fn run_one(&self, plan: GatewayCallPlan, execution_slots: &Semaphore) -> ExecutorResult<GatewayCallResult> {
+    async fn run_one(
+        &self,
+        plan: GatewayCallPlan,
+        execution_slots: &Semaphore,
+        response_budget: &ExecutorResponseBudget,
+    ) -> ExecutorResult<GatewayCallResult> {
         let GatewayCallPlan {
             item_index,
             call,
@@ -244,6 +278,7 @@ impl GatewayScheduler {
                 &call,
                 &format!("gateway tool '{}' has no registered handler", call.name),
             )?;
+            response_budget.consume(output.output.len())?;
             return Ok(GatewayCallResult {
                 item_index,
                 input_item: InputItem::FunctionCallOutput(output.into()),
@@ -256,6 +291,7 @@ impl GatewayScheduler {
             None => None,
         };
         let _execution_slot = execution_slots.acquire().await.expect("semaphore is never closed");
+        let _materialization_permit = self.policy.clone().acquire_materialization_permit().await;
 
         let dispatched = if self.timeout.is_zero() {
             binding.execute(&call.call_id, &call.name, &call.arguments).await
@@ -273,6 +309,13 @@ impl GatewayScheduler {
                 ))),
             }
         };
+        match &dispatched {
+            Ok(output) => enforce_gateway_tool_output_size(output.output.len())?,
+            Err(ToolError::Execution(message) | ToolError::Config(message)) => {
+                enforce_gateway_tool_output_size(message.len())?;
+            }
+            Err(ToolError::MissingOutput { .. }) => {}
+        }
         let (output, status) = match dispatched {
             Ok(output) => (output, GatewayCallStatus::Completed),
             Err(ToolError::Execution(message) | ToolError::Config(message)) => {
@@ -280,6 +323,7 @@ impl GatewayScheduler {
             }
             Err(error @ ToolError::MissingOutput { .. }) => return Err(ExecutorError::from(error)),
         };
+        response_budget.consume(output.output.len())?;
         let public_output = binding.public_output(&call, &output, status);
         Ok(GatewayCallResult {
             item_index,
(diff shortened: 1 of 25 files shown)


crates/agentic-server-core/src/executor/gateway.rs
```diff
@@ -273,13 +309,21 @@ impl GatewayScheduler {
                 ))),
             }
         };
+        match &dispatched {
+            Ok(output) => enforce_gateway_tool_output_size(output.output.len())?,
+            Err(ToolError::Execution(message) | ToolError::Config(message)) => {
+                enforce_gateway_tool_output_size(message.len())?;
+            }
+            Err(ToolError::MissingOutput { .. }) => {}
+        }
```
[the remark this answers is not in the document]
franciscojavierarceo: Fixed in 6f1b1da. The bound-handler path now materializes either the success or JSON-wrapped error output first, then enforces the cap on `output.output.len()`. I added a regression test with escapable error text whose raw message fits the cap but whose materialized output exceeds it; it now reports the gateway output limit rather than the aggregate response-budget error.

crates/agentic-server/src/handler/websocket/responses.rs
```diff
@@ -192,27 +508,32 @@ where
...
+        stream_id: None,
+    })?;
+    let stream_id = value
+        .get("stream_id")
+        .map(|value| serde_json::from_value::<StreamId>(value.clone()))
+        .transpose()
+        .map_err(|error| WsRequestParseError {
+            error: WsError::from(ExecutorError::from(error)),
+            stream_id: None,
+        })?;
```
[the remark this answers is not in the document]
franciscojavierarceo: Addressed in 6f1b1da. `parse_ws_request` now extracts `stream_id` with `Value::as_str` and validates via `StreamId::try_from(&str)`, avoiding the cloned `Value` and `serde_json::from_value` path. Non-string IDs remain client-visible 400 validation errors.

crates/agentic-server-core/src/executor/gateway.rs
```diff
@@ -244,6 +278,7 @@ impl GatewayScheduler {
                 &call,
                 &format!("gateway tool '{}' has no registered handler", call.name),
             )?;
+            response_budget.consume(output.output.len())?;
             return Ok(GatewayCallResult {
```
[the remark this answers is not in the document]
franciscojavierarceo: Fixed in 6f1b1da. The missing-handler branch now enforces `MAX_GATEWAY_TOOL_OUTPUT_BYTES` against the final materialized `ToolOutput.output` before charging the shared response budget, so it preserves the gateway-specific size error semantics.

crates/agentic-server-core/src/tool/web_search.rs
```diff
@@ -207,31 +208,42 @@ impl WebSearchHandler {
...
-            if let Some(results) = response.results.get("web").and_then(Value::as_array) {
-                web.extend(results.iter().cloned());
+        let mut accumulated_bytes = 0usize;
+        while let Some(mut response) = responses.try_next().await? {
+            let response_bytes = serde_json::to_vec(&response.results)
+                .and_then(|results| {
+                    serde_json::to_vec(&response.metadata).map(|metadata| results.len() + metadata.len())
+                })
+                .map_err(|error| ToolError::Execution(format!("failed to size web_search output: {error}")))?;
+            accumulated_bytes = accumulated_bytes.saturating_add(response_bytes);
```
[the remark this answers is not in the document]
franciscojavierarceo: Addressed in 6f1b1da. Per-response sizing now serializes through a counting `io::Write`, retaining the early aggregate bound without allocating temporary `Vec<u8>` buffers. The exact final-output length check remains in place.

crates/agentic-server-core/src/executor/gateway.rs
```diff
@@ -1,33 +1,54 @@
...
+        }
+    }
+
+    /// Acquires one process-wide materialization slot shared by cloned execution contexts.
+    pub(crate) async fn acquire_materialization_permit(self) -> OwnedSemaphorePermit {
+        self.materialization_permits
+            .acquire_owned()
+            .await
+            .expect("materialization semaphore is never closed")
     }
```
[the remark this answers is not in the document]
franciscojavierarceo: Addressed in 6f1b1da. `acquire_materialization_permit` now takes `&self` and returns an owned, non-borrowing future backed by the cloned semaphore `Arc`; call sites no longer clone `GatewaySchedulerPolicy`, including the registry-builder closure.

crates/agentic-server/src/handler/websocket/responses.rs
```diff
@@ -25,7 +28,249 @@ use crate::app::AppState;
...
+        self.lanes.insert(lane.clone(), VecDeque::new());
+        self.spawn(lane, work);
+        Ok(())
+    }
+
+    fn schedule_next(&mut self, lane: Option<StreamId>) {
+        let next = self.lanes.get_mut(&lane).and_then(VecDeque::pop_front);
+        if let Some(work) = next {
+            self.queued_requests -= 1;
+            self.spawn(lane, work);
```
maralbahari: Please recheck the authenticated principal before starting a queued request. The expiry check now runs only when a text frame arrives, and `schedule_next()` starts previously accepted work without checking it again. If a client queues two requests on the same lane before expiry and the first finishes after expiry (including clock skew), the second still reaches inference and can execute tools. This also affects the default lane: `main` checked expiry after dequeuing each request. Preserve that check at dispatch and return the tagged `invalid_token` event instead of starting expired queued work.

crates/agentic-server/src/handler/websocket/responses.rs
```diff
@@ -424,32 +685,68 @@ fn sse_json_data_lines(chunk: &str) -> impl Iterator<Item = &str> {
...
+}
+
+async fn queue_ws_json(
+    outbound_tx: &mpsc::Sender<Value>,
+    value: Value,
+    stream_id: Option<&StreamId>,
+) -> Result<(), WsError> {
+    outbound_tx
+        .send(attach_stream_id(value, stream_id)?)
+        .await
```
maralbahari: need to enforce an outbound event byte limit for the local completion path too?. `generate: false` bypasses `checked_stream_event()`: `complete_without_inference()` copies the rehydrated `instructions` into both lifecycle events and puts them directly into this channel. A valid request with 2 MiB of instructions produced a **2,097,502-byte** `response.created` event, exceeding the new 1 MiB limit. Small continuation requests can inherit those instructions, so the 12 MiB incoming-request budget does not bound the bytes retained by the 64-entry outbound queue. Validate local events before persistence/enqueue, or use a size-checked outbound message type shared by every producer

crates/agentic-server/src/handler/websocket/responses.rs
```diff
@@ -25,7 +28,249 @@ use crate::app::AppState;
...
+            debug!(stream_id = ?lane, queued_requests = queue.len(), "queued websocket request on active lane");
+            return Ok(());
+        }
+
+        self.lanes.insert(lane.clone(), VecDeque::new());
+        self.spawn(lane, work);
+        Ok(())
+    }
+
+    fn schedule_next(&mut self, lane: Option<StreamId>) {
```
maralbahari, line 218: Please recheck the authenticated principal before starting a queued request. The expiry check now runs only when a text frame arrives, and `schedule_next()` starts previously accepted work without checking it again. If a client queues two requests on the same lane before expiry and the first finishes after expiry (including clock skew), the second still reaches inference and can execute tools. This also affects the default lane: `main` checked expiry after dequeuing each request. Preserve that check at dispatch and return the tagged `invalid_token` event instead of starting expired queued work.

crates/agentic-server/src/handler/websocket/responses.rs
```diff
@@ -424,32 +685,68 @@ fn sse_json_data_lines(chunk: &str) -> impl Iterator<Item = &str> {
...
+    }
+    Ok(value)
+}
+
+async fn queue_ws_json(
+    outbound_tx: &mpsc::Sender<Value>,
+    value: Value,
+    stream_id: Option<&StreamId>,
+) -> Result<(), WsError> {
+    outbound_tx
```
maralbahari, line 755: would need to enforce an outbound event byte limit for the local completion path too?! `generate: false` bypasses `checked_stream_event(): complete_without_inference() `copies the rehydrated instructions into both lifecycle events and puts them directly into this channel. A valid request with 2 MiB of instructions produced a 2,097,502-byte response.created event, exceeding the new 1 MiB limit. Small continuation requests can inherit those instructions, so the 12 MiB incoming-request budget does not bound the bytes retained by the 64-entry outbound queue. Validate local events before persistence/enqueue, or use a size-checked outbound message type shared by every producer

crates/agentic-server/src/handler/websocket/responses.rs
franciscojavierarceo: fixed in 93e554c. the request task now rechecks the authenticated principal at dispatch, so work dequeued by `schedule_next()` cannot start after expiry. rejected work gets the `invalid_token` event with its `stream_id` when present. regression tests cover named and default lanes, including clock skew, and verify that the queued request never reaches inference.

crates/agentic-server/src/handler/websocket/responses.rs
franciscojavierarceo: fixed in 93e554c. the outbound queue now accepts only `WsOutboundEvent`, which checks the serialized JSON size, including `stream_id`, against the 1 MiB limit. local completion validates both lifecycle events before persistence or enqueueing either one. tests cover the 2 MiB instructions case, a created event that fits while completion exceeds the limit, and routing/JSON-escaping overhead. after merging `main` in 3751c19, all 948 Rust tests pass locally, with 9 existing ignored tests; Clippy and pre-commit hooks pass too.

maralbahari (approved):

PR #249, opened 2026-09-06
feat: add per-request context to the /v1/messages loop (#137)

### LINKED ISSUE #137: /v1/messages loop: per-request context (typed `MessagesRequest` + raw `Value`)
Follow-up to #131 (raised in review) — the Messages loop currently operates on a raw `serde_json::Value` for lossless upstream forwarding, and separately parses `MessagesRequest` for typed field access. This proposes a single per-request context holding both, so the loop gets safe typed access without a second parse and without giving up lossless forwarding.

### Why raw `Value` is used today

The loop is a pass-through: it forwards the client's request to vLLM `/v1/messages` **verbatim** and only reads `tools` + rewrites `messages`/`stream`. Unlike `RequestPayload::to_upstream_request` (which *transforms* — flattens Codex namespaces, normalizes tool types, resolves `tool_choice`), there's no transform step here, so a typed round-trip would be pure overhead. One real catch: `ContentBlock` has `#[serde(other)] Unknown`, so deserializing→reserializing through `MessagesRequest` would silently drop unmodeled block types (`redacted_thinking`, future blocks, provider extensions) — raw `Value` forwards those untouched.

### Proposed

A `MessagesRequestContext { typed: MessagesRequest, raw: Value }` built once per request:
- typed view for safe field access (tools, stream, model) in routing + the loop (today the handler parses `MessagesRequest` only for routing and drops it; the loop mutates the raw `Value` alone);
- raw `Value` remains the forwarded-upstream body (lossless);
- the two views only diverge where the loop touches the raw body — forcing `stream:false` and appending the assistant + `tool_result` round to `messages` (`append_round_to_history`). The context would own those mutations so the views can't drift.

Alternatively, evaluate a `to_upstream`-style method on `MessagesRequest` that re-serializes losslessly — but note the loss is at the **`ContentBlock`** level (`#[serde(other)] Unknown` is a unit variant that captures no data), not top-level, so the top-level `extra` flatten doesn't cover it. This needs a block-level capture variant (or custom (de)serialize) on `ContentBlock`, and serde won't combine `#[serde(other)]` with a data-capturing catch-all on an internally-tagged enum — so it's not a drop-in. Confirm `redacted_thinking`/unknown blocks round-trip before adopting.

Scoped as a follow-up so it doesn't reshape the routing/loop plumbing inside #131. Ref: #131 review thread on `messages_loop.rs`.

Zheng-Lu:
## Summary

Closes #137. Follow-up to #131, addressing the code review concern that the Messages loop took a bare `serde_json::Value` at a public API boundary while the handler separately parsed a `MessagesRequest` for typed field access.

## Technical Details

Introduces `MessagesRequestContext { typed: MessagesRequest, raw: Value }`, built once per request, and threads it through `run_messages_loop` and `run_messages_stream` in place of the untyped body.

The raw body stays the thing forwarded upstream, and is deliberately not re-serialized from the typed view. `ContentBlock` catches unmodeled block types in `#[serde(other)] Unknown` and several variants model only the fields the gateway reads, so a typed round-trip would drop `cache_control` and `is_error` and collapse `image`/`redacted_thinking` into a literal `{"type":"unknown"}` block.

- The typed view exposes only `tools()`, `stream()` and `model()` — the fields the loops read and never mutate. `messages` and `system` are unreachable through it, so a stale typed view cannot be read back after a round is appended.
- The context owns every mutation to the upstream body (`force_stream`, `append_round`) plus the native web-search budget, replacing the two duplicated `append_round_to_history` copies with one method.
- The two views diverge only where the gateway rewrites for upstream: `normalize_native_web_search` rewrites the native `web_search_20250305` declaration in the raw body, while the typed view keeps the client's original so the tool seam can still classify it.
- `MessagesRequestContext::new` reuses the handler's routing parse, so neither view is built twice and a proxied request never pays for a raw parse it does not use. Native web-search validation folds into the constructor, still ahead of a streaming response committing its status, and `validate_native_web_search_request` drops off the public API.

Also removes the single-field `ResolvedCall`/`ResolvedStreamCall` wrappers, which every caller immediately unwrapped, and a redundant deep clone of the assistant turn per round in the non-streaming loop.

Note this keeps routing on the full `MessagesRequest` parse, so a body that declares a gateway tool but fails that parse on an unrelated field still falls through to the proxy silently. That behaviour is unchanged here and is worth a separate issue.

## Test Plan
- `cargo fmt -- --check`, `cargo clippy --all-targets -- -D warnings` and `cargo test --workspace` all clean: 889 passed, 0 failed, cassette replays included.
- New `messages_loop_preserves_unmodeled_blocks_and_tool_result_fields_across_rounds` covers `image`, `redacted_thinking`, `citations` and a `tool_result` carrying `is_error` and `cache_control`, asserting they reach upstream unchanged across a gateway round.
- Negative control: pointing `upstream_body` at the typed view fails 8 tests, including the existing Claude Code cache_control replays, which confirms the guard is not vacuous.


franciscojavierarceo (commented):
the handler builds both views from the same request, and the loops keep forwarding the raw body when appending tool rounds. i didn't find a behavior regression in the refactor. this was a source review; i haven't independently run the tests.

crates/agentic-server-core/src/executor/messages_stream.rs
```diff
@@ -446,17 +449,19 @@ fn executor_error_sse(error: &ExecutorError) -> String {
...
+/// Returns one `tool_result` block per call, fed back next round. (The assistant
+/// turn — including each call's `tool_use` block — is reconstructed from the
+/// accumulator's buffered blocks in [`MessagesStreamAccumulator::take_round`].)
 async fn execute_gateway_calls(
     calls: &[StreamedCall],
     registry: &ToolRegistry,
     gateway_map: &tool_seam::GatewayToolMap,
     allowed_searches: usize,
-) -> Vec<ResolvedStreamCall> {
+) -> Vec<Value> {
```
maralbahari: Can we return `Vec<GatewayToolResult>` with typed `tool_use_id`, `content`, and `is_error` fields? These blocks are generated by the gateway and have a known schema, so `Vec<Value>` loses compile-time guarantees unnecessarily. Please carry that type through both execution helpers and `append_round`, converting to JSON when assembling the upstream body. The existing `ContentBlock::ToolResult` would need extending before reuse because it lacks `is_error`.

crates/agentic-server-core/src/executor/messages_context.rs
```diff
@@ -0,0 +1,266 @@
...
+use crate::types::messages::{MessagesRequest, ToolParam};
+use crate::utils::common::serialize_to_string;
+
+/// One `/v1/messages` request, in both the typed and raw views the gateway tool
+/// loops need. See the module docs for why both exist.
+#[derive(Debug)]
+pub struct MessagesRequestContext {
+    /// The client's request as received. Read-only: routing and registry
+    /// construction only.
+    typed: MessagesRequest,
```
maralbahari: Can we move only `tools`, `stream`, and `model` into the context’s typed state? Keeping the entire `MessagesRequest` also retains its owned message history and system prompt, although those fields are deliberately inaccessible. Previously the streaming handler dropped that parsed request after returning the response; now `ctx` carries this duplicate prompt data throughout the stream. For long prompts and concurrent streams, this increases retained memory without providing additional functionality.

crates/agentic-server-core/src/executor/messages_context.rs
```diff
@@ -0,0 +1,266 @@
...
+    /// a streaming handler commits its HTTP status — an invalid declaration must
+    /// surface as an error response, not as a mid-stream event.
+    ///
+    /// # Errors
+    /// Returns [`ExecutorError::JsonError`] if `body` is not valid JSON, or
+    /// [`ExecutorError::InvalidRequest`] if it carries an unsupported or invalid
+    /// native web-search declaration.
+    pub fn new(typed: MessagesRequest, body: &[u8]) -> ExecutorResult<Self> {
+        let raw = serde_json::from_slice(body).map_err(ExecutorError::JsonError)?;
+        Self::from_parts(typed, raw)
```
maralbahari: Can construction guarantee that the typed and raw views come from the same input? `new` accepts them independently and only checks that `body` parses as JSON. A valid `MessagesRequest` paired with `b"[]"` currently returns `Ok`, and mismatched model/stream values are accepted too. The current HTTP handler pairs them correctly, but the public constructor does not enforce the invariant the loops rely on. Deriving both views internally, or accepting a validated paired input, would make that guarantee enforceable.

crates/agentic-server-core/src/executor/messages_stream.rs
Zheng-Lu: Agreed. I added a dedicated `GatewayToolResult` type with typed `tool_use_id`, `content`, and `is_error` fields plus a fixed `tool_result` discriminator. Both streaming and non-streaming execution helpers now return `Vec<GatewayToolResult>`, and `append_round` carries that type until it assembles the raw upstream body.

I used a dedicated type instead of extending `ContentBlock::ToolResult` because gateway-generated blocks have a narrower schema: textual content and a required `is_error` value.

crates/agentic-server-core/src/executor/messages_context.rs
Zheng-Lu: Yes, I agreed. `MessagesRequestContext` now retains only a compact typed state containing `model`, `tools`, and `stream`. The owned message history, system prompt, and other parsed fields are dropped during context construction, before the tool loop or response stream begins. The raw JSON remains the sole source of truth for the upstream body and accumulated message history.

crates/agentic-server-core/src/executor/messages_context.rs
Zheng-Lu: Agreed. I replaced the independently supplied typed request and body with `ParsedMessagesRequest<'a>`. Its private fields can only be created by parsing the exact immutable body bytes it retains. The handler uses that typed view for routing and then consumes the validated pair to construct `MessagesRequestContext`.

This makes mismatched typed/raw inputs unrepresentable through the public API while preserving the proxy fast path: the raw `Value` is only parsed for requests routed into the gateway tool loop.

maralbahari (approved):

PR #259, opened 2026-09-06
fix: exclude image bytes from compaction token estimates (#255)

### LINKED ISSUE #255: Exclude inline image bytes from Responses compaction token estimates
Parent: #251

## Problem

`estimate_input_tokens` serializes model-facing input and divides its byte count by four. Inline image data is therefore counted as text: a 4 MiB base64 string contributes roughly one million estimated tokens. This can trigger premature automatic compaction.

## Scope

- Replace whole-JSON byte counting with a typed traversal of model-facing input that estimates textual content and structural overhead separately from image content.
- Cover image content in user messages and structured function/custom tool call outputs. Preserve accounting for textual tool outputs, function arguments, reasoning, and the effective compacted window.
- Use an explicit, documented image-budget heuristic or configurable allowance; do not treat base64 length as vision tokens and do not count images as free. Exact image tokens depend on the model/processor and are outside this issue.
- Avoid serializing/copying large image payloads solely to estimate their token cost.
- Preserve compaction's retained image-bearing messages and existing text-only semantics.

## Acceptance criteria

- [ ] Larger inline encoding alone does not proportionally increase estimated text tokens.
- [ ] Additional images contribute a nonzero documented budget; text growth still increases estimates.
- [ ] Image-bearing messages and structured tool outputs follow the same estimation policy.
- [ ] Threshold tests show that base64 payload size alone no longer triggers compaction, while sufficiently large textual context still does.
- [ ] Existing compacted-window behavior and retained image content remain covered.
- [ ] Approximation limits and any configuration are documented.

## Starting points

[crates/agentic-server-core/src/executor/compaction.rs](https://github.com/vllm-project/agentic-api/blob/c04ac196f44133c37be652c4ee3f777db32ae6df/crates/agentic-server-core/src/executor/compaction.rs)

Can ship independently; required before claiming automatic-compaction support for image workflows. No model-specific tokenizer or image decoder is required by this issue.

Zheng-Lu:
## Summary

Fixes #255.

Replace whole-input JSON serialization in Responses compaction token estimation with a borrowed, typed traversal of the effective model-facing input.

The new estimator:

- Counts textual fields using the existing four UTF-8 bytes-per-token approximation.
- Adds deterministic structural allowances for Responses items and content parts.
- Assigns each image a documented fixed allowance of 1,024 tokens.
- Excludes image URLs, file IDs, and inline base64 bytes from textual token estimates.
- Covers user messages, function and custom tool call outputs, function arguments, reasoning content, and compacted windows.
- Reuses a borrowed model-item iterator so estimation does not clone large image payloads.
- Preserves retained image-bearing user messages during compaction.

The Responses compaction guide now documents the heuristic and its model-specific limitations.

## Test Plan

- `cargo test`
- `cargo test -p agentic-server-core --test compaction_cassette_test`

Added coverage verifies:

- 100 KiB and 5 MiB inline encodings produce the same estimate.
- Each additional image adds exactly 1,024 estimated tokens.
- User messages and structured `view_image` tool call outputs use the same image policy.
- Text growth across messages, tool calls, tool call outputs, and reasoning increases estimates.
- Large image payloads alone do not trigger automatic compaction.
- Genuine long text still triggers compaction.
- Retained image content survives compaction unchanged.
- Text-only estimates remain comparable to the previous JSON-size heuristic.


franciscojavierarceo (commented):
checked the estimator across message images, structured function/custom tool outputs, text, reasoning, and the effective compaction window. image payload size no longer drives the estimate, and each image still contributes a fixed allowance. no additional changes requested from the source review; i haven't rerun the tests locally.

franciscojavierarceo (approved):

Issue #251, opened 2026-09-06
Support Codex CLI image attachments with vLLM


franciscojavierarceo:
## Outcome

Codex CLI users can attach local images and use `view_image` through Agentic API with a supported vision-capable vLLM model. Images must reach inference and survive Responses continuation without being silently removed or treated as base64 text for context budgeting.

Related: #54 (broader Codex integration). This tracker scopes attachment handling only.

## Findings

Source investigation used OpenAI Codex main at `6af345407d9c2a568da9d01b6c4b81a9e61495c0` and Agentic API at `c04ac196f44133c37be652c4ee3f777db32ae6df`.

- Codex converts local image attachments to inline data URLs in `input_image`; `view_image` returns structured image content in a function call output. Ordinary file mentions identify local paths for client-executed tools.
- The gateway already models images in messages and tool call outputs, but the launcher hardcodes text-only model metadata and the HTTP catalog relies on upstream capability metadata that may be absent.
- HTTP and WebSocket input limits are fixed at 10 MiB. Automatic compaction estimates tokens from serialized JSON, including base64 payloads.
- User-message `input_file` falls through to `Unknown`; this is a separate validation gap, not a requirement for the ordinary CLI file-mention workflow.

## Scope and sequencing

**Initial CLI image milestone:** resolve capabilities consistently across catalog and launcher paths, then prove attachment and `view_image` behavior with deterministic coverage and a pinned live vLLM/model check. Existing request limits remain applicable until the limits sub-issue lands.

**Hardening:** configurable transport limits and media-aware compaction estimation. Compaction support is not complete until that sub-issue lands.

**Related API correctness:** explicitly reject unsupported file content on typed execution paths without changing transparent proxy semantics. This can ship independently and does not block the initial CLI image milestone.

Sub-issues will define implementation boundaries and acceptance criteria.

## Explicitly out of scope

- A Codex fork or changes to Codex attachment serialization.
- Files API endpoints, multipart uploads, blob storage, file-ID resolution, PDF/DOCX ingestion, OCR, or document rendering services.
- Audio/video, image generation, original-detail guarantees, or a generic model-capability platform.
- Assuming vision capability from a model name or claiming all vLLM model/template combinations work.

## Completion

All linked sub-issues are complete; docs identify the tested Codex version, vLLM version, model/template, image limits, and unsupported file behavior. A pinned live test must establish image understanding; replay tests alone establish transport behavior, not model capability.

## Work breakdown

[...]

PR #260, opened 2026-09-06
feat: make max request body size configurable (#254)

### LINKED ISSUE #254: Make Responses HTTP and WebSocket request limits configurable
Parent: #251

## Problem

The gateway uses a fixed 10 MiB limit for HTTP bodies and WebSocket messages/frames. Inline base64 adds roughly one-third overhead, and replayed history can contain multiple images. Valid image workflows can exceed the limit before inference.

## Scope

- Add one validated positive request-size setting through existing server configuration and CLI/environment precedence conventions; retain 10 MiB as the default.
- Resolve the setting once and use it consistently for Responses HTTP, `/v1/responses/compact`, and Responses WebSocket messages/frames. Audit other consumers of the shared body reader and preserve their defaults deliberately.
- Forward relevant options when `agentic run` starts the server; an attached harness uses the running server's policy.
- Preserve HTTP 413 behavior and documented WebSocket transport-level rejection. Do not promise a JSON error for a frame rejected before parsing.
- Document that the limit covers serialized request bytes, including history and base64, and is distinct from upstream image-count and context limits.

## Acceptance criteria

- [ ] Default behavior is unchanged.
- [ ] Zero, invalid, and out-of-range values fail configuration validation.
- [ ] Small configurable limits allow deterministic boundary tests for HTTP and WebSocket transports; over-limit requests never reach upstream.
- [ ] Increasing the configured limit permits an otherwise valid request above the previous default.
- [ ] CLI/config/environment precedence and launcher forwarding are tested and documented.

## Starting points

- [crates/agentic-server/src/handler/common.rs](https://github.com/vllm-project/agentic-api/blob/c04ac196f44133c37be652c4ee3f777db32ae6df/crates/agentic-server/src/handler/common.rs)
- [crates/agentic-server/src/handler/websocket/responses.rs](https://github.com/vllm-project/agentic-api/blob/c04ac196f44133c37be652c4ee3f777db32ae6df/crates/agentic-server/src/handler/websocket/responses.rs)
- [crates/agentic-server/src/app.rs](https://github.com/vllm-project/agentic-api/blob/c04ac196f44133c37be652c4ee3f777db32ae6df/crates/agentic-server/src/app.rs)
- [crates/agentic-server/src/main.rs](https://github.com/vllm-project/agentic-api/blob/c04ac196f44133c37be652c4ee3f777db32ae6df/crates/agentic-server/src/main.rs)

Can ship independently. Does not add image resizing, blob storage, or per-user quotas.

Zheng-Lu:
## Summary

The gateway hardcoded a 10 MiB ceiling for HTTP request bodies and WebSocket messages (`MAX_BODY_SIZE`). Inline base64 image attachments cost roughly a third more than the source image, and replayed conversation history can carry several of them, so valid image workflows exceeded the limit before reaching inference. Closes #254 (parent: #251).

This introduces a validated positive `max_request_body_size`, resolved as **CLI argument > environment variable > configuration file > 10 MiB default**:

| Surface     | Name                                                                     |
| ----------- | ------------------------------------------------------------------------ |
| CLI         | `--max-request-body-size-bytes` (global; valid before and after `serve`) |
| Environment | `AGENTIC_MAX_REQUEST_BODY_SIZE_BYTES`                                    |
| Config file | `[server] max_request_body_size_bytes`                                   |
| Default     | `DEFAULT_MAX_REQUEST_BODY_SIZE` = 10485760                               |

Design notes:

- **`NonZeroUsize` throughout**, so zero and negative values are unrepresentable and rejected at parse time by both serde and clap. This mirrors the existing `tools.max_concurrent_gateway_calls` setting rather than inventing a new pattern.
- **An explicit CLI argument short-circuits the lower-priority sources**, so a stale or malformed `AGENTIC_MAX_REQUEST_BODY_SIZE_BYTES` inherited from the environment cannot block startup when the operator names a valid value.
- **The resolved limit is stored once in `AppState`** and applied uniformly to every request-bearing endpoint and to WebSocket messages and frames. `/v1/messages` carries the same inline attachments as `/v1/responses`; `/v1/conversations` bodies are small enough that a raised ceiling has no effect on them.
- **Transport semantics are unchanged.** Oversized HTTP requests still receive `413 Payload Too Large` with `body_too_large`; oversized WebSocket messages and frames are still rejected by the transport, closing the connection before any JSON is parsed — no JSON error event is promised for frames rejected pre-parse.
- **Transport settings travel in a new `GatewayOptions` struct** rather than in core's `Config`, keeping transport concerns out of core per `AGENTS.md`.
- `agentic run` forwards the flag only when set, so the server's own precedence chain still applies when it is omitted. The Python launcher needed no change — it already passes the process environment through to the binary.

Documentation in `README.md` and `docs/deploying/container.md` states that the limit covers **serialized request bytes** — JSON overhead, replayed history, and base64 payloads — and is unrelated to the upstream model's token context limit, so raising it does not raise what the inference server will accept.

`crates/agentic-llm-d` is deliberately out of scope; its separate `MAX_HYDRATE_BODY_SIZE` / `MAX_PERSIST_BODY_SIZE` constants are unchanged.

## Test Plan

Eleven new tests, plus one rewritten to drop a 10 MiB allocation.

**Configuration (`config_file.rs`, `main.rs`, `agentic_process.rs`)**

- `rejects_zero_max_request_body_size` — `[server] max_request_body_size_bytes = 0` fails to load, and the error names the key.
- `accepts_positive_max_request_body_size` — a positive value parses.
- The generated-config round-trip test now asserts `[server]` and the byte value are written back.
- `max_request_body_size_is_configurable_from_cli_env_and_file` — the full precedence matrix: default, file over default, environment over file, CLI over environment.
- `max_request_body_size_rejects_invalid_environment_overrides` — `0`, `-1`, and `not-a-number` are rejected with no CLI override; with one, the CLI value wins instead of the malformed environment value blocking startup.
- `max_request_body_size_argument_is_global` — the flag parses before and after the `serve` subcommand, and `0` is rejected by clap.
- `request_size_limit_is_forwarded_only_when_configured` — `server_args` emits the flag pair when set and omits it otherwise.

**HTTP (`responses_test.rs`, `conversations_test.rs`)** — boundary tests use a 500-byte limit so they are deterministic and allocate nothing large.

- `/v1/responses` over the limit → `413` with `body_too_large`.
- `/v1/responses/compact` over the limit → `413` (covers the `read_json` path).
- `/v1/conversations` over the limit → `413` (locks in the uniform-policy decision).
- `test_raised_request_size_limit_admits_larger_body` — a 4 KiB body under a 64 KiB limit reaches the mock upstream with its payload intact, proving raising the ceiling works rather than only lowering it.

**WebSocket (`responses_websocket_test.rs`)**

- `test_websocket_rejects_messages_larger_than_the_configured_request_size_limit` — rewritten against the 500-byte limit; asserts transport error-or-`Close` and that no upstream request was made.
- `test_websocket_accepts_messages_within_a_raised_request_size_limit` — a payload that the small limit would close streams to completion under a 64 KiB limit, with an assertion that the payload does exceed the small limit so the ceiling is demonstrably what changed.

**Commands**

- `cargo clippy --all-targets -- -D warnings` — clean
- `cargo test --workspace` — **891 passed, 0 failed** (3 pre-existing ignored)


franciscojavierarceo (commented):
checked the CLI > environment > file precedence and the shared limit across HTTP body readers and WebSocket frames/messages. all 12 focused configuration, forwarding, and request-size tests pass locally, including rejecting oversized requests and admitting requests under a raised limit. no changes requested.

franciscojavierarceo (approved):

PR #261, opened 2026-09-06
fix: resolve image capabilities consistently for Codex catalogs and launchers (#252)

### LINKED ISSUE #252: Resolve image capabilities consistently for Codex catalogs and launchers
Parent: #251

## Problem

The HTTP Codex catalog advertises image input only when upstream `/v1/models` includes `capabilities: ["image"]`. Missing metadata becomes text-only. Separately, `prepare_codex_home` always writes `input_modalities: ["text"]`. Codex strips image content when its model metadata excludes images, so a vision-capable upstream can receive only a placeholder.

## Scope

- Add typed per-model input-modality overrides to the existing configuration, keyed by the served model ID. Suggested TOML: `[models."served-model-id"]` with `input_modalities = ["text", "image"]`.
- Resolve explicit configuration before recognized upstream metadata; retain a conservative text-only fallback. Do not guess from model names. Validate unsupported modality values at startup.
- Expose resolved modalities in the Codex `/v1/models?client_version=...` response; preserve ordinary `/v1/models` pass-through behavior.
- Make both `agentic run codex` and `agentic harness codex` obtain resolved capabilities from the gateway before writing their isolated catalog. Reuse the capability contract without blindly replacing launcher-specific tool settings with the HTTP template.
- Keep `supports_image_detail_original` false. Document configuration, metadata-fetch failure behavior, and the need to regenerate existing isolated catalogs.
- Use typed Rust contracts for new public boundaries rather than adding untyped JSON APIs.

## Acceptance criteria

- [ ] A configured vision model advertises text and image even if upstream has no capability metadata.
- [ ] An explicit text-only override wins over upstream image metadata; an unknown model without metadata remains text-only.
- [ ] Both launcher modes write the same resolved input modalities as the HTTP catalog.
- [ ] Model selection, authentication, launcher tool settings, and ordinary models pass-through remain covered by regression tests.
- [ ] Missing selected model and catalog-fetch failures produce actionable errors rather than silently writing incorrect metadata.
- [ ] Configuration and launcher behavior are documented.

## Starting points

- [crates/agentic-server/src/config_file.rs](https://github.com/vllm-project/agentic-api/blob/c04ac196f44133c37be652c4ee3f777db32ae6df/crates/agentic-server/src/config_file.rs)
- [crates/agentic-server/src/handler/http/models.rs](https://github.com/vllm-project/agentic-api/blob/c04ac196f44133c37be652c4ee3f777db32ae6df/crates/agentic-server/src/handler/http/models.rs)
- [crates/agentic-server/src/agentic_harness.rs](https://github.com/vllm-project/agentic-api/blob/c04ac196f44133c37be652c4ee3f777db32ae6df/crates/agentic-server/src/agentic_harness.rs)
- [crates/agentic-server/src/agentic_process.rs](https://github.com/vllm-project/agentic-api/blob/c04ac196f44133c37be652c4ee3f777db32ae6df/crates/agentic-server/src/agentic_process.rs)

Independent implementation; image CLI acceptance coverage depends on this issue.

Zheng-Lu:
## Summary

Fixes #252 (parent #251). Codex previously received a text-only isolated model catalog from the launcher even when the gateway advertised image support, causing the client to remove image attachments. Resolve input modalities once and use the gateway catalog for both `agentic harness codex` and `agentic run codex`.

Resolution uses an explicit `[models."served-model-id"] input_modalities = ["text", "image"]` override, then recognized upstream metadata, then a conservative text-only fallback. Explicit text-only overrides win; model names never imply capabilities. Typed configuration rejects unknown, duplicate, empty, and image-only modality lists. `supports_image_detail_original` remains false.

Model selection and modalities come from one bounded, authenticated catalog snapshot using the installed Codex client version. Catalog discovery preserves readiness deadlines and fails clearly for incompatible catalogs. Private catalog resolution is separated from process lifecycle management; existing tool settings are preserved.

Merge current main, including the Responses image-preservation changes. Extend the existing CI smoke with actual pinned Codex 0.149.1: attach the committed PNG through both launcher modes and require byte-for-byte equality in the upstream capture. A text-only control must send no image. Replay uses an existing recorded gateway/vLLM response; it verifies client/catalog/transport behavior, not fresh vision inference. No cassette contents were fabricated or modified.

## Test Plan

- Full `agentic-server` test suite: 439 distinct tests passed, no failures or ignored tests.
- Complete `scripts/codex-smoke.sh`: HELLO baseline, attached-image, integrated-image, and text-only control passed using Codex 0.149.1 and rebuilt binaries.
- `scripts/tests/agentic-launchers-test.sh` passed.
- Python replay/smoke regression suite: 20 passed, including exact image validation and timeout descendant cleanup.
- Strict Clippy and all pre-commit hooks passed; production file-size baseline decreased.
- Independent source review passed after resolving main conflicts.


franciscojavierarceo (commented):
checked the capability precedence, catalog parsing, and both launcher paths; no new findings from the source review. we should fold in the replay-server `/v1/models` fix already shared by LOGO127 and rerun `scripts/codex-smoke.sh` before merging, since the current launcher now requires that discovery request. i haven't rerun the pinned Codex smoke locally.

franciscojavierarceo:
the replay catalog fix is included, and the capability, launcher, model-handler, and replay tests pass locally. this branch still conflicts with `main`, and `scripts/codex-smoke.sh` only exercises the text-only `HELLO` turn. once #263 lands, we should rebase this PR and add the agreed Codex smoke that attaches a real image and verifies its bytes reach the upstream capture before enabling image modality in the harness.

franciscojavierarceo (approved):
the branch is updated with main, and the pinned Codex 0.149.1 smoke now verifies exact image bytes upstream through both launcher modes, with a text-only negative control. the full server suite, launcher tests, Clippy, and pre-commit pass locally; independent review found no remaining blockers. hosted CI is still running.

PR #262, opened 2026-09-06
fix: preserve upstream error headers on non-streaming Responses

### LINKED ISSUE #250: Non-streaming Responses executor drops upstream error headers
### Description

For non-streaming `POST /v1/responses` requests routed through the executor (for example, `store: true`), upstream HTTP errors lose their response headers at the HTTP adapter. The upstream status and body are preserved, but `Retry-After` and request IDs are dropped, and `Content-Type` is forced to `application/json` even when the forwarded error body is plain text.

**Validation status (updated 2026-09-06 UTC):** reproduced in Linux CI using real HTTP requests through the gateway to a local mock upstream, against unchanged runtime source from `main` at `c04ac196f44133c37be652c4ee3f777db32ae6df`. The two executor HTTP regressions and the adapter regression below fail on the baseline; all six tests pass after applying an isolated adapter patch. [CI run and logs](https://github.com/MecicoJay/agentic-api/actions/runs/34006465795). This AI-assisted report was initially based on source inspection; runtime evidence is now available.

### Runtime evidence

The mock returns HTTP 429 with `Retry-After: 7`, `x-request-id: req_example`, `x-ratelimit-remaining-requests: 0`, and either JSON or plain text. All requests below use `stream:false`.

| Case | Baseline | With isolated patch |
| --- | --- | --- |
| `store:false`, JSON | All three tested metadata headers and content type preserved | Same; pass |
| `store:false`, plain text | All three tested metadata headers and content type preserved | Same; pass |
| `store:true`, JSON | All three tested metadata headers missing | Headers restored; pass |
| `store:true`, plain text | All three tested metadata headers missing; content type incorrectly `application/json` | Headers restored and `text/plain` retained; pass |
| Malformed client JSON | HTTP 400 with JSON `invalid_request_error` | Same; pass |
| Direct adapter regression below | Fails on missing `Retry-After` | Pass |

HTTP 429 and the exact mock error body are preserved in all four upstream-error cases. The tested `Connection`-nominated header, `x-upstream-hop`, remains absent before and after the patch.

- Validation commit: [`b2548adfd8d1ca916c7606fce64486122e2473fa`](https://github.com/MecicoJay/agentic-api/commit/b2548adfd8d1ca916c7606fce64486122e2473fa).
- [HTTP reproduction tests](https://github.com/MecicoJay/agentic-api/blob/b2548adfd8d1ca916c7606fce64486122e2473fa/crates/agentic-server/tests/response_error_headers_repro.rs), [causal patch](https://github.com/MecicoJay/agentic-api/blob/b2548adfd8d1ca916c7606fce64486122e2473fa/verification/issue250-causal-fix.patch), and [CI workflow](https://github.com/MecicoJay/agentic-api/blob/b2548adfd8d1ca916c7606fce64486122e2473fa/.github/workflows/issue250-reproduction.yml).
- CI first compiles both targets, requires the baseline controls to pass and the three regressions to fail, then applies the patch and requires the same tests to pass. The green run therefore includes verified baseline failures. Formatting and scoped Clippy with `-D warnings` also pass.
- Scope: mock HTTP integration plus adapter tests, using the existing disabled-storage test harness; the upstream 429 occurs before persistence. No live model, GPU, database persistence, streaming path, or full workspace suite was tested. The patch is validation evidence, not an upstream merge.

### Steps to reproduce

With a normally configured gateway using a mock upstream:

1. Configure the mock's `POST /v1/responses` to return:
   ```http
   HTTP/1.1 429 Too Many Requests
   Content-Type: text/plain
   Retry-After: 7
   x-request-id: req_example

   rate limited
   ```
2. Send this valid request to the gateway:
   ```json
   {"model":"test","input":"hi","store":true,"stream":false}
   ```
3. Inspect the gateway's HTTP status, headers, and body. A `store:false` request with no in-process features is a useful direct-proxy control.

A smaller regression can exercise the responsible adapter directly. Place the following in `crates/agentic-server/tests/responses_error_headers_test.rs` and run:

```bash
cargo test -p agentic-server --test responses_error_headers_test
```

<details>
<summary>Adapter regression test (executed in Linux CI: fails before patch, passes after)</summary>

```rust
use agentic_core::executor::ExecutorError;
use agentic_server::handler::executor_error_response;
use axum::body::to_bytes;
use http::{HeaderMap, HeaderValue, StatusCode};

#[tokio::test]
async fn responses_executor_preserves_upstream_error_headers() {
    let mut headers = HeaderMap::new();
    headers.insert("retry-after", HeaderValue::from_static("7"));
    headers.insert("x-request-id", HeaderValue::from_static("req_example"));
    headers.insert("content-type", HeaderValue::from_static("text/plain"));

    let response = executor_error_response(ExecutorError::LLMRequest {
        status: StatusCode::TOO_MANY_REQUESTS,
        body: "rate limited".to_string(),
        headers,
    });

    assert_eq!(response.status(), StatusCode::TOO_MANY_REQUESTS);
    assert_eq!(
        response.headers().get("retry-after"),
        Some(&HeaderValue::from_static("7"))
    );
    assert_eq!(
        response.headers().get("x-request-id"),
        Some(&HeaderValue::from_static("req_example"))
    );
    assert_eq!(
        response.headers().get("content-type"),
        Some(&HeaderValue::from_static("text/plain"))
    );
    let body = to_bytes(response.into_body(), 1024).await.unwrap();
    assert_eq!(&body[..], b"rate limited");
}
```

</details>

### Expected behavior

Preserve the upstream error's status, body, and eligible processed headers, including `Retry-After`, request IDs, and a content type consistent with the emitted body. Keep the existing filtering of hop-by-hop and invalidated representation headers.

This matches the intent documented by [processed_response_headers](https://github.com/vllm-project/agentic-api/blob/c04ac196f44133c37be652c4ee3f777db32ae6df/crates/agentic-server-core/src/proxy.rs#L168), which explicitly retains request IDs, retry guidance, and rate-limit metadata.

### Actual behavior confirmed at runtime

For the example above, the adapter constructs HTTP 429 with the unchanged body `rate limited` and `Content-Type: application/json`, without `Retry-After` or `x-request-id`.

The relevant chain is:

- [The non-streaming Responses executor forwards errors to executor_error_response](https://github.com/vllm-project/agentic-api/blob/c04ac196f44133c37be652c4ee3f777db32ae6df/crates/agentic-server/src/handler/http/responses.rs#L31).
- [Inference already captures the processed upstream error headers in ExecutorError::LLMRequest](https://github.com/vllm-project/agentic-api/blob/c04ac196f44133c37be652c4ee3f777db32ae6df/crates/agentic-server-core/src/executor/inference.rs#L90).
- [executor_error_response builds a response with only a fixed Content-Type and the error body](https://github.com/vllm-project/agentic-api/blob/c04ac196f44133c37be652c4ee3f777db32ae6df/crates/agentic-server/src/handler/common.rs#L30).

Clients therefore lose upstream retry timing and request correlation; plain-text errors can also be mislabeled as JSON.

### Environment

- Inspected upstream commit: `c04ac196f44133c37be652c4ee3f777db32ae6df`.
- Endpoint: HTTP `POST /v1/responses`, `store:true, stream:false`.
- Runtime: Ubuntu Linux x86_64 in GitHub Actions; Rust and Cargo 1.98.0.
- Validation commit: `b2548adfd8d1ca916c7606fce64486122e2473fa`, with unchanged baseline runtime source before the CI applies the patch.
- Both regression targets were compiled and executed in CI using synthetic error data; no GPU or live model service was needed. The local Windows linker limitation was bypassed by running on Linux.

### Additional context

The neighboring [Messages error adapter](https://github.com/vllm-project/agentic-api/blob/c04ac196f44133c37be652c4ee3f777db32ae6df/crates/agentic-server/src/handler/http/messages.rs#L44) already preserves the processed upstream headers, with [a regression checking retry-after and request-id](https://github.com/vllm-project/agentic-api/blob/c04ac196f44133c37be652c4ee3f777db32ae6df/crates/agentic-server/tests/messages_test.rs#L235). That provides a local implementation and testing precedent.

I checked issue/PR searches for `retry-after` and `Responses headers`; I did not find a report covering this non-streaming executor case. Related transport work includes #161 and #217. A focused fix could preserve the headers already carried by `ExecutorError::LLMRequest` in the common HTTP adapter.

ZichenYuan:
Route ExecutorError::LLMRequest through a shared upstream_error_response adapter so the non-streaming Responses executor path returns the upstream status, body, and processed metadata headers (retry-after, request IDs, rate-limit headers), and only defaults content-type to application/json when the upstream did not label its body. The Messages handler now uses the same adapter instead of its private copy.

Closes #250


crates/agentic-server/tests/responses_error_headers_test.rs
```diff
@@ -0,0 +1,169 @@
+//! Upstream error metadata must survive the non-streaming Responses executor path.
+//!
+//! Regression coverage for <https://github.com/vllm-project/agentic-api/issues/250>: when a
+//! request is routed through the in-process executor (for example `store: true`) and the
+//! upstream returns a non-2xx response, the gateway must preserve the upstream status, body,
+//! and processed metadata headers (`retry-after`, request IDs, rate-limit headers) and must not
+//! relabel a non-JSON error body as `application/json`.
+
+#[allow(dead_code)]
+mod common;
```
maralbahari, line 12: is it the import of whole `common` mod in this file causing clippy warning with `dead_code`? if so can just import only the function needed from `common` like line 18 of this file.

ZichenYuan: Good question — I checked, and the warning isn't from the import. The `use` line already pulls in only `spawn_gateway`, `test_config`, and `test_state`. Removing `#[allow(dead_code)]` gives:

```
error: function `spawn_mock_llm` is never used
  --> crates/agentic-server/tests/common/mod.rs:54:14
```

Each integration test file is its own crate, so `mod common;` compiles all of `common/mod.rs` into this binary, and `spawn_mock_llm` (which this test doesn't need) becomes dead code. The other tests that don't use it (`messages_test`, `compaction_test`, `oidc_auth_test`, `responses_websocket_test`) carry the same attribute for the same reason. I kept it and added a short comment explaining why in f61bdef.

franciscojavierarceo (commented):
checked the shared error adapter and the upstream header-filtering path. this preserves the upstream metadata and content type while keeping local failures in their existing error envelopes. no changes requested from the source review; i haven't rerun the new integration tests locally.

maralbahari (approved):

Issue #170, opened 2026-08-07
Add OpenAI-compatible shell tool support for hosted and local execution


maralbahari:
### Problem statement / motivation

The [Server-Side Tool Execution roadmap](https://github.com/vllm-project/agentic-api/blob/main/ROADMAP.md#3-server-side-tool-execution) lists **Shell tools** as an expected tool area, but `agentic-api` does not currently support the OpenAI Responses API `shell` tool.

The Rust request model has no typed `shell` declaration or `shell_call` / `shell_call_output` items, and the tool loop has no configurable local shell executor. This prevents a self-hosted `agentic-api` deployment from acting as the caller-controlled local runtime described by the OpenAI API.

This enhancement is intentionally limited to the current local execution shape:

```json
{
  "type": "shell",
  "environment": {
    "type": "local"
  }
}
```

It does **not** require a Containers API, hosted execution environments, or container lifecycle management.

The older tool named `local_shell`, with `local_shell_call` items, is documented by OpenAI as outdated. New implementation work should target the current `shell` tool with `environment.type: "local"`.

### Proposed solution

Add first-class support for the current OpenAI-compatible local `shell` tool and allow deployments to register an explicit local shell executor.

### Wire types

- Accept and preserve `{"type": "shell", "environment": {"type": "local"}}`.
- Add typed input/output items for:
  - `shell_call`
  - `shell_call_output`
- Preserve:
  - `call_id`
  - `action.commands[]`
  - `timeout_ms`
  - `max_output_length`
  - per-command `stdout` and `stderr`
  - `exit` outcomes with `exit_code`
  - `timeout` outcomes
  - call status
- Keep the wire representation non-exhaustive so future fields can be preserved safely.
- Do not treat unsupported shell environments as permission to execute. Preserve/pass through or return a clear unsupported-capability error according to the existing tool ownership rules.

### Tool routing and execution

- Register the local `shell` tool in the request-scoped tool registry.
- If upstream inference requires function-shaped tools, normalize the declaration internally and restore model calls to the public typed `shell_call` shape.
- Execute shell calls only when an explicit gateway shell handler and execution policy are configured.
- Treat the configured `agentic-api` execution environment as the local runtime: execute the requested commands, capture stdout/stderr and outcomes, append `shell_call_output`, and continue the inference loop.
- If no gateway shell handler is configured, preserve/return the `shell_call` so a client-controlled runtime can execute it and submit the matching `shell_call_output`.
[...]

PR #264, opened 2026-09-07
feat: add shell tool wire types

bojiang3:
## Summary

- Add typed local shell declarations, `shell_call` and `shell_call_output` items, command limits, stdout/stderr, and exit/timeout outcomes.
- Normalize shell declarations, selectors, and inference history to function tools, then restore public shell items and incremental command events. Native shell streams share the existing strict lifecycle validation.
- Preserve submitted public shell items and extension fields in storage; lower the inference copy when preparing new input or rehydrating later turns. Restore public shell declarations and selectors in response metadata, including inherited settings and requests using tool search.
- Keep execution client-side: callers can run commands through their own engine or framework and submit `shell_call_output`. This PR does not register a gateway-side shell executor; that remains follow-up work.
- Include 16 real OpenAI-reference and gateway recordings covering success, nonzero exits, timeouts, and multiple commands in streaming and non-streaming modes, with two-turn continuation and replay/parity tests.

Part of #170.

## Test Plan

- `cargo test --offline --workspace`: 1,198 passed, 9 ignored.
- `cargo test --offline -p agentic-server-core`: 870 passed, 9 ignored after the final assertions.
- `cargo clippy --offline --workspace --all-targets -- -D warnings`.
- `uvx --from pre-commit pre-commit run --all-files`, including formatting and merge-conflict checks.
- Regression tests first reproduced the persistence and streaming metadata failures, then passed with the fixes. Coverage includes response and conversation storage, extension fields, a third turn loading persisted shell output, inherited tools/selectors, and shell metadata with and without tool search.
- Existing recordings were replayed; no new live provider recordings were made for the final two fixes.


crates/agentic-server-core/src/tool/normalize.rs
```diff
@@ -77,6 +79,10 @@ impl ResponsesTool {
                 tracing::debug!("code_interpreter tool skipped in normalize - handler not yet registered");
                 vec![]
             }
+            Self::Shell(_) => {
+                tracing::debug!("shell tool skipped in normalize - handler not yet registered");
```
maralbahari: @bojiang3 `shell` tool type as below taken from openai examples in https://developers.openai.com/api/docs/guides/tools-shell
`
   "tools": [
      {
        "type": "shell",
        "environment": {
          "type": "container_auto",
          "network_policy": {
            "type": "allowlist",
            "allowed_domains": ["pypi.org", "files.pythonhosted.org", "github.com"]
          }
        }
      }
`

is not recognized by vLLM upstream. even if some of the types might be recognized by upstream still to keep agentic-api consistent and compatible with multiple upstreams for inference we always normalize/canonicalize any tool types into `function` types. so all tool support on agentic-api either client owned like shell or owned on gateway by agentic-api need to implement a `ToolHandler` interface so that we can normalize and validate the tools's schema for inference. kindly follow ARCHITECTURE.md and TERMINOLOGY.md.

there should be also request and responses cassettes recording for this tool type from openai as ground truth and agentic-api (gateway) to match the behavior of gateway with openai Responses API. you can read https://github.com/vllm-project/agentic-api/blob/main/crates/agentic-server-core/tests/cassettes/README.md how to record a cassettes and there are shell script examples like https://github.com/vllm-project/agentic-api/blob/main/crates/agentic-server-core/tests/cassettes/record_web_search_cassettes.sh, https://github.com/vllm-project/agentic-api/blob/main/crates/agentic-server-core/tests/cassettes/record_mcp_cassettes.sh
that use our python cassettes recorder.

franciscojavierarceo (commented):
the local shell wire types, call/output pairing, and storage round-trip look consistent with this slice. i don't have additional findings beyond the existing handler and cassette feedback. this was a source review; i haven't independently run the tests.

crates/agentic-server-core/src/tool/normalize.rs
bojiang3: Addressed the handler/normalization feedback in 0aaebc9: `shell` now has a client-owned `ShellHandler`, validates local-only environments, registers as client-owned, normalizes to a vLLM-compatible function schema, and restores both blocking and streaming calls to typed `shell_call` items. Unsupported environments and malformed actions fail closed; no gateway command execution was added. I also added `record_shell_cassettes.sh` for matching OpenAI/gateway streaming and non-streaming recordings. I could not capture authentic YAML in this environment because `OPENAI_API_KEY` is unset and the configured gateway endpoint is unavailable, so I intentionally did not hand-author fixtures. Validation: full `cargo test -p agentic-server-core` passed (448 passed, 8 environment-dependent ignored), plus fmt and clippy with warnings denied.

maralbahari: @bojiang3 do you have resources to record cassettes for both gatewayand for openai ? openai would need api key! if you cant actually record cassettes , I will record and push in your branch then you can write integration test comparing openai ground truth to agentic-api behavior. cassettes are a required for landing features specially for tool supports.

franciscojavierarceo (approved):
the remaining persistence and response metadata issues are fixed in bb9fc65. public shell items now survive storage, later continuations normalize correctly, and response metadata preserves the shell declaration and selector. the regressions, workspace tests, Clippy, and pre-commit checks pass. client execution stays framework-independent; a gateway-side executor remains follow-up work.

Issue #241, opened 2026-09-04
[RFC]: one linear ingest pipeline for the executor streaming path


maralbahari:
### Problem statement / motivation

The llm-d split stream relay landed in
[#235](https://github.com/vllm-project/agentic-api/pull/235) and
[#236](https://github.com/vllm-project/agentic-api/pull/236). Both were the right fix for
the problem in front of them, and together they pushed a second ingest leg and a second
validation regime into the core, landing the cost on three already-coupled files:
`ResponseAccumulator`, `upstream.rs` and `inference.rs`.

The strict validator carries no state of its own; it reuses the accumulator's `in_flight`
and `completed` maps, so validation and accumulation are now one object. That is why this
needs a redesign rather than a tidy-up: the three files can no longer be changed
independently.

Where that shows up today:

- **Adding one output-item kind edits 8 sites in `accumulator.rs`.** `ARCHITECTURE.md:348`
  calls the file "a stability contract" that "should not grow", but its documented extension
  point is "add a variant and the match arms".
- **Four of its six public methods have no production caller**: `from_stream` (bench only),
  `from_sse_lines` (tests only), `mark_incomplete` (nothing), `process_stream_chunks` (only
  reachable from `from_stream`). The `spawn_blocking` design in the module doc benchmarks
  well, and the live path re-implements the loop inline without it.
- **Ingest and emission are interleaved** in one 70-line loop (`upstream.rs:208-252`), so
  neither can be tested or changed alone.
- **The same rules are written several times and disagree**: five in-flight lookup rules;
  three enumerations of the terminal-event set, one of which silently drops `error` on
  `response.completed`; two deferred-frame drains that order index-less frames oppositely;
  two emit funnels with the same tail; byte-identical helpers in `upstream.rs` and
  `events/validate.rs`.
- **The two ingress legs disagree on `data:` spacing.** The live path requires the space
  (`inference.rs:187`), the relay path does not. The tolerance is tested but unreachable
  live, and `data:[DONE]` without a space fails to terminate the stream.

**The entanglement has already produced a live bug.** A repeated
`response.output_item.done` yields **two** output items instead of one on the lenient path.
The strict path rejects it, so only `fetch_stream_payload` is affected. It cannot be fixed
locally: the change that repairs the lenient path makes every strict decode fail with
`upstream stream ended with unfinished output items`.

### Proposed solution

One ingest pipeline, with the non-linear part confined to one named component.

[...]

PR #274, opened 2026-09-10
refactor: unify response processing under `AgentPipeline`

### LINKED ISSUE #243: refactor: unify synchronous streaming ingestion and item lifecycle
the live and strict decode paths have separate ingestion rules, and the current active/completed split lets repeated `response.output_item.done` events append duplicate output items on the lenient path. we should give both paths one synchronous state machine before changing where it runs.

proposed first slice of #241; coordinate with the still-open #236 so we don't duplicate its provider-compatibility work. this is a suggested scope, not a fixed implementation plan; feel free to revise, combine, or remove it.

we should be able to verify that:

- [ ] one synchronous entry point owns line filtering, normalization, lifecycle validation, folding, translation, and finalization, with strict/lenient validation policies over the same transitions.
- [ ] typed active-item variants encode the item kind and its valid construction/buffer state; completed slots remain available for lifecycle checks.
- [ ] slots use validated `output_index` values scoped to an inference round, with canonical item IDs and checks for mismatches or ID reuse across indexes. missing/invalid indexes have an explicit rejection or recovery policy before lookup, not an implicit `0` key.
- [ ] strict validation rejects repeated `done`; lenient validation suppresses semantically identical repeats in both accumulated output and emitted events, and rejects conflicting authoritative content. define which item kinds allow `done` without `added` under lenient validation.
- [ ] consuming finalization preserves the terminal-event/error contract; its documentation promises no reuse or double-finalization, not that dropping the pipeline is impossible.
- [ ] regression tests cover duplicate/conflicting completion, supported done-only items, index/ID reuse, wrong-kind deltas, truncated streams, authoritative completion payloads, and `data:` spacing across affected ingress paths, including the Messages bridge.

keep worker placement and relay extraction outside this slice. the existing [index normalizer](https://github.com/vllm-project/agentic-api/blob/39180af6357662bde94b60163b3e374cf43ed2bc/crates/agentic-server-core/src/events/normalize.rs#L111) and [strict decode tests](https://github.com/vllm-project/agentic-api/blob/39180af6357662bde94b60163b3e374cf43ed2bc/crates/agentic-llm-d/tests/split_execution_integration.rs) are useful starting points.

maralbahari:
## Summary

The Responses executor previously split JSON/SSE accumulation, tool-call translation, and stream emission across `upstream.rs`, `accumulator.rs`, and `function_sse.rs`. This made ownership unclear and allowed live and collected responses to follow different processing paths.

This PR implements #243 from RFC #241 by introducing a request-owned `AgentPipeline`:

- Route complete JSON and live SSE bodies through the same pipeline and finalization path.
- Add a synchronous, round-scoped `RoundIngestion` that combines typed response accumulation with tool-call translation.
- Split `ResponseAccumulator` into focused modules with typed active/completed slots keyed by validated `output_index`.
- Apply strict and lenient validation policies through the same state transitions.
- Replace the monolithic function SSE translator with an inline dispatcher and tool-specific translators for function, custom, namespace, and tool-search calls.
- Keep `ToolRegistry` as the request-scoped lookup for available tools and their client/gateway ownership. Translators consume an owned classification snapshot.
- Keep gateway-owned tool execution and public lifecycle generation in the gateway path.
- Move ordered, awaited stream delivery into `AgentPipeline`, preserving cross-round sequencing, output-index rebasing, deferred event ordering, and downstream backpressure.
- Encapsulate the tool loop, registry, response budget, and pipeline under `EngineOrchestration`.
- Keep `upstream.rs` focused on request serialization, inference transport, response-budget accounting, and selecting the JSON or SSE pipeline runner.
- Document the final ownership boundaries, canonical tool normalization, client translation, and `OutputItem::to_input_item` as the continuation conversion policy.
- Refresh the gateway replay cassettes for custom tools, MCP, parallel tool calls, tool search, and web search.

Closes #243.
Part of #241.

## Test Plan

- `cargo test --locked --workspace -- --test-threads=2`
  - 1120 passed, 0 failed, 9 ignored
- `cargo clippy --locked --workspace --all-targets -- -D warnings`
- `git diff --check`
- Re-recorded and replayed the affected gateway cassettes.


franciscojavierarceo (approved):
Reviewed ingestion, streaming delivery, tool translation, and persistence boundaries at e03fb7e. No actionable regressions found. Validation: 877 core tests and 19 split-execution tests passed; 9 tests were ignored. Cassette replay coverage passed; live providers were not tested.

PR #296, opened 2026-09-13
fix: let Messages answer after forced tool use

StevenWang-CY:
## Summary

Let a Messages request that forces a gateway tool finish with an answer from its result. Previously, `any` remained forced on every inference round, causing repeated searches and eventual round exhaustion. vLLM 0.29.0 also labels complete named calls `end_turn`, which could prevent gateway execution or make clients skip their own tool calls.

- Preserve the first selector, then relax a fulfilled `any` or matching `tool` to `auto` after appending the full assistant turn and paired tool results. Preserve parallel-use settings and extensions; remove `name` only for a named selector.
- Accept the provider's `end_turn` internally for a matching, explicitly selected gateway call. Surface completed client-executed function calls with `stop_reason: tool_use` in JSON and the terminal SSE `message_delta`. Hide gateway calls in mixed rounds. Streaming normalization requires `message_stop`; token limits, other stop reasons and unfinished streams keep their terminal semantics.
- Model Messages `tool_choice` as a tagged enum with variant-specific fields, the existing non-empty tool-name type and flattened extensions. Use exhaustive matches for fulfillment and transition, while keeping raw JSON for lossless forwarding. Invalid gateway-tool requests return HTTP 400 with `invalid_request_error` before inference instead of escaping through the proxy fallback. Requests without gateway tools retain transparent proxying and upstream validation.
- Add recorded gateway and client conversations, negative controls and OpenAPI validation. No new wire fields, runtime dependencies, configuration or storage migration.

Anthropic clients distinguish [completed turns from requests for client tool execution](https://platform.claude.com/docs/en/build-with-claude/handling-stop-reasons). The provider compatibility case is visible in vLLM's [named-call finish reason](https://github.com/vllm-project/vllm/blob/v0.29.0/vllm/entrypoints/openai/chat_completion/serving.py#L1071-L1078) and [Messages conversion](https://github.com/vllm-project/vllm/blob/v0.29.0/vllm/entrypoints/anthropic/serving.py#L633-L644).

## Test Plan

- Live, unmodified vLLM 0.29.0 and pinned `Qwen/Qwen3-4B`: four gateway recordings cover `any`/named selection × JSON/SSE. Each completes with one local search and two provider requests; retaining the forced selector causes another call in all four live controls.
- Two additional live recordings use Anthropic SDK 1.5.0. The SDK receives `tool_use`, submits the client output with the returned call ID, and receives the complete verification string. Both raw provider recordings contain `end_turn`. Replay checks the complete requests, response metadata and streaming lifecycle, repeats conversations on one gateway, and verifies no gateway search or Messages persistence. All six captures are unmodified recorder output.
- Identical tests in a separate build of prior head `1e422c3`: both new recorded client conversations and the client-stop/invalid-selector regressions fail as intended; all four gateway recordings still pass. The corrected commit passes all six. Earlier unchanged-main validation established failure of all four gateway scenarios before the original fix.
- 1,402 Rust tests pass with nine existing default ignores. This includes malformed selectors, configured aliases, extension preservation, mixed ownership, truncation/unfinished-stream controls and OpenAPI schema conformance.
- Locked Clippy across the workspace, all targets and features; formatting; all repository hooks; strict docs; gateway binaries; 13 recorder tests plus three subtests; and validation of all 139 cassettes / 308 turns pass.


crates/agentic-server-core/src/executor/messages_loop.rs
```diff
@@ -133,11 +134,16 @@ pub async fn run_messages_loop(
             }
         }
 
-        // Terminal when the model didn't ask for a gateway tool, or stopped for
-        // another reason. A client-owned tool_use is also terminal (the client
-        // must run it) — but the gateway tool_use, if any, must still be hidden
+        // The shared context accepts tool_use and vLLM's end_turn for a matching
+        // named call. Other stops remain terminal. A client-owned tool_use is
```
maralbahari: Named client-executed tool calls leak vLLM’s incorrect `end_turn`

When vLLM returns a named client-executed `tool_use` with `stop_reason: end_turn`, the blocking path returns it unchanged, while the streaming path re-emits the buffered `end_turn`. The new test at `messages_tool_choice_test.rs:365` codifies this behavior.

Anthropic defines `tool_use` as meaning the client must execute the tool, while `end_turn` means the response is complete. Its canonical tool loop explicitly continues only when `stop_reason == "tool_use"`. See the [Anthropic stop-reason documentation](https://platform.claude.com/docs/en/build-with-claude/handling-stop-reasons) and [tool-use loop](https://platform.claude.com/docs/en/agents-and-tools/tool-use/how-tool-use-works).

Impact: Anthropic-compatible clients and tool runners may treat this response as final, skip execution of the client tool, and never submit the required `tool_result`.

Fix: whenever a client-executed `tool_use` is surfaced, normalize the public non-streaming `stop_reason` and streaming `message_delta.delta.stop_reason` to `tool_use`. Keep the `end_turn` compatibility exception internal to hidden gateway-executed calls.

crates/agentic-server-core/src/executor/messages_context.rs
```diff
@@ -194,10 +194,35 @@ impl MessagesRequestContext {
...
+        &self,
+        stop_reason: Option<&str>,
+        mut gateway_names: impl Iterator<Item = &'a str>,
+    ) -> bool {
+        if stop_reason == Some("tool_use") {
+            return true;
+        }
+        stop_reason == Some("end_turn")
+            && self.raw.get("tool_choice").is_some_and(|choice| {
+                choice["type"] == "tool"
```
maralbahari: The new selector state machine is driven by untyped JSON

The PR now makes control-flow decisions by indexing raw `serde_json::Value` fields and comparing string literals such as `"any"`, `"tool"`, and `"name"`. `MessagesRequest.tool_choice` remains `Option<Value>`, even though Anthropic defines `tool_choice` as a tagged union of `auto`, `any`, `tool { name }`, and `none`. See the [Anthropic Messages type definition](https://platform.claude.com/docs/en/api/typescript/messages).

This conflicts with the repository’s typed-boundary requirement and the Rust skill rules `api-parse-dont-validate`, `type-enum-states`, and `type-no-stringly`. Invalid states—including arrays, booleans, missing tool names, and unknown discriminators—remain representable, while future protocol changes could silently bypass selector relaxation.

Impact: selector validation and transitions are enforced only through runtime string comparisons, so malformed or newly introduced variants can produce inconsistent behavior without compiler assistance.

Fix: model `tool_choice` as a typed enum under `types/messages`, with variant-specific fields and flattened extension fields where forward compatibility is required. Preserve raw JSON for lossless forwarding, but perform fulfillment checks and the transition to `auto` using the typed representation.

franciscojavierarceo (approved):

franciscojavierarceo (approved):

Issue #244, opened 2026-09-04
refactor: extract StreamRelay and bound client delivery


franciscojavierarceo:
ingestion and client emission currently share control flow, while the final client-facing channel is unbounded. bounding proposed worker queues alone would still allow slow-client buffering to grow. we should isolate emission and make backpressure reach the client boundary, independently of any worker-placement decision.

proposed second slice of #241, building on the synchronous ingestion boundary in #243. this is a suggested scope; feel free to revise, combine, or remove it.

we should be able to verify that:

- [ ] `StreamRelay` owns upstream/local emission, deferred-frame ordering, and sequence numbering, while normalization and item lifecycle state stay in the synchronous pipeline.
- [ ] upstream response-ID rewriting and tool-name restoration are not applied to already-correct local events; deferred output preserves the gateway-executed tool ordering contract and contiguous sequence numbers.
- [ ] client delivery and deferred buffers have explicit entry/byte limits, including handling for a single oversized frame. document retained response-state memory separately rather than claiming queue bounds cap the whole request.
- [ ] slow consumers apply backpressure without unbounded spillover or silently dropped semantic events; client disconnect and producer failure release resources and propagate errors without hanging.
- [ ] tests exercise deferred/index-less ordering, mixed upstream/local events, tiny queue capacities, oversized frames, slow clients, and cancellation. expose queue occupancy and buffered-byte measurements to tune limits.

the [existing client channel](https://github.com/vllm-project/agentic-api/blob/39180af6357662bde94b60163b3e374cf43ed2bc/crates/agentic-server-core/src/executor/engine.rs#L463) is part of this boundary. no `spawn_blocking` migration is required for this slice.

maralbahari:
Sub-issue of [#241](https://github.com/vllm-project/agentic-api/issues/241). Builds on
[#243](https://github.com/vllm-project/agentic-api/issues/243), which leaves ingestion as a
synchronous `AgentPipeline`. This issue takes everything ingestion is not: emission,
ordering, and backpressure. Worker placement stays out.

## Interfaces

```rust
// executor/relay.rs (new)

#[non_exhaustive]
pub struct RelayLimits {
    pub deferred_frames: usize,
    pub deferred_bytes: usize,     // today's hardcoded 256 KiB
    pub client_queue_depth: usize, // new: the client channel is unbounded today
}

pub struct StreamRelay {
    gateway: GatewayStreamAccumulator,
[...]

PR #302, opened 2026-09-14
fix: make stream delivery cancellation-safe

LOGO127:
## Summary

- unify gateway-synthesized and translated upstream events behind the same awaited client-emission path
- commit sequence/lifecycle presentation state only after a bounded-channel send succeeds, so cancellation, receiver closure, or oversized events do not consume state
- bound deferred delivery by both serialized bytes (256 KiB) and event count (1024)
- keep index-less deferred events behind indexed output and preserve unsent frames plus byte accounting when a flush is cancelled
- document the delivery ownership and cancellation semantics

This is the implementation work for #244. It keeps ingestion and worker-placement changes out of scope.

## Test Plan

- `cargo test -p agentic-server-core pipeline::delivery` — 11 passed
- `cargo test --workspace --all-features` — full Linux-native workspace run passed with 0 failures
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- existing cassette replay tests pass as part of the workspace run

Fresh provider/gateway cassette recordings have not been generated yet. The checked-in replay set is unchanged; if re-recording is required as a merge gate, I can refresh the affected recordings before merge.

Refs #244


crates/agentic-server-core/src/executor/pipeline/delivery.rs
```diff
@@ -143,17 +145,47 @@ fn should_defer_stream_event(frame: &EventFrame, defer_from_output_index: Option
         frame
             .wire
             .output_index
-            .is_some_and(|output_index| output_index >= first_hidden_index)
+            .is_none_or(|output_index| output_index >= first_hidden_index)
```
franciscojavierarceo, line 148: this now buffers an indexless upstream `error` after a hidden gateway tool call, but the engine returns on `response.failed` without flushing deferred events. that silently drops the error's diagnostic fields; previously, the error passed through immediately.

i reproduced this through `ExecuteRequest`: the client receives `response.failed`, but the preceding error's unique diagnostic disappears. we should preserve deferred diagnostics on the failure path without executing tools and add an executor regression test for this sequence.

franciscojavierarceo (approved):

crates/agentic-server-core/src/executor/pipeline/delivery.rs
LOGO127: Thanks for catching this and fixing it in b3040be. I missed the early return on `response.failed` when changing index-less deferral.

I reran the failure-path regression and the delivery/web-search tests on `51e2405`; all pass. The regression confirms that the diagnostic arrives before the terminal event with contiguous sequence numbers, without executing tools. Thanks for the regression test and merge!

PR #304, opened 2026-09-15
fix(executor): account logical retained bytes instead of raw SSE line bytes in response budget (#288)

### LINKED ISSUE #288: Bug: cumulative response budget rejects small outputs based on SSE chunking (0.6.0/main)
### Description

Executor-backed responses can fail after only **7,655 bytes of answer text** because the request's fixed **1 MiB cumulative response budget** counts serialized SSE envelopes, every delta, and repeated done/terminal snapshots. The same 8,000-byte answer succeeds when sent with coarser deltas or as non-streaming JSON. A fast, continuously draining client does not prevent the failure.

This is still reproducible at main commit **[`2509342dcbc12b89350cabdda17f7bb1a6b8a41b`](https://github.com/vllm-project/agentic-api/commit/2509342dcbc12b89350cabdda17f7bb1a6b8a41b)**, including its new WebSocket `store: false` session path. It affects the shared executor, not only WebSocket transport. The fixed upstream SSE-line and JSON-body ceilings are separate constraints.

The resource protections added in [#240](https://github.com/vllm-project/agentic-api/pull/240) are useful, but lifetime wire-byte accounting makes ordinary response capacity depend on chunking and repeated representations. This report proposes separating configurable retained-response, upstream-wire, and client-event limits while keeping bounded queues and cancellation.

### Steps to reproduce

The following deterministic upstream uses Python's standard library. It supplies an ordinary Responses lifecycle with deltas and matching full done/terminal snapshots. It requires no model, GPU, SDK, or external service.

1. Save the script below as `public_repro.py`, then run `python public_repro.py server`.
2. In a separate terminal, build/run the gateway from the selected checkout with its locked dependencies:

   ```sh
   cargo run --locked -p agentic-server --bin agentic-server -- --llm-api-base http://127.0.0.1:18001 --gateway-host 127.0.0.1 --gateway-port 18002 --db-url 'sqlite://budget-repro.db?mode=rwc'
   ```

3. Run `python public_repro.py client`. It requests the same 8,000-byte answer with one-byte and 1,024-byte chunks, then a 220,000-byte answer with 1,024-byte chunks.

Use an isolated config/home and database; the recorded harness clears unrelated gateway/provider environment overrides. On Windows, use the native-build caveat in Environment below. The fixture bytes in this reduced script were checked against the fixture bytes used by the full HTTP/SSE/WebSocket harness.

<details>
<summary>Self-contained synthetic upstream and HTTP SSE client</summary>

```python
"""Run `python public_repro.py server`, start the gateway, then run `... client`."""
import http.client
import json
import sys
from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer

def encode(value):
    return json.dumps(value, separators=(',', ':')).encode()

def fixture(size, chunk):
    text = 'x' * size
    part = dict(type='output_text', text=text, annotations=[])
    item = dict(id='msg_fixture', type='message', role='assistant', status='completed', content=[part])
    response = dict(id='resp_fixture', object='response', created_at=1789151000,
                    model='fixture', status='completed', output=[item], error=None, incomplete_details=None)
    initial = dict(response, status='in_progress', output=[])
    events = []
    def add(kind, **fields):
        events.append(dict(type=kind, sequence_number=len(events), **fields))
    add('response.created', response=initial)
    add('response.in_progress', response=initial)
    add('response.output_item.added', output_index=0, item=dict(item, status='in_progress', content=[]))
    add('response.content_part.added', output_index=0, item_id=item['id'], content_index=0, part=dict(part, text=''))
    for at in range(0, size, chunk):
        add('response.output_text.delta', output_index=0, item_id=item['id'], content_index=0, delta=text[at:at+chunk])
    add('response.output_text.done', output_index=0, item_id=item['id'], content_index=0, text=text)
    add('response.content_part.done', output_index=0, item_id=item['id'], content_index=0, part=part)
    add('response.output_item.done', output_index=0, item=item)
    add('response.completed', response=response)
    stream = b''.join(b'data: ' + encode(event) + b'\n\n' for event in events) + b'data: [DONE]\n\n'
    return response, stream

class Upstream(BaseHTTPRequestHandler):
    protocol_version = 'HTTP/1.1'
    def log_message(self, *args):
        pass
    def reply(self, body, content_type):
        self.send_response(200)
        self.send_header('Content-Type', content_type)
        self.send_header('Content-Length', str(len(body)))
        self.end_headers()
        try:
            self.wfile.write(body)
        except (BrokenPipeError, ConnectionResetError, ConnectionAbortedError):
            pass
    def do_GET(self):
        self.reply(encode({'object':'list','data':[{'id':'fixture','object':'model','owned_by':'fixture'}]}), 'application/json')
    def do_POST(self):
        request = json.loads(self.rfile.read(int(self.headers['Content-Length'])))
        _, size, chunk = request['model'].split('-')
        response, stream = fixture(int(size), int(chunk))
        self.reply(stream if request.get('stream') else encode(response),
                   'text/event-stream' if request.get('stream') else 'application/json')

def client():
    for size, chunk in [(8000, 1), (8000, 1024), (220000, 1024)]:
        conn = http.client.HTTPConnection('127.0.0.1', 18002, timeout=30)
        conn.request('POST', '/v1/responses', encode(dict(model=f'fixture-{size}-{chunk}',
                     input='size reproduction', stream=True, store=True)), {'Content-Type':'application/json'})
        response = conn.getresponse()
        raw = response.read()
        conn.close()
        events = [json.loads(line[5:]) for line in raw.splitlines()
                  if line.startswith(b'data:') and line[5:].strip() != b'[DONE]']
        print(json.dumps({'size':size, 'chunk':chunk, 'http':response.status,
              'received_text_bytes':sum(len(e.get('delta','')) for e in events if e.get('type')=='response.output_text.delta'),
              'terminal':[e['type'] for e in events if e['type'] in ('response.completed','response.failed','response.incomplete')],
              'errors':[e for e in events if e['type']=='error']}))

if __name__ == '__main__':
    if sys.argv[1:] == ['server']:
        ThreadingHTTPServer(('127.0.0.1', 18001), Upstream).serve_forever()
    elif sys.argv[1:] == ['client']:
        client()
    else:
        raise SystemExit('Use: python public_repro.py server|client')
```

</details>

Additional controls in the full harness: non-streaming JSON, HTTP pass-through with `store: false`, WebSocket `response.create`, exact JSON-body boundary and boundary + 1, and a healthy request on the same WebSocket after a failed request.

### Expected behavior

- An otherwise identical retained response should not fail solely because the upstream sends more, smaller deltas or repeats the same text in completion snapshots.
- Keep independent, configurable limits on actual retained response data, each upstream JSON body/SSE line, and each client stream event. Keep shared limits across inference rounds, gateway tool results, and MCP discovery, plus bounded queues/backpressure.
- Raising one allowance must not silently disable the others. Exceeding a resource limit should produce an explicit resource-specific error, terminate the affected request, and avoid publishing a successful durable response or transient checkpoint.
- Preserve #257's transient/durable continuation rules independently of per-request response resource accounting.

### Actual behavior

The same 14-case fixture set was run against these source revisions:

| Build | Completed | Size failures |
| --- | ---: | ---: |
| `3bb55fdf55ea1a7e756cf79694b187909faf62b7`, parent of #240 | 14 | 0 |
| `e87eeae1a8ca09156ececbf18d04487d80324701` | 8 | 6 |
| `2509342dcbc12b89350cabdda17f7bb1a6b8a41b`, including #257 | 8 | 6 |
| Local response-resource patch applied on `2509342` | 14 | 0 |

The six baseline failures are four HTTP cases (fine SSE, repeated-snapshot SSE, oversized SSE line, JSON body above the fixed cap) and two WebSocket cases (fine deltas and repeated snapshots).

1. **8,000 answer bytes / one-byte deltas:** each relevant delta line is only 137 bytes; the largest fixture line is 8,364 bytes. After delivering 7,655 answer bytes, the budget has charged 1,048,445 bytes. The next line would increase this to **1,048,582**, exceeding 1,048,576. HTTP SSE and WebSocket fail at that same predicted point. With 1,024-byte chunks, the identical answer uses 42,775 cumulative data-line bytes and completes. The full fine-grained fixture would use 1,128,619 bytes.
2. **220,000 answer bytes / 1,024-byte deltas:** all answer deltas arrive, but the terminal `response.completed` snapshot increases cumulative accounting from 910,273 to **1,130,636** bytes and is rejected. Every individual line remains below 256 KiB. JSON with the same answer succeeds.
3. **270,000 answer bytes:** a full-content done event exceeds the distinct **262,144-byte upstream SSE-line cap**. HTTP pass-through succeeds.
4. **JSON-body boundary:** exactly **1,048,576 upstream bytes** succeeds; one more byte returns HTTP 500 (`upstream response exceeded 1048576 bytes`). HTTP pass-through succeeds.

For the cumulative SSE failure, HTTP has already returned 200; the stream ends with an `error` event (`status: 500`, `code: "server_error"`, message containing `executor response budget exceeded 1048576 bytes`) and `[DONE]`. It has no `response.completed`, `response.failed`, or `response.incomplete` terminal event. The failed responses are not stored. These are resource errors, not token-limit truncation.

### Environment

- Native Windows execution; Rust **1.98.1** (`48a229cea`), Cargo **1.98.1** (`797e8a9bc`), each revision's `Cargo.lock`, `--locked`.
- The pre-#240 parent reports package **0.5.0**; the two tested main revisions report **0.6.0**. Source commits are necessary to distinguish them.
- Python **3.12.14**; standard-library HTTP upstream/client; **websockets 15.0.1** for WebSocket controls; isolated SQLite databases.
- No vLLM server, model, LiteLLM, or GPU was used in this reproduction. This isolates the gateway behavior; it is not a live-model performance or memory benchmark.
- Native-build caveat: each baseline needed the same unrelated borrow correction in `config.rs`: after Windows replaces path separators and obtains a `String`, pass a borrowed string to `utf8_percent_encode`. No budget, parser, or storage behavior was changed in the baseline builds. The local response patch contains a Windows-only `as_str()` binding for this correction.

### Additional context

**Source cause.** At `2509342`, [`response_budget.rs`](https://github.com/vllm-project/agentic-api/blob/2509342dcbc12b89350cabdda17f7bb1a6b8a41b/crates/agentic-server-core/src/executor/response_budget.rs) still initializes a 1 MiB atomic allowance that only decreases. [`fetch_stream_payload`](https://github.com/vllm-project/agentic-api/blob/2509342dcbc12b89350cabdda17f7bb1a6b8a41b/crates/agentic-server-core/src/executor/upstream.rs) charges `line.len()` before semantic accumulation. It counts the normalized `data: ` prefix and serialized event data, including repeated content, rather than current retained output or queued bytes. Framing newlines and `[DONE]` are not charged. [`inference.rs`](https://github.com/vllm-project/agentic-api/blob/2509342dcbc12b89350cabdda17f7bb1a6b8a41b/crates/agentic-server-core/src/executor/inference.rs) separately caps upstream lines and JSON bodies. Increasing only an outbound WebSocket event cap cannot fix the cumulative failure.

**Effect of #257 / `2509342`.** The new commit honors `store: false` and adds bounded connection-local checkpoints. It does not modify the cumulative response budget or upstream wire caps. A separate seven-check baseline session run reproduced the unstored fine-delta failure while coarse output, transient continuation, and durable restart controls passed. Its checkpoint ceilings (16 MiB per checkpoint, 32 MiB aggregate serialized retained checkpoints) address a different resource and should remain separate. In particular, cached/pinned/prepared checkpoint accounting should not be replaced by the per-request fix below.

**Implemented and tested local approach (proposed defaults, not current upstream settings):**

| Resource | Proposed default | Config field in `[responses]` | Environment override |
| --- | ---: | --- | --- |
| Logical retained response data across the turn | 8 MiB | `max_retained_bytes` | `AGENTIC_MAX_RETAINED_RESPONSE_BYTES` |
| One upstream JSON body | 16 MiB | `max_upstream_json_bytes` | `AGENTIC_MAX_UPSTREAM_JSON_BYTES` |
| One upstream SSE line | 16 MiB | `max_upstream_sse_line_bytes` | `AGENTIC_MAX_UPSTREAM_SSE_LINE_BYTES` |
| One serialized client stream event | 1 MiB | `max_stream_event_bytes` | `AGENTIC_MAX_STREAM_EVENT_BYTES` |

- Positive, validated limits; environment overrides TOML independently. No lifetime streamed-byte limit. Raising defaults alone is insufficient: retained accounting replaces the previous item-state charge as deltas/completions update that state, rather than charging all wire snapshots.
- Keep the existing single normalization/ingestion path. Wire reads are bounded in `inference.rs`; retained output is measured in typed accumulator slots; stream delivery validates its own serialized event size; the engine owns the shared multi-round/tool/discovery allowance and persistence order.
- The prototype's retained-output metric uses UTF-8 content and 32 bytes per serialized value/container, with active text/argument buffers measured by length. Strings are not rescanned on every delta. Raw gateway tool output and serialized MCP discovery retain their existing byte measurements. This is **logical resource accounting, not an RSS/allocator bound**; transient parsing/serialization, execution copies, and allocator capacity require separate consideration. The configurable values are policy choices to review, not benchmark-derived safe memory ceilings.
- Preserve bounded delivery queues and cancellation. Retain the 1 MiB default client-event bound; reserve WebSocket routing overhead before core terminal validation so a response is not stored and then rejected only after `stream_id` is attached. `generate: false` validates both events before commit.
- Use error code `response_resource_limit_exceeded` with the resource/limit. Non-streaming executor failures return 502. Started streams emit `response.failed` when the diagnostic fits, otherwise a bounded `error` diagnostic. Reject terminal serialization before durable persistence or checkpoint publication; keep healthy-request recovery and #257's failure/eviction semantics.

**Verification of the local patch on `2509342`:**

- **52 production-binary acceptance checks passed:** 14 original regression/control cases; 24 independent-limit, boundary, TOML/environment precedence, failure, recovery, and database-count checks; two tiny-event-limit diagnostic checks; 12 transient/durable session and restart checks.
- Session checks include an unstored fine-delta response, same-connection continuation, a stored child of an unstored parent, full history restored after process restart, an unstored ID rejected after restart, and no cached/stored failed response. A failed same-lane continuation evicts its referenced parent; an independent request then succeeds.
- Eight new Rust unit tests passed: chunk/snapshot invariance, genuinely oversized active arguments, shared retained rounds, atomic boundary/overflow behavior, UTF-8/escape/structural accounting, retained JSON metadata, defaults/partial overrides, and invalid settings. Existing focused coverage also passed for MCP/tool aggregate limits, oversized-terminal non-persistence, lifecycle validation, backpressure/disconnect, and session budget/cancellation behavior.
- `cargo fmt --all -- --check`, `git diff --check`, and `cargo clippy --locked --workspace --all-targets --all-features -- -D warnings` passed.
- **Full native Rust suite is not green:** `cargo test --locked --workspace --all-features --no-fail-fast` reports the same **193 failing test cases on unpatched `2509342` and the patch**, with matching names/counts: 192 fail opening Windows temporary SQLite paths; one oversized-request transport test fails with a connection-abort error. There are no additional failing test cases from the patch. The production harness uses working relative SQLite URLs and verifies persistence separately. Linux CI and live-model validation remain to be done.

The patch is a locally verified proposal, not an upstream change or a claim of complete platform validation. The regression report and the proposed default sizes can be evaluated independently.

Zheng-Lu:
## Summary

Fixes #288. Fine-grained Responses streams previously exhausted the response budget by charging repeated SSE framing and completion snapshots. Charge cumulative logical retained output across inference rounds and tool outputs instead, using the existing shared JSON/SSE ingestion path.

Expose independent `[responses]` / environment limits: 8 MiB retained output and 16 MiB upstream JSON, upstream SSE lines, and client events by default, with validated wire headroom. Align WebSocket event limits and account for routing metadata before terminal validation, persistence, or checkpoint publication. Resource-limit failures use typed errors.

Close retained-accounting gaps for nested empty JSON values, unrestricted role/type/status strings, empty query entries, pending and late-bound item identities, and terminal error/incomplete details. Identity growth is charged before insertion. Reasoning and shell completion measure only the affected part around the existing completion operation; prior completed parts and static metadata are not rescanned on each event.

Keep lifecycle coordination in SlotMap, per-kind state/accounting together, wire limits in inference and transport, and delivery/persistence ordering in orchestration. Extract cohesive private modules to meet the production file-size ratchet without raising limits. Update ARCHITECTURE, configuration documentation, and release notes. Integrate current main, including Codex image capabilities.

## Test Plan

- Red/green regressions reproduced nested JSON, unrestricted strings, pending/late identity, and terminal-detail budget bypasses before the fixes.
- 49 focused budget tests passed, including JSON/SSE equivalence, repeated snapshots, sparse/empty reasoning parts, and shell lifecycle checks.
- Full integrated workspace tests passed with no failures. Standard ignored tests include eight PostgreSQL tests requiring `TEST_POSTGRES_URL`, one doctest, and two manual scaling probes; both probes were run explicitly.
- Debug scaling for 4k / 8k / 16k completions: reasoning improved from 139 / 444 / 1577 ms to 55 / 107 / 215 ms; shell improved from 231 / 570 / 1700 ms to 114 / 197 / 402 ms. These are local scaling measurements, not production latency guarantees or timing-based test assertions.
- Strict workspace/all-target Clippy and all pre-commit hooks passed. Existing recorded replay tests passed; no cassettes were edited or fabricated.
- Full pinned Codex 0.149.1 smoke passed after integration: HELLO, both image launcher modes, and text-only negative control with exact upstream capture.
- Independent review and scoped re-review found no remaining actionable issues.


crates/agentic-server-core/src/executor/engine.rs
```diff
@@ -563,7 +570,12 @@ fn run_stream(
         let (event_tx, mut event_rx) = mpsc::channel(STREAM_EVENT_BUFFER);
         let exec_ctx_for_run = Arc::clone(&exec_ctx);
         let event_tx_for_run = event_tx.clone();
-        let mut agent = agent_pipeline(ctx, tool_search_state, Some(event_tx_for_run));
+        let mut agent = agent_pipeline_with_limits(
```
maralbahari, line 573: The core receives the new 16 MiB event limit, but WebSocket delivery still enforces a fixed 1 MiB limit.
Reproduced: two 600,000-byte messages delivered all 1,200,000 text bytes, then returned websocket event exceeded 1048576 bytes instead of `response.completed`. The response was already persisted.

Fix: pass the effective transport limit, including routing metadata overhead, into terminal validation before persistence or checkpoint publication. Add a WebSocket regression covering this exact case.

crates/agentic-server-core/src/executor/response_budget.rs
```diff
@@ -2,31 +2,296 @@ use std::sync::Arc;
...
             })
     }
 }
+
+pub(in crate::executor) fn retained_output_item_bytes(item: &OutputItem) -> usize {
+    match item {
+        OutputItem::Message(msg) => {
+            let mut bytes = RETAINED_CONTAINER_OVERHEAD_BYTES + msg.id.len();
+            for part in &msg.content {
+                bytes += RETAINED_CONTAINER_OVERHEAD_BYTES + part.text.len();
```
maralbahari, line 59: `retained_output_item_bytes` ignores message annotations and compaction `encrypted_content`, while nested tool-search arguments receive only a shallow estimate. Final reconciliation uses the same estimate, so it cannot catch these omissions.

I reproduced this with `max_retained_bytes = 4096`: a message containing 100,000 bytes of annotation text and a compaction item containing 100,000 bytes both completed successfully through JSON and streaming.

Please include all variable-sized retained fields, recursively account for nested JSON, and add regression tests asserting these responses exceed the configured budget.

crates/agentic-server-core/src/executor/accumulator/slot.rs
```diff
@@ -475,102 +562,241 @@ impl ActiveItem {
...
-            },
+        }
+    }
+
+    fn apply_message(parts: &mut IndexMap<u32, StreamedPart>, payload: &EventPayload) -> usize {
+        match payload {
+            EventPayload::TextDelta {
+                delta, content_index, ..
+            } => {
+                let part = parts.entry(*content_index).or_default();
```
maralbahari, line 614: Each distinct `content_index` inserts a `StreamedPart`, but only `delta.len()` is charged. Empty deltas therefore allocate map entries without consuming the retained-byte budget. Finalization discards empty parts, so reconciliation does not catch them.

I reproduced this with `max_retained_bytes = 4096`: a stream containing 100,000 distinct empty text deltas completed successfully.

Please charge structural overhead before inserting each new entry, including entries in the reasoning maps, and add a regression test verifying that excessive empty parts exhaust the configured budget promptly.

crates/agentic-server-core/src/executor/accumulator/slot.rs
```diff
@@ -475,102 +562,241 @@ impl ActiveItem {
...
+                let streamed = summary_streamed.remove(summary_index).unwrap_or(0);
+                let additional = text.len().saturating_sub(streamed);
+                item.apply_done(payload, &mut String::new());
+                additional
+            }
+            _ => 0,
+        }
+    }
+
+    fn apply_function_call(
```
maralbahari, line 676: The retained-byte accounting belongs in synchronous ingestion, but could we consolidate how it is implemented here? Some additional tracking is necessary to distinguish deltas from repeated completion snapshots. My concern is that the PR spreads the retention rules across several places that must remain consistent.

**There are two forms of duplication:**

1. **Completion behavior and accounting describe the same update separately.**  
   `apply_function_call` calculates `additional_args`, `additional_name`, and `additional_cid`, then calls `item.apply_done(payload, arguments)`, which independently decides how arguments and metadata are replaced or preserved. `apply_custom_tool_call` follows the same pattern. A change to completion semantics now requires a matching change to this accounting logic, with no compiler-enforced relationship between them.

2. **Budget charging and reconciliation are repeated across lifecycle operations.**  
   `open` computes and charges initial bytes, `apply` charges incremental bytes, and both `complete` and `drain_output_with_budget` compare the completed size against `slot.retained_bytes` and charge the difference. Meanwhile, `initial_added_retained_bytes` and `retained_output_item_bytes` describe different stages of the same retained item using separate field lists.

Active-state accounting and completed-item measurement cannot necessarily be identical, but they need a clear shared contract. The omitted annotations/compaction content and uncharged content-part entries identified elsewhere in this review show the consequences when those representations diverge.

**Could we refactor this along the existing architecture boundaries?**

- **Keep `SlotMap` responsible for identity and lifecycle coordination:** resolve indexes and IDs, enforce active/completed transitions, handle duplicate completion, and dispatch to the appropriate typed state.
- **Encapsulate each kind’s active state and update logic together.** For example, message state would own its parts and their accounting; function-call state would own its argument buffer and completion tracking. These can live in a private module under `accumulator/`, with `ActiveItem` retaining exhaustive enum dispatch.
- **Keep growth calculation and mutation in the same per-kind operation.** That operation should check the required budget before growing retained state. It should account for both content and newly created containers, including empty parts. Avoid maintaining a separate event-based approximation of what `ApplyDone` will do.
- **Preserve one completion policy.** Reuse the existing `ApplyDone`/`MergeDone` behavior, or refactor their private integration so accounting observes the same completion decision. Please avoid adding another implementation of fallback, replacement, or snapshot-deduplication semantics.
- **Extract one small accounting primitive** for charging growth and reconciling a slot’s previously accounted bytes with its completed retained size. Use it consistently from opening, updates, completion, and finalization.
- **Make completed-item measurement comprehensive**, including nested variable-sized fields. Final reconciliation should verify incremental accounting against that measurement.

Please preserve incremental accounting for deltas: measuring or serializing the entire growing item after every event would introduce repeated scans and potentially quadratic work. Keep budget policy within the executor/accumulator boundary, and retain the existing shared JSON/SSE ingestion path.

I would also add tests for equivalent delta/done-only representations, repeated completion snapshots, empty multipart entries, metadata supplied at completion, and rejection before retained state grows beyond the budget. The objective is to make each retention rule have one clear owner, so adding or changing an item field does not require updating several independent interpretations of its lifecycle.

crates/agentic-server-core/src/executor/engine.rs
Zheng-Lu: Fixed in `b33431e`. 

- **Aligned WebSocket limit**: Replaced the hardcoded 1 MiB constant with `WsEventLimit`, derived directly from `max_stream_event_bytes`.
- **Pre-persistence terminal validation**: Calculated the exact `stream_id` JSON routing overhead (plus re-serialization slack) and passed the effective transport limit into the executor via `ExecuteRequest::with_max_stream_event_bytes`. In `engine.rs`, the terminal `response.completed` is now validated against this effective limit before `persist_if_needed` or checkpoint publication. If it exceeds the limit, it aborts without writing any rows to storage.
- **Added regression tests** in `responses_websocket_test.rs`:
  - `websocket_delivers_a_terminal_event_larger_than_the_old_fixed_socket_limit`: verifies delivery of >1 MiB terminal frames under the aligned ceiling.
  - `websocket_undeliverable_terminal_event_is_rejected_before_persistence`: verifies that an undeliverable terminal frame returns an error event and asserts 0 rows are persisted in `responses` and `items`.

crates/agentic-server-core/src/executor/response_budget.rs
Zheng-Lu: Fixed in `b33431e`.

- **Comprehensive recursive measurement**: Introduced the `RetainedSize` trait across all output items and `serde_json::Value`. Arrays and objects are now measured recursively with container overhead.
- **Included omitted variable fields**:
  - `OutputTextContent`: recursively counts `annotations`.
  - `CompactionItem`: counts `encrypted_content.len()`.
  - `ToolSearchCall`: recursively measures `arguments` as JSON.
  - Also covers `ShellCall` `extra` maps, and MCP error/schema annotations.
- **Added regression tests** in `budget_tests.rs`:
  - `oversized_annotations_are_rejected_on_both_ingestion_paths` (JSON & stream)
  - `oversized_compaction_content_is_rejected_on_both_ingestion_paths` (JSON & stream)
  - `oversized_tool_search_arguments_are_rejected_recursively`

crates/agentic-server-core/src/executor/accumulator/slot.rs
Zheng-Lu: Fixed in `b33431e`.

- **Pre-charged container overhead**: In `active.rs`, `RETAINED_CONTAINER_OVERHEAD_BYTES` is now charged *before* inserting a new entry into `parts` (message) or the reasoning content/summary index maps. Empty deltas (`""`) are no longer free.
- **Added regression test**:
  - `empty_multipart_entries_are_charged_and_exhaust_the_budget_promptly`: asserts that 100,000 empty parts exhaust a 4 KiB budget immediately at the exact affordable threshold.

crates/agentic-server-core/src/executor/accumulator/slot.rs
Zheng-Lu: Refactored in `b33431e` following your suggested architectural boundaries:

1. **Encapsulated active state**: Extracted `accumulator/active.rs`. Each kind (`MessageState`, `FunctionCallState`, etc.) now owns its streaming buffer, growth check, and update logic. `SlotMap` now strictly owns identity resolution, lifecycle transitions, and dispatch.
2. **Unified accounting primitive (`RetainedAccount`)**:
   - `charge()`: for growth known in advance (deltas, new container entries). Checked before the buffer/map grows.
   - `grow()`: wraps the single `ApplyDone`/`MergeDone` invocation, measuring state before and after. Accounting now directly observes the completion policy rather than maintaining a parallel approximation.
   - `reconcile()`: reconciles the slot's accounted bytes with the completed `RetainedSize` measurement without refunding.
3. **Preserved incremental delta accounting**: Deltas are still charged by incremental length without re-scanning or serializing full items.
4. **Added comprehensive test suite** (`accumulator/budget_tests.rs`):
   - Delta vs done-only equivalence (`delta_and_done_only_streams_charge_the_same_retained_bytes`)
   - Repeated completion snapshot idempotency (`repeated_identical_completion_snapshots_are_free`)
   - Metadata supplied only at completion (`metadata_supplied_only_at_completion_is_charged_at_completion`)
   - Rejection before buffer mutation (`a_delta_is_rejected_before_the_retained_state_grows`)

maralbahari:
@Zheng-Lu please update the `ARCHITECTURE.md` based on the refactoring work.

crates/agentic-server-core/src/executor/response_budget.rs
```diff
@@ -1,32 +1,564 @@
...
+}
+
+impl RetainedSize for Value {
+    /// Approximates the serialized footprint of arbitrary JSON without serializing it.
+    fn retained_bytes(&self) -> usize {
+        match self {
+            Value::Null => 4,
+            Value::Bool(_) => 5,
+            Value::Number(_) => 8,
+            Value::String(text) => text.len(),
```
maralbahari: `Value::String("")` currently contributes zero bytes, while `Value::Array` charges only one container plus its children’s measurements. An array containing 100,000 empty strings therefore measures just 32 bytes, despite retaining 100,000 `Value` entries.

this can be reproduced through both JSON and streaming: an annotation containing that array completed successfully with `max_retained_bytes = 4096`. The returned output was approximately 300 KB.

Please charge structural overhead per retained JSON value or collection entry, including empty strings, and add regression tests for this nested-array case on both ingestion paths. Final reconciliation currently uses the same measurement, so it cannot catch the bypass.

crates/agentic-server-core/src/executor/response_budget.rs
```diff
@@ -1,32 +1,564 @@
...
+    }
+}
+
+impl<T: RetainedSize> RetainedSize for Option<T> {
+    fn retained_bytes(&self) -> usize {
+        self.as_ref().map_or(0, RetainedSize::retained_bytes)
+    }
+}
+
+impl RetainedSize for OutputTextContent {
```
maralbahari, line 183: The `RetainedSize` documentation excludes `type`, `role`, and `status` because they are supposedly bounded by enums. However, fields such as `OutputMessage.role` and `OutputTextContent.type_` are unrestricted `String`s, and the live executor’s lenient ingestion accepts and retains arbitrary values.

this can be reproduced through both JSON and streaming: a message with a 100,000-byte `role`, and separately a content part with a 100,000-byte `type`, each completed successfully with `max_retained_bytes = 4096`.

Please include unrestricted strings in retained measurements, or enforce a bounded vocabulary during deserialization. Audit the other excluded fields for the same assumption and add regression tests on both ingestion paths. Accounting should only omit fields whose bounds are enforced by their owning types.

crates/agentic-server-core/src/executor/accumulator/active.rs
```diff
@@ -0,0 +1,581 @@
...
+    ) -> ExecutorResult<()> {
+        match payload {
+            EventPayload::ReasoningTextDelta {
+                delta, content_index, ..
+            } => count_streamed(&mut self.content_streamed, *content_index, delta, account, budget),
+            EventPayload::ReasoningSummaryTextDelta {
+                delta, summary_index, ..
+            } => count_streamed(&mut self.summary_streamed, *summary_index, delta, account, budget),
+            EventPayload::ReasoningTextDone { content_index, .. } => {
+                account.grow(budget, self, Self::text_retained_bytes, |state| {
```
maralbahari: Each `ReasoningTextDone` calls `RetainedAccount::grow` with `text_retained_bytes`, which scans all accumulated content, summaries, and tracking maps before and after the mutation. `ReasoningSummaryTextDone` follows the same pattern.

Completing N separate parts therefore performs O(N²) accounting work, even when each completion simply appends a small part. This processing happens synchronously in ingestion.

A synthetic reproduction with sequential one-byte reasoning parts showed:

- 4,000 parts: 98 ms previously → 302 ms after this change.
- 8,000 parts: 197 ms previously → 956 ms after this change.
- 16,000 parts: 394 ms previously → 3,421 ms after this change.

These are debug-build measurements, including parsing; they demonstrate the scaling regression rather than production latency.

Please measure only the affected part, or maintain cached totals updated through the existing completion operation. Keep completion behavior in one place so accounting does not duplicate its reconciliation rules, and reserve full-item measurement for final reconciliation.

maralbahari (dismissed):

franciscojavierarceo:
the latest accounting blockers are still present in `ea3c6c1`: an array of empty JSON strings is charged only one container, `OutputMessage.role` and `OutputTextContent.type_` are omitted despite being unrestricted strings, and each reasoning-part completion rescans the accumulated state. the first two leave retained-memory budget bypasses; the last adds quadratic work to synchronous ingestion.

we should address the existing review threads with JSON/SSE regression coverage, keep completion accounting incremental, and update `ARCHITECTURE.md` before merging. the current pre-commit failure also needs to be resolved.

franciscojavierarceo (approved):
the requested nested-JSON, unrestricted-string, and reasoning completion fixes are in 1e920f3. the follow-up audit also closed pending/late-bound ID and terminal-detail budget gaps and removed shell completion rescans. the fixes keep the existing ingestion/completion path and reduce the file-size baselines.

the integrated workspace suite, 49 focused budget tests, strict Clippy, pre-commit, and the pinned Codex smoke pass locally. independent re-review found no remaining blockers; hosted CI is running on the pushed head. @maralbahari we should re-review this head to clear the outstanding change request before merging.

PR #311, opened 2026-09-15
fix(mcp): follow tools/list pagination cursors

Zheng-Lu:
## Summary

Follow MCP `tools/list` pagination so discovery includes tools beyond the first page.

- Preserve `list_tools(None)` for the first request and pass each subsequent cursor through unchanged, including empty strings: cursors are opaque tokens.
- Finish only when `nextCursor` is absent. Return typed `PaginationLoop` for a repeated cursor and `PaginationExceeded` after 32 pages instead of returning a partial list.
- Apply the existing 60-second timeout to the complete traversal; propagate failures from any page.

The explicit loop preserves first-page wire compatibility and adds termination guards missing from `rmcp 1.8.0`'s `list_all_tools()`.

## Test Plan

- Deterministic Python stdio MCP server exercises multi-page discovery, exact empty-cursor forwarding, repeated-cursor errors, and runaway pagination.
- The empty-cursor and repeated-cursor regressions failed before the review fix and pass afterward.
- All 9 MCP client tests passed after the final patch; core library: 688 passed, 3 PostgreSQL-gated tests ignored.
- `cargo clippy -p agentic-server-core --all-targets -- -D warnings` passed.
- `pre-commit run --all-files` passed.


franciscojavierarceo (approved):
patched empty-cursor forwarding and repeated-cursor errors, with regressions verified before and after the fix. all 688 core tests, Clippy, and pre-commit checks pass locally.

Issue #291, opened 2026-09-12
[RFC] Pluggable Web Search Providers and Typed Result Normalization


Zheng-Lu:
### Summary

The gateway-owned `web_search_preview` tool is hardwired to You.com. This RFC proposes making the search
backend selectable through a typed `WebSearchProviderKind` config enum, adding **Brave Search** as the first
alternative provider, and — as a prerequisite — replacing the current untyped pass-through of provider JSON with
a typed normalization contract (`WebSearchResult`). The work is split into two PRs: a behavior-preserving refactor,
then the Brave provider with configuration, tests, and docs. SearXNG (keyless, self-hosted) and Tavily follow as
separate PRs against the same contract.

The design keeps the existing `WebSearchProvider` trait private, keeps the public `WebSearchHandler` constructors
unchanged, and adds only additive configuration. When no provider is selected, behavior is identical to today.

### Motivation

`web_search_preview` is executed by the gateway (`ARCHITECTURE.md`, `docs/design/tool-framework.md`), which is a
real differentiator for a self-hosted, OpenAI-compatible stack. Today it can only be enabled by setting
`YOU_API_KEY` and `YOU_API_BASE_URL`
(`crates/agentic-server-core/src/tool/web_search.rs:20-21`, `docs/deploying/kubernetes.md:419-420`).

PR #85 (@franciscojavierarceo) anticipated this:

> Splits the You.com HTTP integration behind an internal `WebSearchProvider` abstraction so additional search
> providers can be added without changing the OpenAI tool adapter.

The abstraction exists (`web_search.rs:253-260`), but nothing else was ever plugged into it. Practical consequences:

- **Barrier to entry.** As of this writing, You.com's Search API is a paid product without a self-serve free
  developer tier. Brave Search offers a free plan (≈2,000 queries/month), Tavily ≈1,000/month. A personal or
  evaluation deployment of Agentic API cannot exercise `web_search` at all without a commercial contract.
- **Air-gapped / private-cloud deployments** cannot use `web_search` because the only supported backend is a public
  SaaS endpoint. A SearXNG instance (or an internal search API behind the same contract) is the realistic answer for
  those environments.
- **Vendor coupling in the tool contract.** The function schema shown to the model, the argument struct, and the
  model-facing tool output are all You.com-shaped (details below), so even an embedder who implements their own
  executor inherits You.com semantics.

### Goals

1. Select the search backend by configuration; `you` remains the default.
[...]

PR #321, opened 2026-09-17
feat(tool): add Brave Search as a selectable web_search provider

### LINKED ISSUE #294: [Feature] Add Brave Search as a selectable web_search provider

**Part of Phase 2 of #291** 
**Status**: Blocked on Phase 1 #293 
**Scope**: Brave Search integration, configuration plumbing, and tests

With the typed contract established in Phase 1 #293, `web_search_preview` still only has one backend (You.com).  
**Brave Search** is the ideal first alternative:
- High-quality search results with web and news combined in a single request.
- Free developer tier (≈2,000 queries/month) allowing local and evaluation deployments without paid contracts.
- Differences from You.com that must be explicitly adapted: no server-side domain filtering, `count` capped at 20, distinct freshness syntax, and a free tier rate limit of ~1 QPS.

#### A. Provider Implementation (`tool/web_search/brave.rs`)
- Endpoint: `GET {base_url}/res/v1/web/search` with `result_filter=web,news` and `X-Subscription-Token` header.
- Deserialization: Minimal structs with `#[serde(default)]` and **without** `deny_unknown_fields` (forward-compatible).
- **Transport Rule**: Do **not** send `Accept-Encoding: gzip` (core `reqwest` is built without gzip support in `crates/agentic-server-core/Cargo.toml:26`).
#### B. Input Adaptation Policy Matrix
| Parameter / Input | You.com Behavior | Brave Search Behavior | Gateway Adaptation Policy |
| :--- | :--- | :--- | :--- |
| **`allowed_domains` / `blocked_domains`** | Native server-side filter | **No server-side filter** | Post-filter results client-side by host suffix on label boundary using `url::Url`. An allowlist is a hard contract; returning fewer than `count` is correct. |
| **`count`** | 1..=100 | 1..=20 | Clamp to 20 with `tracing::debug!`. Never error, as models cannot predict provider limits. |
| **`freshness`** | `day`/`week`/`month`/`year` | `pd`/`pw`/`pm`/`py` or date range | Parsed into typed `Freshness` enum in `args.rs` and rendered per provider. |
| **News** | Single request | Single request | Included in the same call via `result_filter=web,news`. |
| **`country` / `language`** | Native | `country` / `search_lang` | Mapped to Brave's corresponding query parameters. |
| **You.com specifics** (`livecrawl`, etc.) | Native | Not supported | Ignored with `tracing::debug!`; documented as You.com-specific. |
#### C. Concurrency & 429 Rate Limiting
- **Concurrency Ceiling**: Add `WebSearchProvider::max_concurrent_requests()` (Brave defaults to 1; You.com inherits gateway limit).
- **No Automatic 429 Retry in Phase 2**: Automatic retries risk masking tool timeouts (`executor/gateway.rs:263-274`). A 429 immediately returns a failed `web_search_call` containing the upstream `Retry-After` header.
#### D. Configuration & Precedence Matrix
| Setting | Environment Variable | `config.toml` Key | Default Value |
| :--- | :--- | :--- | :--- |
| **Provider** | `AGENTIC_WEB_SEARCH_PROVIDER` | `[web_search].provider` | `"you"` |
| **API Key** | Named by `api_key_env` | `[web_search].api_key_env` | `"BRAVE_API_KEY"` (for Brave) |
| **Base URL** | `AGENTIC_WEB_SEARCH_BASE_URL` | `[web_search].base_url` | `https://api.search.brave.com` |
| **Max Concurrency** | `AGENTIC_WEB_SEARCH_MAX_CONCURRENT_QUERIES` | `[web_search].max_concurrent_queries` | `1` (for Brave) |

*Note: `YOU_API_BASE_URL` remains fully supported for backward compatibility when `provider = "you"`.*

### Decisions on #291 Open Questions (Defaults)
| Question | Default Decision | Rationale |
| :--- | :--- | :--- |
| **Q2 (You.com Base URL Default)** | No default URL | Avoid changing existing behavior for You.com. Doc inconsistencies to be fixed in a separate `docs:` PR. |
| **Q3 (Base URL Env Name)** | `AGENTIC_WEB_SEARCH_BASE_URL` | Clean generic override for alternative providers; preserves `YOU_API_BASE_URL` for You.com. |
| **Q5 (Metadata)** | Include provider name | Surface provider name in `metadata[]` for visibility. |

### Scope Boundaries
| In Scope for Phase 2 | Explicitly Out of Scope (Deferred) |
| :--- | :--- |
| ✅ `BraveSearchProvider` implementation | ❌ SearXNG and Tavily (follow-up issues) |
| ✅ Domain post-filtering & count clamping | ❌ Query rewriting with `site:` operators |
| ✅ Configuration plumbing (env, toml, generated) | ❌ Automatic retry on HTTP 429 |
| ✅ Local mock HTTP tests (Axum) | ❌ Public plugin trait / dynamic registration |
| ✅ Documentation updates (`README`, `kubernetes.md`) | ❌ Pagination support |

### Verification Checklist
- [ ] **200 OK (Web + News)**: Verify mixed results properly map to `WebSearchResult` and public sources.
- [ ] **Empty Results**: Handled gracefully without errors.
- [ ] **401 / 403 Authentication**: Fails `web_search_call`, mentions the env variable, never leaks secrets.
- [ ] **429 Rate Limit**: Fails `web_search_call` without retry, surfaces `Retry-After` in message.
- [ ] **Domain Filtering**: Comprehensive test covering allowlist, blocklist, subdomains, and label boundaries.
- [ ] **Query Parameter Clamping**: `count` clamped from 50 → 20; `freshness` converted to Brave syntax.
- [ ] **Concurrency Limit**: Multi-query batch respects `max_active == 1` under Brave.
- [ ] **Backward Compatibility**: With `provider` unset, You.com behaves identically to Phase 1.
- [ ] **Code Hygiene**:
  - `cargo clippy --all-targets -- -D warnings` (clean)
  - `cargo fmt -- --check` (clean)
  - `pre-commit run --all-files` (clean)

LeaderOnePro:
## Summary

Closes #294 (Phase 2 of #291).

Add Brave Search as a selectable `web_search` provider alongside You.com, selected via
`AGENTIC_WEB_SEARCH_PROVIDER` (env) > `[web_search].provider` (config file) > default `"you"`.

**Provider behavior** (`crates/agentic-server-core/src/tool/web_search/brave.rs`):

- `GET /res/v1/web/search` with `X-Subscription-Token` auth and `result_filter=web,news`,
  normalized into the existing You.com-shaped model-facing wire (no change to the output
  schema: Brave results serialize through the same struct as You.com results).
- `count` clamped to the free-tier 20-results-per-section cap (with a debug log); `freshness`
  mapped to Brave's format.
- No server-side domain filtering: `include_domains`/`exclude_domains` are post-filtered
  client-side with label-boundary matching, IDNA normalization, trailing-dot handling, and
  fail-closed behavior on unparseable URLs.
- At most one request in flight: providers declare a ceiling via `WebSearchProvider::max_concurrent_requests`,
  and `cap_provider_concurrency` folds it into the gateway-wide limit (hard 1 for Brave —
  the ~1 QPS free-tier limit is service-imposed, not an operator knob).
- 429 responses surface `Retry-After` verbatim (the header may also be an HTTP-date, so no
  unit is appended); error bodies are read through the existing size-limited reader.

**Configuration plumbing**:

- `api_key_env` now falls back to the selected provider's default variable (`YOU_API_KEY` /
  `BRAVE_API_KEY`) when unset, and the generated config file no longer hardcodes
  `YOU_API_KEY` — switching providers never requires hand-editing the config file.
- `AGENTIC_WEB_SEARCH_PROVIDER` is parsed through the same serde wire names the config file
  accepts (`WebSearchProviderKind::parse_name`, strum-backed `VARIANTS`), trimmed and
  case-insensitive, with the variant list in the error derived rather than hand-maintained.
- You.com path is unchanged when no provider is configured (base URL override stays
  `YOU_API_BASE_URL`; Brave uses `AGENTIC_WEB_SEARCH_BASE_URL`).

**Docs**: Brave configuration section in `docs/deploying/kubernetes.md`; README documents the
`api_key_env` fallback and the cross-provider include/exclude semantics difference
(You.com rejects the combination, Brave applies the blocklist on top of the allowlist).

**Dependency note**: `strum` is added solely for `VariantNames`/`EnumIter` on the
2-variant `WebSearchProviderKind`, keeping the config error message's variant list in
lockstep with what serde actually accepts (enforced by a round-trip test). A hand-written
`VARIANTS` array would avoid the dependency but reintroduces the drift this exists to
prevent; the lockfile adds exactly two crates, `strum` and `strum_macros` — `heck`, `syn`,
`quote`, and `proc-macro2` are already in the tree via other derives.

### Deviations from the issue

- **Provider name in `metadata[]`**: the issue's Q5 default decision was to surface the
  provider name in metadata. This PR defers that — `WebSearchProviderMetadata.provider`
  remains wire-invisible so the model-facing payload schema is unchanged (no fields added
  or removed); surfacing it is an additive wire change that recorded replays and strict
  clients would still need to absorb, so it is better landed deliberately as its own change.
- **`AGENTIC_WEB_SEARCH_MAX_CONCURRENT_QUERIES`**: not implemented. Brave's ~1 QPS guidance
  is a service-side limit, and a tunable knob would only invite 429s. The provider-ceiling
  plumbing (`max_concurrent_requests` + `cap_provider_concurrency`) is in place if a future
  provider has a genuinely configurable budget.

## Test Plan

- `uv tool run --from pre-commit pre-commit run --all-files` — all hooks pass.
- `cargo fmt --all -- --check` and `cargo clippy --all-targets -- -D warnings` — clean.
- `cargo test --workspace` — 64 suites, 0 failures (including 697 tests in
  `agentic-server-core` lib and the 30-test `web_search_tool_test` integration suite).
- `scripts/check_rust_file_sizes.py` — all 134 production files within the 500-line policy.
- New coverage for this PR: Axum mock integration tests for 200 (web+news), empty results,
  401, and 429 + `Retry-After` propagation; `DomainFilter` label-boundary/IDNA/fail-closed
  cases; an end-to-end concurrency test asserting (via in-flight atomic counters on the mock
  server) that a gateway limit of 5 still serializes Brave to exactly 1 concurrent and
  exactly 3 total requests; `parse_name` trimmed/case-insensitive and round-trip tests;
  api-key-env fallback tests. All provider interactions run against local Axum mocks — no
  replay cassettes or live-provider calls are involved.
- Review fixes applied on this branch (api-key fallback bug, verbatim `Retry-After`,
  unreachable branch removal, serde-backed provider parsing, docs) were verified by multiple
  independent review passes, including a full re-review of the fix commit.

Signed-off-by: LeaderOnePro <LeaderOnePro@outlook.com>


franciscojavierarceo:
`Freshness::Range` is accepted and then dropped when building the Brave request, and `safesearch` is never forwarded. a date-limited search therefore becomes unrestricted, while `safesearch: "strict"` falls back to Brave’s default behavior. we should preserve both supported options and assert the actual HTTP query parameters in the tests.

this also overlaps with #324, which handles those options and the provider metadata/concurrency settings requested in #294. i’m preparing that implementation for merge; we should consolidate any remaining useful coverage there rather than land two competing provider implementations.

LeaderOnePro:
Thanks for the review! `safesearch` was an oversight on my part. On `Freshness::Range`: dropping it was deliberate — Brave's short codes (`pd`/`pw`/`pm`/`py`) have no date-range equivalent, so the filter is omitted with a debug log — but I agree the observable result is the same, and #324 handles ranges properly.

#324 is the more complete implementation, so I'm closing this in its favor. Thanks again!

PR #322, opened 2026-09-17
[Release] [CI] Add release github action

tjtanaa:
## Summary

Add a Docker Hub publishing workflow for release, nightly, and manually selected builds.

The workflow:

- is called directly by the crates release workflow after it creates a GitHub Release, because releases created with
  `GITHUB_TOKEN` do not emit a second workflow run;
- also publishes the corresponding image tag for releases created outside the crates release workflow;
- supports manual builds from a Git branch, Git tag, or commit SHA, with an optional image-tag override;
- builds the current `main` commit every day at 13:00 GMT+8 (05:00 UTC);
- publishes nightly builds as both the moving `nightly` tag and `nightly-<commit-sha>`;
- removes `nightly-<commit-sha>` tags whose Docker Hub `last_updated` timestamp is more than 30 days old;
- keeps Docker Hub credentials in the protected `docker-publish` GitHub environment; and
- restricts manual publication to repository maintainers or administrators; and
- hardens secret handling with whole-string input validation, derived-token masking, constrained API pagination,
  read-only GitHub permissions, and commit-pinned third-party actions.

Release sources use fully qualified Git tag refs, commit inputs require the full 40-character SHA, and nightly cleanup
discovers every expired tag before deleting any tags so Docker Hub's page-number pagination cannot skip results.

Repository maintainers must configure the `docker-publish` environment with:

- the `DOCKERHUB_USERNAME` secret;
- the `DOCKERHUB_TOKEN` secret, with image-push, tag-read, and tag-delete permissions; and
- the `DOCKERHUB_IMAGE` variable, containing the target Docker Hub repository such as `vllm/agentic-api`.

The `nightly` and `nightly-*` tags must remain mutable so the moving nightly tag can be updated and expired nightly
tags can be deleted.

| Trigger | Source built | Docker image tags |
|---|---|---|
| Daily at 13:00 GMT+8 (05:00 UTC) | Current `main` | `nightly` and `nightly-<full-SHA>` |
| Manual: Actions → Release container → Run workflow | Your specified branch, tag, or full SHA | If it matches current `main`: both nightly tags, plus any override. Otherwise: override or source-ref-derived tag |
| Successful **Release crates** run with `dry_run=false`, run from `main` | Newly created release tag | Release tag, such as `v0.4.0` |
| GitHub Release is published, including a prerelease | That release’s Git tag | Release tag |

**Additional conditions:**

- Manual publishing requires the initiating user’s role to be `maintain` or `admin`.
- Every publishing job must pass the `docker-publish` environment rules.
- If that environment permits only `main`, the standalone GitHub Release trigger is blocked because it runs from a tag. The crates-release handoff still works.
- Ordinary pushes to `main` do not trigger publication.
- Cleanup runs after a successful image push and removes `nightly-*` tags older than 30 days.

## Test Plan

- Parsed `.github/workflows/release-container.yml` successfully as YAML.
- Ran `git diff --check` successfully.
- Verified that all workflow lines comply with the repository's 120-character line limit.
- Verify a non-dry-run crates release invokes the reusable container workflow with its newly created Git tag.
- After configuring the `docker-publish` environment, manually run the workflow against a non-`main` commit and
  verify that the requested image tag is published.
- Manually run the workflow against the current `main` commit and verify that both `nightly` and
  `nightly-<commit-sha>` are published.
- Verify that `nightly-*` tags older than 30 days are removed while the moving `nightly` tag is retained.


franciscojavierarceo:
- the [`release: published` trigger](https://github.com/vllm-project/agentic-api/blob/bcfeeafd16c23c7471f417f1a88a5db00a0de291/.github/workflows/release-container.yml#L9-L10) won't run for normal project releases: `release-crates.yml` creates them with `github.token`, and those release events [don't trigger another workflow](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/trigger-a-workflow). we should explicitly invoke container publication from the existing release workflow.

- the [manual permission check](https://github.com/vllm-project/agentic-api/blob/bcfeeafd16c23c7471f417f1a88a5db00a0de291/.github/workflows/release-container.yml#L41-L50) rejects users with the Maintain role. GitHub returns `write` in `.permission` for maintainers, so `admin|maintain` never accepts them. we should check [`role_name`](https://docs.github.com/en/rest/collaborators/collaborators#get-repository-permissions-for-a-user), which preserves the assigned role. reproduced with a mocked Maintain response.

- the [cleanup loop](https://github.com/vllm-project/agentic-api/blob/bcfeeafd16c23c7471f417f1a88a5db00a0de291/.github/workflows/release-container.yml#L232-L246) deletes the current page before following `.next`. Docker Hub uses page-number pagination, so deleting tags shifts later results into earlier pages. with 150 expired tags, the mocked cleanup deleted 100 and left 50. we should collect candidates across all pages before deleting any.

- [`source_ref="$RELEASE_TAG"`](https://github.com/vllm-project/agentic-api/blob/bcfeeafd16c23c7471f417f1a88a5db00a0de291/.github/workflows/release-container.yml#L92-L93) passes an unqualified ref to checkout, which prefers a matching branch over a tag. if both exist, the release image contains the branch's code. we should use `refs/tags/$RELEASE_TAG` for release events.

- the [source validation](https://github.com/vllm-project/agentic-api/blob/bcfeeafd16c23c7471f417f1a88a5db00a0de291/.github/workflows/release-container.yml#L98-L102) accepts 7–40 hexadecimal characters, but the pinned checkout action recognizes only full 40-character SHAs. abbreviated SHAs become branch/tag lookups and fail checkout. we should resolve abbreviations before checkout or explicitly require full SHAs.

YAML parsing, Bash syntax checks, and `git diff --check` passed. live image publication wasn't exercised.

franciscojavierarceo (approved):

Issue #279, opened 2026-09-11
feat: add OpenTelemetry traces and metrics


franciscojavierarceo:
## Summary

Add opt-in OpenTelemetry (OTel) traces and metrics to vLLM Agentic API, with correlated local logs and OTLP export to a configurable Collector or compatible backend.

The gateway currently initializes a `tracing` log subscriber and emits scattered events, including request routing decisions. It lacks a consistent request trace and operational metrics for understanding whether failures or latency originate in inference, tool execution, state management, or client delivery.

The first milestone should make it possible to follow a request through the gateway, identify where time was spent, and distinguish successful execution from failures and disconnects—even after streaming HTTP headers have been sent.

### 1. OTel foundation

- Retain Rust `tracing` instrumentation and bridge spans through `tracing-opentelemetry`.
- Add opt-in OTLP trace and metric export with service identity, documented standard `OTEL_*` configuration, and configurable trace sampling.
- Keep SDK/provider initialization and exporter lifecycle in the server. Core instrumentation must work with providers supplied by embedding applications and must not initialize a global subscriber.
- Preserve local logs and include trace/span correlation where a request context exists. Keep local log filtering independent of trace sampling.
- Bound export buffering and record size, define overflow behavior, and ensure exporter outages do not fail requests or stall client delivery.
- Flush/shut down providers after request draining, with a bounded deadline.
- Verify dependency compatibility with the repository's Rust 1.85 minimum.

### 2. Request and execution traces

- Extract incoming W3C trace context and inject the active context into upstream inference requests.
- Cover both pass-through and executor paths across HTTP and WebSocket transports, including Responses and Messages execution where applicable.
- Add child spans for rehydration, inference rounds, gateway-executed built-in tools, compaction, and persistence.
- Preserve parentage across asynchronous tasks and streamed body polling.
- Keep streaming request spans alive through body completion or disconnect; represent execution and delivery outcomes separately when persistence or other work continues after a disconnect.
- Give each WebSocket request its own execution span rather than treating an entire connection as one request.
- Record typed failure categories and terminal outcomes; HTTP 200 alone must not imply execution success.

### 3. Metrics and operational validation

- Request counts, active requests, durations, and execution/delivery outcomes.
- Stage durations, inference-round counts, and token usage when upstream usage is available.
- Explicit timing definitions for first upstream data, first client-visible semantic event, and first text output where observable. Do not label all of these as time to first token.
[...]

PR #341, opened 2026-09-18
feat: OpenTelemetry foundation (#338 Phrase 1 of #279)

### LINKED ISSUE #338: feat: OpenTelemetry foundation — opt-in OTLP export and correlated logs (#279 phase 1)

Parent: #279 (phase 1 of the three-PR sequence proposed in the design comment on #279).

Make telemetry a first-class, **off-by-default** capability of the `agentic-server` binary: SDK/provider lifecycle, OTLP/HTTP export, the `tracing` → OpenTelemetry bridge, trace-correlated local logs, bounded export buffering, and deadline-bounded shutdown. Ship one root `http.server.request` span and the two HTTP semantic-convention metrics so the pipeline is verifiable end-to-end against a Collector. Execution-level spans and project metrics follow in phases 2 and 3.

A prototype of the config/lifecycle/subscriber half already exists on a local branch (typed `OTEL_*` config, providers, bridge, log correlation, 18 unit + 5 integration tests against an OTLP/HTTP stub); this issue covers finishing it into a mergeable PR.

### Proposed solution

1. Land the existing prototype (config, lifecycle, subscriber, tests) and fold in review follow-ups: gate `CorrelatedFormat` on bridge presence so disabled mode does no per-event work; make the exporter's TLS feature explicit instead of relying on unification with `rmcp-reqwest`.
2. Add `TelemetryLayer` + body wrapper + request guard; wire it in `build_router_with_auth`.
3. Integration tests against an in-process OTLP/HTTP stub decoding protobuf: service identity on traces and metrics; root span for `/v1/responses` (proxy route) with the attribute allow-list; body-drop closes the span and finalizes the gauge; unresponsive/slow Collector does not affect request latency; shutdown honours the deadline.
4. Docs page; note the `Runtime::shutdown_timeout` behaviour change in the PR description.

#### Acceptance criteria (maps to #279 phase 1)
- [ ] With no exporter selected (or `OTEL_SDK_DISABLED=true`) no provider/exporter/thread is created and a test proves no connection is made to the configured endpoint.
- [ ] With `otlp` selected, a Collector/stub receives spans and metrics whose resource carries `service.name` (default `agentic-api`), `service.version` (crate version), and merged `OTEL_RESOURCE_ATTRIBUTES`.
- [ ] `grpc`/`http/json` protocol values and unknown exporter values fail startup with a typed error; `--help`/`--version` never build exporters.
- [ ] A request produces an `http.server.request` span that closes after the last body byte (streaming included) and carries only allow-listed attributes.
- [ ] `http.server.request.duration` and `http.server.active_requests` finalize exactly once on success, error status, and mid-body client disconnect.
- [ ] Log lines inside an exported span carry `trace_id`/`span_id`; `RUST_LOG=agentic_core=error` leaves export unchanged; `OTEL_TRACES_SAMPLER_ARG=0` leaves local logs unchanged; SDK export failures appear locally at `WARN` by default.
- [ ] A Collector that accepts the connection and never responds does not stall request delivery; shutdown completes within its deadline; `SIGTERM` exit is clean on the disabled, healthy, and black-hole-endpoint paths.
- [ ] `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test --workspace`, `pre-commit run --all-files` pass; all new crates declare `rust-version ≤ 1.85` (CI enforcement is tracked separately, see the MSRV issue).

#### Non-goals
Executor-stage spans, W3C extraction/injection for requests, WebSocket instrumentation, execution/delivery outcome semantics (phase 2); project metrics, timing definitions, Collector compose example (phase 3); OTLP log export; gRPC; `agentic-llm-d` wiring (the `telemetry::init` entry point is reusable for it later).

Zheng-Lu:
Fix #338 
Part of #279 — phase 1 of the three-phase sequence proposed on that issue (foundation → execution traces → metrics). 

## Summary

Adds an **opt-in** OpenTelemetry foundation to the `agentic-server` binary. With nothing selected the gateway creates no provider, exporter thread, or network connection and logs exactly as before. When enabled it exports OTLP over HTTP/protobuf, bridges `tracing` spans to OpenTelemetry, and correlates local log lines with the exported trace.

- **Typed `OTEL_*` configuration** (`telemetry/config.rs`): `OTEL_SDK_DISABLED`, `OTEL_{TRACES,METRICS}_EXPORTER` (`none` | `otlp`, default `none` — a documented deviation from the spec default so telemetry stays opt-in), `OTEL_EXPORTER_OTLP_PROTOCOL` (+ per-signal), `OTEL_SERVICE_NAME`. Invalid values are startup errors; `grpc` is rejected with a message naming the reason (the gRPC exporter needs `tonic 0.14`, `rust-version 1.88`, above the MSRV). Endpoint/headers/timeout/sampler/batch settings are read by the SDK and exporter directly with the standard precedence. The parser takes a lookup function so tests never touch the process environment (`std::env::set_var` is `unsafe` in edition 2024).
- **Provider lifecycle** (`telemetry/lifecycle.rs`): providers are built before the Tokio runtime and owned by `main`. After `server::run*` returns on *any* exit path, `main` first stops the runtime (`shutdown_timeout(1 s)`) — dropping any connection task the gateway drain abandoned, which finalizes its span and metrics while the providers still accept them — and only then shuts the providers down from a plain thread under a 3 s deadline (`TelemetryGuard::shutdown_blocking`; an async `shutdown` exists for embedders). `SdkMeterProvider::shutdown_with_timeout` ignores its argument in SDK 0.32, so the deadline is enforced by the guard. Drain (8 s) + runtime (1 s) + telemetry (3 s) stays inside the 30 s Kubernetes grace period. The W3C `TraceContextPropagator` is registered globally.
- **Subscriber stack** (`telemetry/subscriber.rs`): the `fmt` layer keeps `RUST_LOG`; the bridge layer has its own fixed filter (spans only, `INFO`+, and an **allow-list** of this repository's crates — spans declared by dependencies such as `rmcp`, `hyper`, or the SDK are never exported, so export cannot feed back into itself). Log lines inside an exported span end with `trace_id=… span_id=…`. The default `RUST_LOG` now includes `opentelemetry_sdk=warn,opentelemetry-otlp=warn` so queue overflow and export failures are visible locally.
- **Root HTTP span and metrics** (`telemetry/http.rs`): one `http.server.request` span per request (semconv name and attributes, bounded — no query strings, headers, or client addresses; `url.scheme` is the listener's, never the client-controlled request-target scheme), parented on a valid incoming W3C `traceparent` so the parent-based sampler honours the caller's decision (absent/invalid → new root), held open by a response-body wrapper until the body completes or is dropped, marked as errored on 5xx, handler panics, and failed bodies (bounded `error.type`), plus `http.server.request.duration` and `http.server.active_requests` recorded by a `Drop` guard exactly once per request. Installed as the outermost `Router::layer`. `tower_http::TraceLayer` was considered and not used because it has no drop hook, which would let the active-request gauge drift on client disconnects.
- **Docs**: `docs/deploying/observability.md` — enabling, configuration reference, exported data, sampling, log correlation, overflow/shutdown budgets, Kubernetes example.

Dependency delta: `opentelemetry` 0.32, `opentelemetry_sdk` 0.32, `opentelemetry-otlp` 0.32 (`http-proto` + `reqwest-blocking-client` + `reqwest-rustls`), `tracing-opentelemetry` 0.33 (+ `http-body`); all declare `rust-version ≤ 1.85`; no `tonic`; reuses the `reqwest 0.13` line already present via `rmcp`. `opentelemetry 0.33` shipped this week but `tracing-opentelemetry` still pins `^0.32`. `agentic-server-core` is untouched.

### Design defaults (please object here if any is wrong)

| Decision | Default |
| --- | --- |
| Transport | OTLP HTTP/protobuf only; gRPC deferred until the MSRV allows `tonic 0.14` |
| Opt-in | `OTEL_TRACES_EXPORTER` / `OTEL_METRICS_EXPORTER` = `otlp`; `OTEL_SDK_DISABLED=true` wins |
| Sampling | `parentbased_always_on` (spec default); ratio sampling documented |
| Log filter | `RUST_LOG` for local output; bridge filter independent of it |
| Shutdown | gateway drain (8 s) → `Runtime::shutdown_timeout(1 s)` → telemetry flush (3 s) |

### Behaviour notes

- `main` is now a sync `fn` with an explicit runtime. Previously `#[tokio::main]` waited indefinitely for in-flight `spawn_blocking` work at exit; the process now stops the runtime 1 s after telemetry shutdown. After `drain_gateway` and `websocket_tracker.wait_until_idle()` no request work remains, and SQLite recovers an abandoned WAL checkpoint, so this is intentional.
- Correlated log lines are rendered without ANSI colour (the buffered formatter path); other lines are unchanged.
- **MSRV**: `cargo +1.85.0 check --locked` fails on `main` today for reasons unrelated to this PR (`home 0.5.12` → 1.88, `process-wrap 9.1.0` → 1.87 with a real trait-upcasting compile error, `icu_* 2.2.0` → 1.86). This PR adds nothing to that list; a CI MSRV job is therefore tracked separately rather than added here.

### Deferred to phase 2 / 3

Upstream context injection, execution-stage spans (rehydration, inference rounds, tools, compaction, persistence), WebSocket session/request spans, typed execution and delivery outcomes, gateway metrics and timing definitions, Collector compose example, overhead benchmarks.

## Test Plan

- `cargo fmt -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `pre-commit run --all-files` (including the Rust file-size policy): pass.
- `cargo test --workspace --no-fail-fast`: 70 suites, 1672 passed, 0 failed, 11 ignored (require `TEST_POSTGRES_URL`). Existing WebSocket suites pass with the body wrapper in place.
- New tests (+38):
  - `telemetry::config` (9): defaults, per-signal selection, `OTEL_SDK_DISABLED` precedence and strict booleans, unknown exporter values, `grpc` rejected on every protocol variable, service-name trimming, programmatic endpoint never read from env.
  - `telemetry::subscriber` (8): bridge level filter, repository-crate allow-list (`rmcp::service` and other dependency spans rejected even at `INFO`), trace/span ids appended inside exported spans and inherited under non-exported `DEBUG` spans, plain output without a bridge, `RUST_LOG=error` does not affect export, default filter surfaces SDK/exporter warnings.
  - `tests/telemetry_test.rs` (7): disabled config builds nothing and opens no connection; exports carry `service.name` on traces and metrics (protobuf decoded by an in-process OTLP/HTTP stub); shutdown completes when the collector times out; the guard's deadline is honoured against an unresponsive collector (async and blocking paths); dropping an enabled guard inside the runtime does not panic; a request still streaming when the runtime is stopped reaches the collector with `active_requests == 0`.
  - `tests/telemetry_http_test.rs` (10): span attributes and allow-list, streaming span not finished until the body is consumed (gauge = 1 mid-stream), mid-stream body drop finalizes exactly once (gauge back to 0), 5xx marks error status, unmatched routes carry no `http.route` and no URL/query, valid `traceparent` becomes the parent, unsampled `traceparent` suppresses the span but not the metrics, invalid `traceparent` starts a new root, handler panics are recorded as errors and finalized once, failing bodies mark the span.
  - `tests/telemetry_e2e_test.rs` (1): `telemetry::init` + real router over TCP → stub receives `GET`, `GET /health`, `POST /v1/responses` spans, 3 duration samples, `active_requests == 0`, service identity on both signals.
- Manual: real binary with `OTEL_*` set, requests, then `SIGTERM` → exit 0 in ~10 ms; the stub collector received the span export from the shutdown flush; the same with an unroutable endpoint exits cleanly. `mkdocs build --strict` passes.
- Cassette decision: none needed — no client-facing or upstream wire contract, persistence, tool-loop, or streaming-event behaviour changes.


crates/agentic-server/src/telemetry/http.rs
```diff
@@ -0,0 +1,370 @@
...
+    fn completed(&mut self) {
+        self.completed = true;
+    }
+}
+
+impl Drop for SpanOutcome {
+    fn drop(&mut self) {
+        if !self.completed {
+            self.span.record(ATTR_OTEL_STATUS, "ERROR");
+            self.span.record(ATTR_ERROR_TYPE, ERROR_TYPE_PANIC);
```
franciscojavierarceo, line 304: dropping the request future before `next.run(request)` returns also reaches this branch without a panic. for example, a timeout around a pending handler exports `error.type=handler_panic`; aborting an in-flight handler during runtime shutdown has the same path. this turns normal cancellation into a false panic signal. we should distinguish cancellation from unwinding, only record `handler_panic` for an actual panic, and add a pre-header cancellation test alongside the panic test.

docs/deploying/observability.md
```diff
@@ -0,0 +1,194 @@
...
+
+Variables read directly by the OpenTelemetry SDK and OTLP exporter, with the
+standard precedence *signal-specific variable → generic variable → default*:
+
+| Variable | Purpose | Default |
+| --- | --- | --- |
+| `OTEL_EXPORTER_OTLP_ENDPOINT` (`_TRACES_`, `_METRICS_`) | Collector base URL; `/v1/traces` and `/v1/metrics` are appended to the generic endpoint | `http://localhost:4318` |
+| `OTEL_EXPORTER_OTLP_HEADERS` (`_TRACES_`, `_METRICS_`) | `key=value,...` headers, for example vendor authentication | none |
+| `OTEL_EXPORTER_OTLP_TIMEOUT` (`_TRACES_`, `_METRICS_`) | Per-export request timeout in milliseconds; the effective bound on every export call | `10000` |
+| `OTEL_EXPORTER_OTLP_COMPRESSION` | `gzip` or unset | unset |
```
franciscojavierarceo, line 58: `OTEL_EXPORTER_OTLP_COMPRESSION=gzip` currently prevents startup when an OTLP exporter is enabled. the dependency disables default features and does not enable `gzip-http`, so exporter construction returns `UnsupportedCompressionAlgorithm("gzip compression requested but gzip-http feature not enabled")`. we should either enable and test `gzip-http` or document compression as unsupported instead of advertising `gzip` here.

franciscojavierarceo (changes requested):
two issues to fix before merging: cancellation before response headers is reported as a handler panic, and the documented gzip configuration fails exporter initialization. i reproduced both with a focused probe using this commit’s HTTP middleware and the same OpenTelemetry exporter version/features. CI is green; the local server test run was blocked by missing OpenSSL development files.

crates/agentic-server/src/telemetry/http.rs
Zheng-Lu: Fixed in 5308a2b. `SpanOutcome::drop` now checks `std::thread::panicking()`: tokio drops a task's future inside its unwind guard, so a handler panic still reaches the drop while unwinding, whereas a timeout or abort drops it with the panic flag clear. A real panic keeps `otel.status_code=ERROR` + `error.type=handler_panic`; a request dropped before headers is now labelled `error.type=cancelled` with the span status left unset and no status code, so it no longer reads as a failure.

Added two tests next to the panic test in `telemetry_http_test.rs`: one wraps a never-responding handler in `tokio::time::timeout`, the other `abort()`s the spawned task (the runtime-shutdown case). Both assert `cancelled`/Unset and that `active_requests` and the duration histogram stay balanced. Both fail on the previous commit with `left: Error, right: Unset`. Docs updated to list `cancelled` as the third `error.type` value.

docs/deploying/observability.md
Zheng-Lu: Confirmed — reproduced the same `UnsupportedCompressionAlgorithm("gzip compression requested but gzip-http feature not enabled")`. Went with enabling and testing it rather than removing it from the docs, since it's the one knob most collectors expect:

- `opentelemetry-otlp` now has `gzip-http`. It only adds `miniz_oxide`/`adler2` (pure Rust, edition 2021, no `rust-version` floor), so MSRV 1.85 is unaffected.
- `TelemetryConfig` validates `OTEL_EXPORTER_OTLP_COMPRESSION` and the `_TRACES_`/`_METRICS_` variants at startup with the signal-specific → generic → unset precedence, and passes the resolved value to each exporter builder via `with_compression`. `zstd` and the literal `none` are rejected with a clear message (zstd isn't compiled in; the exporter's own parser rejects `none`, so accepting it would just move the failure later).
- The OTLP test stub now inflates `content-encoding: gzip` bodies, and `gzip_compressed_exports_reach_the_collector` asserts every export was gzip-encoded and decodes to the expected span and metric. With the feature removed the test fails with the exact error above.

The doc row moved into the gateway-validated table with the accepted values spelled out.

franciscojavierarceo (approved):
both findings are addressed: pre-header cancellation is distinguished from panic, and gzip export is enabled and covered through protobuf decoding at the collector. i ran 45 telemetry tests in an isolated harness against the updated source, including timeout, abort, panic, and compressed export coverage; all passed. CI is green on 5308a2b.

